PR #46, opened 2014-09-13
Add support for a 'waitDuration' option in startApp

JamieEdge:
This might be required for https://github.com/appium/appium/issues/780.

### DIFF
--- a/lib/adb.js
+++ b/lib/adb.js
@@ -1144,7 +1144,11 @@ ADB.prototype.startApp = function (startAppOptions, cb) {
       }
 
       if (startAppOptions.waitActivity) {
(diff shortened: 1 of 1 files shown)


chrigrahcisco:
Can you provide details on how to exercise the waitDuration option?

edgy360:
It's been a while since I directly used appium-adb, so not sure if it has changed, but you should just need to set this as part of the launch options.

chrigrahcisco:
I happen to be using Appium.app   and see no such option, nor any documentation of a server flag, assuming one exists.

edgy360:
This was a pull request into appium-adb rather than Appium so it didn’t expose a server flag. It would need to be added into the main Appium source too for that functionality to exist if it hasn’t been already.

PR #68, opened 2014-12-29
add JAVA_HOME env variable check

hyeung-trulia:
add JAVA_HOME env variable check before running path.resolve with process.env.JAVA_HOME


jlipps:
Unfortunately because of the way JS's async works, if you want to do this check you'll need to put everything after the check inside the check's callback. Does that make sense?

lib/adb.js
```diff
@@ -494,16 +494,21 @@ ADB.prototype.checkApkCert = function (apk, pkg, cb) {
 };
 
 ADB.prototype.checkCustomApkCert = function (apk, pkg, cb) {
+  var self = this;
```
jlipps: instead of aliasing `this` to `self`, we simply `.bind(this)` any function which needs the `this` context

jlipps:
made another comment, otherwise looks good. and we'll ask you to squash everything into one commit, please

hyeung-trulia:
I think messed the pull request up when trying to squash the commits. I created a new one with just the necessary changes: https://github.com/appium/appium-adb/pull/73

PR #71, opened 2015-02-16
remove testZip step for selendroid

aluedeke:
fix for appium/appium#4414

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -11,64 +11,25 @@ var logger = require('./logger'),
 
 exports.unzipFile = function (zipPath, cb) {
   logger.debug("Unzipping " + zipPath);
-  exports.testZipArchive(zipPath, function (err, valid) {
-    if (valid) {
-      if (exports.isWindows()) {
-        var zip = new AdmZip(zipPath);
-        zip.extractAllTo(path.dirname(zipPath), true);
-        logger.debug("Unzip successful");
-        cb(null, null);
-      } else {
-        var execOpts = {cwd: path.dirname(zipPath), maxBuffer: 524288};
-        exec('unzip -o ' + zipPath, execOpts, function (err, stderr, stdout) {
-          if (!err) {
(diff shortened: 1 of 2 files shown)


aluedeke:
appium/appium#4541

bootstraponline:
:+1:

lib/helpers.js
```diff
@@ -11,64 +11,25 @@ var logger = require('./logger'),
...
-            logger.error("Stdout: " + stdout);
-            cb("Archive could not be unzipped, check appium logs.", null);
-          }
-        });
-      }
+    if (exports.isWindows()) {
+      var zip = new AdmZip(zipPath);
+      zip.extractAllTo(path.dirname(zipPath), true);
+      logger.debug("Unzip successful");
+      cb(null, null);
```
jlipps: can we take the opportunity to fix this pattern? this should just be `cb();`

lib/helpers.js
```diff
@@ -11,64 +11,25 @@ var logger = require('./logger'),
...
+          logger.debug("Unzip successful");
+          cb(null, stderr);
         } else {
-          logger.error("Zip file " + zipPath + " was not valid");
+          logger.error("Unzip threw error " + err);
           logger.error("Stderr: " + stderr);
           logger.error("Stdout: " + stdout);
-          cb("Zip archive did not test successfully, check appium server logs " +
-             "for output", false);
+          cb("Archive could not be unzipped, check appium logs.", null);
```
jlipps: can we take the opportunity to fix this? we should be calling back with an error object, e.g., `cb(new Error("Archive could not be unzipped, check appium logs."))`, and we don't want to pass another param if we have an error.

jlipps:
made a few comments, then good to go

paymand:
The latest unzip version on Ubuntu (6.0-12ubuntu1.3) has the fix and testZip step doesn't fail anymore.

PR #72, opened 2015-02-24
Made adb present check part of instantiation

0x1mason:
Created a factory method to consolidate instantiation with checking adb exists.

@jlipps @sebv The taking of ganders, please if you will.

### DIFF
--- a/lib/adb.js
+++ b/lib/adb.js
@@ -67,6 +67,13 @@ var ADB = function (opts) {
 // exposing logger
 ADB.logger = logger;
 
+ADB.createADB = function (opts, cb) {
+  var adb = new ADB(opts);
+  adb.checkAdbPresent(function (err) {
+    cb(err, adb);
+  });
+};
+
 ADB.prototype.checkSdkBinaryPresent = function (binary, cb) {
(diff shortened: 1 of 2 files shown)


lib/adb.js
```diff
@@ -46,10 +47,16 @@ var ADB = function (opts) {
   if (opts.remoteAdbPort) {
     this.adb.defaultArgs.push("-P", opts.remoteAdbPort);
   }
+  var initError = null;
+  async.series([
+    this._checkAdbPresent.bind(this)
+  ], function(err) {
+    initError = err;
+  });
+  if (initError) throw initError;
```
jlipps: there's an async bug here: this line will usually execute before `this._checkAdbPresent` ever has a chance to.

we can't run async functions inside a constructor. if we want to do something like this we should create an `init` function or something which is async

0x1mason: Ok, I know what I'll do.

sebv: Was also wondering about this `async.series`.

0x1mason:
@jlipps @sebv Let's try again.

PR #73, opened 2015-03-16
add JAVA_HOME env varible check before running path.resolve with process...

hyeung-trulia:
....env.JAVA_HOME and return a more meaningful error message


jlipps:
thanks! will get this into a new publish of appium adb

jlipps:
ok, this has been published and brought into appium master

PR #75, opened 2015-04-16
Adding isEmulator method

moizjv:
- For future use to find if it is a real device or emulator.


bootstraponline:
[:-1:](http://stackoverflow.com/questions/27002663/adb-shell-dumpsys-iphonesubinfo-not-working-since-android-5-0-lollipop)

moizjv:
Ok let me look into it. I was testing with 4.4.4, I will close this for now I will use services to getDevice id. Thanks @bootstraponline !

PR #76, opened 2015-05-27
Porting appium adb 

moizjv:
- Adding gulp, babel and friends
- Adding new ADB class.
- Dividing ADB class methods in to another file. 
- Adding specs and many skips will remove skip as I continue implementing API.


lib/adb.js
```diff
@@ -1,1700 +1,59 @@
...
+   return await adb.getAdbWithCorrectAdbPath();
+ }
+
+}
+
+for (let [mthds, fn] of _.pairs(methods)) {
+  ADB.prototype[mthds] = fn;
+}
+
+export { ADB };
```
sebv: `export default ADB;`

lib/adb.js
```diff
@@ -1,1700 +1,59 @@
...
+   if (!this.javaVersion || parseFloat(this.javaVersion) < 1.7) {
+     this.jars["appium_apk_tools.jar"] = path.resolve(__dirname, '../jars', "appium_apk_tools_1.6.jar");
+   }
+ }
+
+ async createADB (opts) {
+   let adb = new ADB(opts);
+   return await adb.getAdbWithCorrectAdbPath();
+ }
+
```
moizjv: this should go too.

sebv:
Generally when there is just one thing to export, and it is not  confusing, it is better to stick with default export, could also declare one default and one named.

lib/tools/adb-commands.js
```diff
@@ -0,0 +1,34 @@
...
+    this.binaries.aapt = await this.sysCallsHelper.getSdkBinaryPath("aapt");
+    return true;
+  } catch (e) {
+    // TODO log
+    return false;
+  }
+
+};
+
+methods.isZiAlignPresent = async function() {
```
sebv: Missing `p`.

lib/adb.js
```diff
@@ -1,1700 +1,59 @@
...
+import { SystemCallsHelper } from './helpers/system-calls.js';
+import methods from './tools/index.js';
+
+class ADB {
+
+ constructor (opts = {}) {
+   if (typeof opts.sdkRoot === "undefined") {
+     opts.sdkRoot = process.env.ANDROID_HOME || '';
+   }
+   this.sdkRoot = opts.sdkRoot;
```
jlipps: instead of this list of assignments, could we also do:

```js
Object.assign(this, opts);
```

?

package.json
```diff
@@ -2,9 +2,11 @@
   "name": "appium-adb",
   "version": "1.7.5",
   "description": "appium-adb",
-  "main": "lib/main.js",
+  "main": "./build/index.js",
   "scripts": {
-    "test": "./node_modules/.bin/mocha test/unit/*-specs.js"
+    "prepublish": "./node_modules/.bin/gulp prepublish",
```
sebv: I was wrong with this `./node_modules/.bin` thing, npm already adds it to the path so `gulp prepublish` is the right way.

lib/adb.js
```diff
@@ -1,1700 +1,59 @@
...
+     this.adb.defaultArgs.push("-P", opts.remoteAdbPort);
+   }
+   this.curDeviceId = null;
+   this.emulatorPort = null;
+   this.logcat = null;
+   this.binaries = {};
+   this.instrumentProc = null;
+   this.javaVersion = opts.javaVersion;
+   this.suppressKillServer = opts.suppressAdbKillServer;
+   this.sysCallsHelper = new SystemCallsHelper(this.sdkRoot);
```
jlipps: i'd call this `this.sysCalls` to save space

lib/adb.js
```diff
@@ -1,1700 +1,59 @@
...
+   this.jars = {};
```
jlipps: to keep this constructor from getting too long, i'd move the jar logic into its own function `initJars()` and simply call it from here

lib/adb.js
```diff
@@ -1,1700 +1,59 @@
...
+     this.jars[jarName] = path.resolve(__dirname, '../jars', jarName);
+   }
+   if (!this.javaVersion || parseFloat(this.javaVersion) < 1.7) {
+     this.jars["appium_apk_tools.jar"] = path.resolve(__dirname, '../jars', "appium_apk_tools_1.6.jar");
+   }
+ }
+
+ async createADB (opts) {
+   let adb = new ADB(opts);
+   return await adb.getAdbWithCorrectAdbPath();
```
jlipps: no need to `return await`

lib/helpers/helpers.js
```diff
@@ -0,0 +1,48 @@
+import fs from 'fs';
+import path from 'path';
+import os from 'os';
+import _ from 'lodash';
+import { exec } from 'teen_process';
+import B from 'bluebird';
+
+const isWindows = (os.type() === 'Windows_NT');
+const _fs = {
```
jlipps: would rather this be the other way around. i.e:

```js
import _fs from 'fs';
...
const fs = {
...
}
```

lib/helpers/helpers.js
```diff
@@ -0,0 +1,48 @@
...
+};
+
+async function getDirectories (rootPath) {
+  let files = await _fs.readdir(rootPath);
+  let dirs = [];
+  for (let file of files) {
+    let pathString = path.resolve(rootPath, file);
+    if ((await _fs.lstat(pathString)).isDirectory()) dirs.push(file);
+  }
+  // It is not a clean way to sort it, but in this case would work fine because we have numerics and alphanumeric
```
jlipps: looks like the comment needs to be broken up into more lines, seems longer than 80 chars

lib/helpers/helpers.js
```diff
@@ -0,0 +1,48 @@
...
+  lstat: B.promisify(fs.lstat),
+  readdir: B.promisify(fs.readdir)
+};
+
+async function getDirectories (rootPath) {
+  let files = await _fs.readdir(rootPath);
+  let dirs = [];
+  for (let file of files) {
+    let pathString = path.resolve(rootPath, file);
+    if ((await _fs.lstat(pathString)).isDirectory()) dirs.push(file);
```
jlipps: i don't like one-line ifs. can we do:

```
let stat = await fs.lstat(pathString);
if (stat.isDirectory()) {
  dirs.push(file);
}
```

lib/helpers/system-calls.js
```diff
@@ -0,0 +1,65 @@
...
+      }
+      for (let loc of binaryLocs) {
+        let util = support.util;
+        let flag = await util.fileExists(loc);
+        if (flag) binaryLoc = loc;
+      }
+      if (binaryLoc === null) {
+        throw new Error("Could not find " + binary + " in tools, platform-tools, " +
+                     "or supported build-tools under \"" + this.sdkRoot + "\"; " +
+                     "do you have the Android SDK installed at this location?");
```
imurchie: These errors can be interpolated strings.

lib/tools/android-manifest.js
```diff
@@ -0,0 +1,81 @@
+import { prettyExec } from '../helpers/helpers.js';
+
+let manifestMethods = {};
+
+// android:process= may be defined in AndroidManifest.xml
+// http://developer.android.com/reference/android/R.attr.html#process
+// note that the process name when used with ps must be truncated to the last 15 chars
+// ps -c com.example.android.apis becomes ps -c le.android.apis
+manifestMethods.processFromManifest = async function (localApk) {
+  if(await this.isAaptPresent()) {
```
imurchie: Space after `if`.

test/adb-specs.js
```diff
@@ -0,0 +1,123 @@
...
+const should = chai.should();
+chai.use(chaiAsPromised);
+
+describe('ADB', () => {
+  it('should correctly return adb if present', async () => {
+    let adb = new ADB();
+    let temp = await adb.createADB();
+    should.exist(temp.path);
+  });
+  it('should correctly return adb if present ANDROID_HOME is not present', async () => {
```
imurchie: `and ANDROID_HOME is not present`?

moizjv: get its from the system path

imurchie: The message is ungrammatical, that is all. I _think_ there needs to be an `and`.

moizjv: oops yeah my bad .... thanks for pointing that out, my brain is reading english like code with no emphasis on grammer :p

lib/helpers/helpers.js
```diff
@@ -0,0 +1,48 @@
...
+    s = s.replace(/"/g, '\\"');
+    return '"' + s + '"';
+  }
+  return s;
+}
+
+async function prettyExec (/*cmd, args, opts, cb*/...vargs) {
+  let cmdOpts = vargs[2] || {};
+  delete cmdOpts.wrapArgs;
+  if (_.isEmpty(cmdOpts)) cmdOpts = undefined;
```
jlipps: let's always set to `null` instead of `undefined`

```js
if (_.isEmpty(cmdOpts)) {
  cmdOpts = null;
}
```

lib/helpers/system-calls.js
```diff
@@ -0,0 +1,65 @@
...
+import { getDirectories, prettyExec, isWindows } from './helpers';
+
+class SystemCallsHelper {
+
+  constructor (sdkRoot) {
+    this.sdkRoot = sdkRoot;
+  }
+
+  async getSdkBinaryPath (binary) {
+    log.info("Checking whether " + binary + " is present");
```
jlipps:
```js
log.info(`Checking whether ${binary} is present`);
```

package.json
moizjv: so we don't need it ?

lib/helpers/system-calls.js
```diff
@@ -0,0 +1,65 @@
...
+      }
+      binaryLoc = binaryLoc.trim();
+      log.info("Using " + binary + " from " + binaryLoc);
+      return binaryLoc;
+    } else {
+      log.warn("The ANDROID_HOME environment variable is not set to the Android SDK root directory path. " +
+                  "ANDROID_HOME is required for compatibility with SDK 23+. Checking along PATH for " + binary + ".");
+      try {
+        let {stdout} = await prettyExec(cmd, [binary], { maxBuffer: 524288 });
+        log.info("Using " + binary + " from " + stdout);
```
jlipps: use string interpolation here

lib/helpers/system-calls.js
```diff
@@ -0,0 +1,65 @@
...
+import support from 'appium-support';
+import { getDirectories, prettyExec, isWindows } from './helpers';
+
+class SystemCallsHelper {
+
+  constructor (sdkRoot) {
+    this.sdkRoot = sdkRoot;
+  }
+
+  async getSdkBinaryPath (binary) {
```
jlipps: this function is really long. can we break the conditional blocks out into their own helper functions?

lib/helpers/system-calls.js
```diff
@@ -0,0 +1,65 @@
...
+      if (binaryLoc === null) {
+        throw new Error("Could not find " + binary + " in tools, platform-tools, " +
+                     "or supported build-tools under \"" + this.sdkRoot + "\"; " +
+                     "do you have the Android SDK installed at this location?");
+      }
+      binaryLoc = binaryLoc.trim();
+      log.info("Using " + binary + " from " + binaryLoc);
+      return binaryLoc;
+    } else {
+      log.warn("The ANDROID_HOME environment variable is not set to the Android SDK root directory path. " +
```
jlipps: line length and breaking seem off here

lib/helpers/system-calls.js
```diff
@@ -0,0 +1,65 @@
...
+                  "ANDROID_HOME is required for compatibility with SDK 23+. Checking along PATH for " + binary + ".");
+      try {
+        let {stdout} = await prettyExec(cmd, [binary], { maxBuffer: 524288 });
+        log.info("Using " + binary + " from " + stdout);
+        binaryLoc = '"' + stdout.trim() + '"';
+        return binaryLoc;
+      } catch(e) {
+        throw Error("Could not find " + binary + ". Please set the ANDROID_HOME " +
```
jlipps: `throw new Error`

lib/tools/adb-commands.js
```diff
@@ -0,0 +1,34 @@
+let methods = {};
+
+methods.getAdbWithCorrectAdbPath = async function () {
+  try {
+    this.adb.path = await this.sysCallsHelper.getSdkBinaryPath("adb");
+    this.binaries.adb = this.adb.path;
+  } catch (err) {
+    return err;
```
jlipps: this is a very strange pattern. should probably catch inside whoever's calling this method instead

lib/tools/android-manifest.js
```diff
@@ -0,0 +1,81 @@
...
+    //  logger.debug("Retrieving process from manifest.");
+    try {
+      let {stdout} = await prettyExec(this.binaries.aapt, ['dump', 'xmltree', localApk, 'AndroidManifest.xml'],
```
jlipps: no need for max buffer if prettyExec uses teen_process.exec under the hood

jlipps: take this array and make it its own variable so this line is shorter, i.e.:

```js
let args = ['dump', 'xmltree', localApk, 'AndroidManifest.xml'];
let {stdout} = await prettyExec(this.binaries.aapt, args);
```

package.json
sebv: nope, just `gulp prepublish`, same for other scripts.

lib/tools/android-manifest.js
```diff
@@ -0,0 +1,81 @@
...
+                 { maxBuffer: 524288 });
+      let result = null;
+      let lines = stdout.split("\n");
+      let applicationRegex = new RegExp(/\s+E: application \(line=\d+\).*/);
+      let applicationFound = false;
+      let attributeRegex = new RegExp(/\s+A: .+/);
+      let processRegex = new RegExp(/\s+A: android:process\(0x01010011\)="([^"]+).*"/);
+      for (let i = 0; i < lines.length; i++) {
```
jlipps: since we don't use the index here, this should be:

```js
for (let line of lines) {
```

lib/tools/android-manifest.js
jlipps: rather than making this a conditional, this could just be an assertion function. it saves us space and from having to use a conditional to throw the error. like:

```js
await this.assertAaaptPresent(); 
```

And that line will throw the error without needing a conditional

lib/tools/android-manifest.js
```diff
@@ -0,0 +1,81 @@
...
+    } catch (e) {
+      throw new Error("processFromManifest failed. " + e);
+    }
+  } else {
+    throw new Error("Could not find aapt");
+  }
+};
+
+manifestMethods.packageAndLaunchActivityFromManifest = async function (localApk) {
+  if(await this.isAaptPresent()) {
```
jlipps: same thing here about aapt assertion

lib/tools/android-manifest.js
```diff
@@ -0,0 +1,81 @@
...
+    //logger.debug("Extracting package and launch activity from manifest.");
+    try {
+      let {stdout} = await prettyExec(this.binaries.aapt, ['dump', 'badging', localApk], { maxBuffer: 524288 });
```
jlipps: same thing about maxBuffer and args (if they break the line)

lib/tools/android-manifest.js
```diff
@@ -0,0 +1,81 @@
...
+      } else {
+        apkPackage = null;
+      }
+      let apkActivity = new RegExp(/launchable-activity: name='([^']+)'/g).exec(stdout);
+      if (apkActivity && apkActivity.length >= 2) {
+        apkActivity = apkActivity[1];
+      } else {
+        apkActivity = null;
+      }
+      //logger.debug("badging package: " + apkPackage);
```
jlipps: let's remove commented code

test/adb-specs.js
jlipps: looks like there's an extra 'present' here

moizjv:
Done :)

lib/adb.js
```diff
@@ -1,1700 +1,66 @@
...
+
+ initJars () {
+   let tempJars = ['move_manifest.jar', 'sign.jar', 'appium_apk_tools.jar',
+                   'unsign.jar', 'verify.jar'];
+   for (let jarName of tempJars) {
+     this.jars[jarName] = path.resolve(__dirname, '../jars', jarName);
+   }
+   if (!this.javaVersion || parseFloat(this.javaVersion) < 1.7) {
+     this.jars["appium_apk_tools.jar"] = path.resolve(__dirname,
+                                                      '../jars',
```
jlipps: why mix quote styles here?

lib/adb.js
```diff
@@ -1,1700 +1,66 @@
...
+   }
+ }
+
+ async createADB (opts) {
+   let adb = new ADB(opts);
+   return adb.getAdbWithCorrectAdbPath();
+ }
+}
+
+for (let [mthds, fn] of _.pairs(methods)) {
```
jlipps: it's not really `mthds` it's more like `mthd`, which is ambiguous because it could be the function itself. what about `name`? or `fnName` since you're using `fn`?

lib/tools/android-manifest.js
```diff
@@ -0,0 +1,77 @@
...
+          if (result.length > 15) {
+            result = result.substr(result.length - 15);
+          }
+          break;
+        }
+      }
+    }
+    return result;
+  } catch (e) {
+    throw new Error("processFromManifest failed. " + e);
```
jlipps: interpolated string here please! Also I'd prefer for this to be something like:

```js
throw new Error(`processFromManifest failed. Original error: ${e.message}`);
```

lib/tools/android-manifest.js
```diff
@@ -0,0 +1,77 @@
...
+    } else {
+      apkPackage = null;
+    }
+    let apkActivity = new RegExp(/launchable-activity: name='([^']+)'/g).exec(stdout);
+    if (apkActivity && apkActivity.length >= 2) {
+      apkActivity = apkActivity[1];
+    } else {
+      apkActivity = null;
+    }
+    log.debug("badging package: " + apkPackage);
```
jlipps: interpolate here, next line, and a few lines down

lib/tools/system-calls.js
```diff
@@ -0,0 +1,77 @@
...
+  binaryLoc = binaryLoc.trim();
+  log.info(`Using ${binaryName} from ${binaryLoc}`);
+  return binaryLoc;
+};
+
+systemCallMethods.getBinaryFromPath = async function(binaryName) {
+  let binaryLoc = null;
+  let cmd = this.getCommandForOS(binaryName);
+  try {
+    let {stdout} = await prettyExec(cmd, [binaryName], { maxBuffer: 524288 });
```
jlipps: no need for maxBuffer

lib/tools/system-calls.js
```diff
@@ -0,0 +1,77 @@
...
+    log.info(`Using ${binaryName} from ${stdout}`);
+    binaryLoc = '"' + stdout.trim() + '"';
```
jlipps: I don't think we need to wrap this in spaces anymore since teen_process will take care of it. But we should have tests with and without spaces in binary paths

lib/tools/system-calls.js
```diff
@@ -0,0 +1,77 @@
...
+    return binaryLoc;
+  } catch(e) {
+    throw new Error(`Could not find ${binaryName} Please set the ANDROID_HOME ` +
+                `environment variable with the Android SDK root directory path.`);
```
jlipps: this line should be 4 spaces to the right

test/android-manifest-specs.js
```diff
@@ -0,0 +1,24 @@
+import chai from 'chai';
+import chaiAsPromised from 'chai-as-promised';
+import 'mochawait';
+import ADB from '../lib/adb.js';
+
+//const should = chai.should();
```
jlipps: if we're not using this, let's comment it out

lib/adb.js
```diff
@@ -1,1700 +1,66 @@
...
-    );
-  }.bind(this));
-};
-
-ADB.getAdbServerPort = function () {
-  return process.env.ANDROID_ADB_SERVER_PORT || 5037;
-};
-
-module.exports = ADB;
+// transpile:mocha
```
imurchie: Do we need this?

lib/adb.js
```diff
@@ -1,1700 +1,66 @@
...
+import path from 'path';
+import methods from './tools/index.js';
+
+class ADB {
+
+ constructor (opts = {}) {
+   if (typeof opts.sdkRoot === "undefined") {
+     opts.sdkRoot = process.env.ANDROID_HOME || '';
+   }
+   const defaults = {
```
imurchie: Do we have a set style for these sorts of constants? I myself would prefer it to be in the root scope of the module, as something like `const DEFAULT_OPTS = {...}`, so it is visible as what it is. A constant in function scope seems... odd.

lib/helpers.js
```diff
@@ -1,158 +1,42 @@
...
-    } else {
-      cb("Zip archive was not found.", false);
+import _fs from 'fs';
+import path from 'path';
+import os from 'os';
+import _ from 'lodash';
+import { exec } from 'teen_process';
+import B from 'bluebird';
+
+const isWindows = (os.type() === 'Windows_NT');
```
imurchie: https://github.com/appium/appium-support/blob/master/lib/system.js#L6

PR #77, opened 2015-05-29
Adding new helper methods and android-manifest methods with tests

moizjv:
- New helper methods and tests
- New android-manifest methods and tests


lib/helpers.js
```diff
@@ -1,6 +1,10 @@
 import _fs from 'fs';
 import path from 'path';
 import B from 'bluebird';
+import { system,util } from 'appium-support';
```
jlipps: `system, util`

lib/helpers.js
```diff
@@ -24,4 +28,56 @@ async function getDirectories (rootPath) {
...
+  if (typeof androidHome !== "string") {
+    log.error("ANDROID_HOME was not exported!");
+    return null;
+  }
+  // TODO import it from appium support
+  let locs = ['android-4.2', 'android-17', 'android-4.3', 'android-18',
+      'android-4.4', 'android-19', 'android-L', 'android-20', 'android-5.0',
+      'android-21'];
+
+  for(let loc of locs.reverse()) {
```
jlipps: `for (`

lib/helpers.js
```diff
@@ -24,4 +28,56 @@ async function getDirectories (rootPath) {
...
+    let platforms = path.resolve(androidHome, 'platforms');
```
jlipps: since you're doing `path.resolve(platforms, loc)` twice, I'd put it in a variable

lib/helpers.js
```diff
@@ -24,4 +28,56 @@ async function getDirectories (rootPath) {
...
+  log.debug(`Unzipping ${zipPath}`);
+  try {
+    await assertZipArchive(zipPath);
+    if (system.isWindows()) {
+      let zip = new AdmZip(zipPath);
+      zip.extractAllTo(path.dirname(zipPath), true);
+      log.debug("Unzip successful");
+    } else {
+      let execOpts = {cwd: path.dirname(zipPath)};
+      await exec('unzip', ['-o', zipPath], execOpts);
```
jlipps: does putting the opts directly into the function call here cause the line to be too long? if not, no need to make a variable for them

lib/helpers.js
```diff
@@ -24,4 +28,56 @@ async function getDirectories (rootPath) {
...
+      log.debug("Unzip successful");
+    }
+  } catch (e) {
+    throw new Error(`Error occored while unzipping. Original error ${e.message}`);
```
jlipps: occored --> occurred

and I would say `Original error: ${e.message}` so it reads more cleanly (i.e., add a colon)

lib/helpers.js
```diff
@@ -24,4 +28,56 @@ async function getDirectories (rootPath) {
...
+  }
+}
+
+async function assertZipArchive (zipPath) {
+  log.debug(`Testing zip archive: ${zipPath}`);
+  if (system.isWindows()) {
+    if (await util.fileExists(zipPath)) {
+      log.debug("Zip archive tested clean");
+    } else {
+      return new Error(`Zip archive not present at ${zipPath}`);
```
jlipps: why are we returning an error instead of throwing an error?

lib/tools/android-manifest.js
```diff
@@ -74,4 +81,60 @@ manifestMethods.packageAndLaunchActivityFromManifest = async function (localApk)
...
+  log.debug('Compiling manifest.');
+  try {
+    let args = ['package', '-M', manifest, '--rename-manifest-package',
+                manifestPackage, '--rename-instrumentation-target-package',
+                targetPackage, '-I', path.resolve(platformPath, 'android.jar'),
+                '-F', manifest + '.apk', '-f'];
+    await exec(this.binaries.aapt, args);
+    log.debug("Compiled manifest");
+  } catch (err) {
+    throw new Error(`Error compiling manifest. Original error ${err.message}`);
```
jlipps: `Original error: ${err.message}`

so we don't get things like "Original error could not open file". what? the error couldn't open a file?

lib/tools/android-manifest.js
```diff
@@ -74,4 +81,60 @@ manifestMethods.packageAndLaunchActivityFromManifest = async function (localApk)
...
+  }
+};
+
+manifestMethods.insertManifest = async function (manifest, srcApk, dstApk) {
+  log.debug(`Inserting manifest, src: ${srcApk} dst: ${dstApk}`);
+  try {
+    await unzipFile(`${manifest}.apk`);
+    await ncp(srcApk, dstApk);
+    log.debug("Testing new tmp apk.");
```
jlipps: no need for trailing periods in any of these logs

lib/tools/android-manifest.js
```diff
@@ -74,4 +81,60 @@ manifestMethods.packageAndLaunchActivityFromManifest = async function (localApk)
...
+      await exec(java, args);
+    } else {
+      // Insert compiled manifest into /tmp/appPackage.clean.apk
+      // -j = keep only the file, not the dirs
+      // -m = move manifest into target apk.
+      await exec('zip', ['-j', '-m', dstApk, manifest]);
+    }
+    log.debug("Inserted manifest.");
+  } catch(e) {
+    throw new Error(`Error insterting manifest. Original error: ${e.message}`);
```
jlipps: nice colon :+1:  :stuck_out_tongue:

test/android-manifest-specs.js
```diff
@@ -2,22 +2,65 @@ import chai from 'chai';
...
     await adb.createADB();
+
+  });
+  it('should correctly parse packageAndLaunchActivityFromManifest', async () => {
     let {apkPackage, apkActivity} = await adb.packageAndLaunchActivityFromManifest('./test/ContactManager.apk');
     apkPackage.should.be.equal('com.example.android.contactmanager');
     apkActivity.should.be.equal('com.example.android.contactmanager.ContactManager');
   });
+  it('should correctly have internet permission', async () => {
+    let flag = await adb.hasInternetPermissionFromManifest('./test/ContactManager-selendroid.apk');
```
jlipps: hmm. not sure what i think about these relative paths. what if we run the tests from somewhere other than the main directory? it'd be safer to use something like `path.resolve(__dirname, '..', '..', 'test', 'ContactManager-selendroid.apk')`

which you could just instantiate as a constant at the top of the file

test/android-manifest-specs.js
```diff
@@ -2,22 +2,65 @@ import chai from 'chai';
...
+    flag.should.be.true;
+  });
+  it('should correctly not have internet permission', async () => {
+    let flag = await adb.hasInternetPermissionFromManifest('./test/ContactManager.apk');
+    flag.should.be.false;
+  });
+  it('should compile and insert manifest', async () => {
+    let appPackage    = 'com.example.android.contactmanager',
```
jlipps: i know it kinda looks pretty, but we don't line things up on = in lists like this. i think we should either do it everywhere or nowhere

test/android-manifest-specs.js
```diff
@@ -2,22 +2,65 @@ import chai from 'chai';
...
+        tmpDir        = path.resolve(__dirname, '..', '..', 'test', 'temp'),
+        newServerPath = path.resolve(tmpDir, `selendroid.${appPackage}.apk`),
+        newPackage    = 'com.example.android.contactmanager.selendroid',
+        srcManifest   = path.resolve(__dirname, '..', '..', 'test',
+                                   'selendroid', 'AndroidManifest.xml'),
+        dstDir        = path.resolve(__dirname, '..', '..', 'test', 'temp', appPackage),
+        dstManifest   = path.resolve(dstDir, 'AndroidManifest.xml'),
+        serverPath    = path.resolve(__dirname, "..", "..", "test", "selendroid",
```
jlipps: lot of mixing of single vs double quotes in this section. also some of the second lines aren't lined up correctly underneath the first line

jlipps:
ok done with review! looks really great overall, made a bunch of style comments

sebv:
Try to put e2e tests in files finishing by `e2e-spec.js` (Anything which actually starts an emulator, or call adb.).

lib/helpers.js
```diff
@@ -24,4 +28,56 @@ async function getDirectories (rootPath) {
...
+    return null;
+  }
+  // TODO import it from appium support
+  let locs = ['android-4.2', 'android-17', 'android-4.3', 'android-18',
+      'android-4.4', 'android-19', 'android-L', 'android-20', 'android-5.0',
+      'android-21'];
+
+  for(let loc of locs.reverse()) {
+    let platforms = path.resolve(androidHome, 'platforms');
+    if(await util.fileExists(path.resolve(platforms, loc))) {
```
imurchie: `if (`

lib/tools/android-manifest.js
```diff
@@ -74,4 +81,60 @@ manifestMethods.packageAndLaunchActivityFromManifest = async function (localApk)
...
+  }
+};
+
+manifestMethods.hasInternetPermissionFromManifest = async function (localApk) {
+  try {
+    await this.initAapt();
+    log.debug("Checking if has internet permission from manifest.");
+    let {stdout} = await exec(this.binaries.aapt, ['dump', 'badging', localApk]);
+    return new RegExp("uses-permission:.*'android.permission.INTERNET'").test(stdout);
```
imurchie: I don't know about anyone else, but I'd prefer regexp literals where it's not dynamic. That way everything is at load-time, not run-time.

test/android-manifest-specs.js
```diff
@@ -2,22 +2,65 @@ import chai from 'chai';
...
+                                  "selendroid.apk");
+    // deleting temp directory if present
+    await exec('rm', ['-rf', tmpDir]);
```
imurchie: Are we not using `rimraf` anymore?

moizjv:
@jlipps @imurchie thanks for the feedback made all the changes :). @sebv I am planning to put it under functional folder will that work?

jlipps:
LGTM

imurchie:
:+1:

sebv:
@moiz I've added e2e pattern to the boilerplate yesterday: https://github.com/appium/appium-gulp-plugins/blob/master/lib/boilerplate.js#L73

Even in a `functional` folder, it is probably a good idea to keep the e2e-specs pattern, less config.

PR #78, opened 2015-06-02
Adding more functions of adb and refactoring tests

moizjv:
- Refactored tests added functional and unit tests
- Added more methods and tests to ADB. 
- Unit tests are still work in progress, aiming to achieve more coverage.


lib/tools/adb-commands.js
```diff
@@ -14,4 +16,97 @@ methods.initZipAlign = async function() {
...
+    let engines = [];
+    for (let line of stdout.split('\n')) {
+      if (line.length > 0 && line[0] !== ' ') {
+        // remove newline and trailing colon, and add to the list
+        engines.push(line.trim().replace(/:$/, ''));
+      }
+    }
+    return engines;
+  } catch (e) {
+    throw new Error(`Error getting available IME's. Original error: ${e.message}`);
```
jlipps: in all of these places where we're throwing errors: in the current code, are we also logging the errors? or no

lib/tools/adb-commands.js
```diff
@@ -14,4 +16,97 @@ methods.initZipAlign = async function() {
...
+  }
+};
+
+methods.enabledIMEs = async function () {
```
jlipps: lot of duplication between this and the previous function, suggests a factoring out of something

lib/tools/adb-commands.js
```diff
@@ -14,4 +16,97 @@ methods.initZipAlign = async function() {
...
+      }
+    }
+    return engines;
+  } catch (e) {
+    throw new Error(`Error getting available IME's. Original error: ${e.message}`);
+  }
+};
+
+methods.enableIME = async function (imeId) {
+  return this.shell(['ime', 'enable', imeId]);
```
jlipps: are there actual return values of these ime methods? if not, I would probably write it as this instead:

```js
methods.enableIME = async function (imeId) {
  await this.shell(['ime', 'enable', imeId]);
};
```

since it makes it more semantically transparent that what we're doing is simply waiting for the action to complete

lib/tools/adb-commands.js
moizjv: Yes this was actually a after thought, this function was not present and I added it, becuase i see a value in it. Yes will refactor :+1:

lib/tools/system-calls.js
```diff
@@ -75,4 +76,112 @@ systemCallMethods.getBinaryFromPath = async function(binaryName) {
...
+    try {
+      let devices = await this.getConnectedDevices();
+      if (devices.length < 1) {
+        log.debug("Could not find devices, restarting adb server...");
+        await this.restartAdb();
+        // cool down
+        await sleep(200);
+        await getDevices();
+      }
+      return devices;
```
jlipps: it looks like we return this `devices` value regardless if whether we do the restart in the `if` block above. if that's the case, then why is the `if` block even there? seems like we should be reassigning `devices` inside that block somewhere...

lib/tools/system-calls.js
```diff
@@ -75,4 +76,112 @@ systemCallMethods.getBinaryFromPath = async function(binaryName) {
...
+      await exec(this.adb.path, ['kill-server']);
+    } catch (e) {
+      log.error("Error killing ADB server, going to see if it's online anyway");
+    }
+  }
+};
+
+systemCallMethods.adbExec = async function (cmd, opts = {}) {
+  if (!cmd) {
+    throw new Error("You need to pass in a command to exec()");
```
jlipps: maybe change this to "adbExec"?

lib/tools/adb-commands.js
moizjv:
```
ADB.prototype.enableIME = function (imeId, cb) {
  var cmd = 'ime enable ' + imeId;
  this.shell(cmd, cb);
};
```

This was the earlier implementation. And output looks like `Input method com.example.android.softkeyboard/.SoftKeyboard: already enabled`

lib/tools/system-calls.js
moizjv: copy paste curse will change it

test/functional/syscalls-e2e-specs.js
```diff
@@ -0,0 +1,32 @@
...
+    devices.should.have.length.above(0);
+  });
+  it('adbExec should get devices when with devices', async () => {
+    (await adb.adbExec("devices")).should.contain("List of devices attached");
+  });
+  it('isDeviceConnected should be true', async () => {
+    (await adb.isDeviceConnected()).should.be.true;
+  });
+  it('shell should execute command in adb shell ', async () => {
+    (await adb.shell(['getprop', 'ro.build.version.sdk'])).should.be.equal(apiLevel);
```
jlipps: instead of `should.be.equal(` let's do `should.equal(` in any place that has it. it's more grammatical :-)

test/unit/adb-commands-specs.js
```diff
@@ -0,0 +1,79 @@
...
+    afterEach(async () => {
+      adb.shell.restore();
+    });
+    let verifyShellArguments = function (cmd) {
+      (shell.args[0][0]).should.be.deep.equal(cmd);
+    };
+    it('getApiLevel should call shell with correct args', async () => {
+      createStub('21');
+      (await adb.getApiLevel()).should.be.equal(apiLevel);
+      verifyShellArguments([ 'getprop', 'ro.build.version.sdk' ]);
```
jlipps: no need for spaces here

jlipps:
ok, finished review and made comments. overall looks really great! i do like the unit tests and think we should keep going in this direction.

lib/tools/system-calls.js
moizjv: this is little tricky i had an `else { return devices }`. But actually it will never return as we recursively call `await getDevices()`

lib/tools/adb-commands.js
jlipps: I mean, does it make sense to be returning a random stdout string from this Node-js based API? Thinking about what kind of API we're creating a few things make sense to me:

1. returning nothing (it's just an action after all)
2. returning true or false based on whether the action actually took place
3. throwing an error if the action didn't complete successfully

but returning a stdout string seems a little wonky to me.

lib/tools/system-calls.js
jlipps: but just calling `await getDevices()` recursively doesn't get that value back to the original calling function ever, right? this probably needs some heavy unit tests since there's so much actual logic. on first pass I think that would show up a bug. I think this block should probably look like:

```js
      let devices = await this.getConnectedDevices();
      if (devices.length < 1) {
        log.debug("Could not find devices, restarting adb server...");
        await this.restartAdb();
        // cool down
        await sleep(200);
        devices = await getDevices();
      }
      return devices;
```

but i could be wrong

test/unit/adb-commands-specs.js
```diff
@@ -0,0 +1,79 @@
...
+    });
+  };
+  beforeEach(async () => {
+    await adb.createADB();
+  });
+  describe('shell', () => {
+    afterEach(async () => {
+      adb.shell.restore();
+    });
+    let verifyShellArguments = function (cmd) {
```
sebv: You probably want to use a  a mock for this kind of stuff.

test/unit/adb-commands-specs.js
```diff
@@ -0,0 +1,79 @@
...
+      (shell.args[0][0]).should.be.deep.equal(cmd);
```
sebv: I think you can remove `be`

test/unit/adb-commands-specs.js
```diff
@@ -0,0 +1,79 @@
...
+      (await adb.getApiLevel()).should.be.equal(apiLevel);
+      verifyShellArguments([ 'getprop', 'ro.build.version.sdk' ]);
+    });
+    it('availableIMEs should call shell with correct args', async () => {
+      createStub(imeList);
+      (await adb.availableIMEs()).should.have.length.above(0);
+      verifyShellArguments(['ime', 'list', '-a']);
+    });
+    it('enabledIMEs should call shell with correct args', async () => {
+      createStub(imeList);
```
sebv: I would try to avoid setting up stub/mock inside `it` block.

lib/tools/system-calls.js
```diff
@@ -75,4 +76,112 @@ systemCallMethods.getBinaryFromPath = async function(binaryName) {
   }
 };
 
+systemCallMethods.getConnectedDevices = async function () {
+  log.debug("Getting connected devices...");
+  try {
+    let {stdout} = await exec(this.adb.path, ['devices']);
```
imurchie: This is just out of curiosity, but what is up with `let {...} = ...`?

jlipps: @imurchie this is for convenience. `teen_process.exec()` returns an object that looks like:

```js
{
  stdout: 'some string',
  stderr: 'some string',
  code: 0
}
```

Instead of returning an object, the other option was to return an array. Then we could have retrieved the values like:

```js
let [stdout, stderr, code] = await exec(...);
```

But in many cases we want only one type of thing from `exec`. Let's say we only wanted `stderr`, then it would look like:

```js
let [, stderr] = await exec(...);
```

Which is slightly more mysterious than what we can do now (because of object destructuring):

```js
let {stderr} = await exec(...);
```

In other words, returning an object plus ES6 object destructuring gives us the ability to target simply those return values we're interested in.

imurchie: Ah! Thanks. I knew about object destructuring, but didn't clue in to the pattern here.

lib/tools/adb-commands.js
moizjv: We are currently now logging errors, I am not sure if want to log error messages?

test/unit/adb-commands-specs.js
moizjv: Each test case requires different output. I am not sure how I can avoid creating a new stub every time to have a different output?

test/unit/adb-commands-specs.js
moizjv: sinon does not allow me to mock already stubbed function 

`TypeError: Attempted to wrap shell which is already wrapped`

lib/tools/adb-commands.js
jlipps: the nice thing about logging them where we throw them is we get to see what happened just before. not just the overall failure without an idea of where in the log flow it actually took place

moizjv:
@jlipps @sebv made more changes, going to add logging in `catch (e)` cases

lib/tools/adb-commands.js
```diff
@@ -14,4 +17,86 @@ methods.initZipAlign = async function() {
...
+methods.clear = function (pkg) {
+  return this.shell(['pm', 'clear', pkg]);
+};
+
+methods.stopAndClear = async function (pkg) {
+  try {
+    await this.forceStop(pkg);
+    await this.clear(pkg);
+  } catch (e) {
+    log.error(`Cannot stop and clear ${pkg}. Original error: ${e.message}`);
```
jlipps: let's not duplicate the error message. Instead, we should do:

```js
let msg = `Cannot stop and clear ${pkg}. Original error: ${e.message}`;
log.error(msg);
throw new Error(msg);
```

jlipps: we might want to add a function `logAndThrow(msg)` to appium-support since this is a common pattern. then you could just do:

```js
logAndThrow(`Cannot stop and clear ${pkg}. Original error: ${e.message}`);
```

jlipps: or we could add it as a special method on our appium-logger object, so you could do:

```js
log.throw(`Cannot stop and clear ${pkg}. Original error: ${e.message}`);
```

moizjv:
Also added unit test cases for recursion `getDevicesWithRetry` :)

test/unit/adb-commands-specs.js
sebv: Yeah it's either stub or mock, a mock is a stub + expectations.

moizjv: then i will have to stick to stub as i am mocking shell return value

test/unit/adb-commands-specs.js
sebv: For instance:
```js
describe(''enabledIMEs', () => {
    before(() = {  createStub(imeList);  });
     it('should call shell with correct args', async', () => {
         ...
     });
     after(() => { clearStub() })
});
```

moizjv: Ohh I see what you are saying.

test/unit/adb-commands-specs.js
sebv: btw there are good examples here http://dailyjs.com/2013/02/14/backbone-tutorial-12/
https://www.dropbox.com/s/xft7u9ub4rpy93a/Screenshot%202015-06-03%2009.00.13.png?dl=0

lib/tools/system-calls.js
```diff
@@ -70,9 +71,127 @@ systemCallMethods.getBinaryFromPath = async function(binaryName) {
...
+      log.debug(`${devices.length} device(s) connected`);
+      return devices;
+    }
+  } catch (e) {
+    log.error(`Error while getting connected devices. Original error: ${e.message}`);
+    throw new Error(`Error while getting connected devices. Original error: ${e.message}`);
+  }
+};
+
+systemCallMethods.getDevicesWithRetry = async function (timeoutMs = 20000) {
```
sebv: Catching up with the slack discussion. This should probably be refactored even before we start thinking about test, the error handling code is duplicated.

test/unit/adb-commands-specs.js
sebv: @moizv you can set the mock return value, see above.

moizjv:
@sebv refactored test to include mocks instead of stubs

PR #83, opened 2015-06-11
Adding apk-utilities and functional and unit tests for it

moizjv:
@jlipps @sebv please review.
- Adding apk-utilities 
- Adding functional tests
- Adding unit tests


lib/tools/apk-utils.js
```diff
@@ -0,0 +1,165 @@
...
+  }
+};
+
+apkUtilsMethods.startApp = async function (startAppOptions = {}) {
+  try {
+    if (!startAppOptions.activity || !startAppOptions.pkg) {
+      log.errorAndThrow("activity and pkg is required for launching application");
+    }
+    // initializing defaults
+    _.defaults(startAppOptions, {
```
sebv: Better clone the options before using defaults. It is modifying startAppOptions. 
(https://lodash.com/docs#defaults)

lib/helpers.js
```diff
@@ -120,6 +120,37 @@ function isScreenOnFully (dumpsys) {
...
+  if (startAppOptions.action) {
+    cmd.push('-a', startAppOptions.action);
+  }
+  if (startAppOptions.category) {
+    cmd.push('-c', startAppOptions.category);
+  }
+  if (startAppOptions.flags) {
+    cmd.push('-f', startAppOptions.flags);
+  }
+  if(startAppOptions.optionalIntentArguments) {
```
sebv: space. you can catch them with `ack 'if\(' lib test`

lib/tools/apk-utils.js
```diff
@@ -0,0 +1,165 @@
...
+        /mFocusedApp.+Record\{.*\s([^\s\/\}]+)\/([^\s\/\}]+)(\s[^\s\/\}]+)*\}/);
+  try {
+    let stdout = await this.shell(cmd);
+    let foundNullMatch = false;
+    for (let line of stdout.split("\n")) {
+      let foundMatch = searchRe.exec(line);
+      if (foundMatch) {
+        let appPackage = foundMatch[1].trim(),
+            appActivity = foundMatch[2].trim();
+        return {appPackage, appActivity};
```
sebv: Can use old syntax and get rid of the variable above.

sebv:
Few comments, otherwise :+1:

lib/tools/apk-utils.js
```diff
@@ -0,0 +1,165 @@
...
+let apkUtilsMethods = {};
+
+apkUtilsMethods.isAppInstalled = async function (pkg) {
+  try {
+    let installed = false;
+    log.debug(`Getting install status for ${pkg}`);
+    let apiLevel = await this.getApiLevel();
+    let thirdparty = apiLevel >= 15 ? "-3" : "";
+    let stdout = await this.shell(['pm', 'list', 'packages', thirdparty, pkg]);
+    let apkInstalledRgx = new RegExp('^package:' + pkg.replace(/(\.)/g, "\\$1") +
```
imurchie: Can we use an interpolated string here?

lib/tools/apk-utils.js
```diff
@@ -0,0 +1,165 @@
...
+                                     '$', 'm');
+    installed = apkInstalledRgx.test(stdout);
+    log.debug("App is" + (!installed ? " not" : "") + " installed");
```
imurchie: As above.

lib/tools/apk-utils.js
```diff
@@ -0,0 +1,165 @@
...
+  }
+};
+
+apkUtilsMethods.waitForActivityOrNot = async function (pkg, activity, not,
+                                                       waitMs = 20000) {
+  if (!pkg || !activity) {
+    throw new Error("Package and activity required.");
+  }
+  log.debug(`Waiting for pkg ${pkg} \ and activity ${activity} \ to ` +
+            (not ? `not ` : ``) + `be focused`);
```
imurchie: Why is this not included in the string template? 

```javascript
log.debug(`Waiting for pkg ${pkg} and activity ${activity} to ${not ? 'not ' : ''}be focused`);
```

lib/tools/apk-utils.js
```diff
@@ -0,0 +1,165 @@
...
+  let endAt = Date.now() + waitMs;
+  let activityRelativeName = getActivityRelativeName(pkg, activity);
+  while (Date.now() < endAt) {
+    let {appPackage, appActivity} = await this.getFocusedPackageAndActivity();
+    let foundAct = ((appPackage === pkg) && (activityRelativeName === appActivity));
+    if ((!not && foundAct) || (not && !foundAct)) {
+      return;
+    }
+    // cool down
+    await sleep(750);
```
imurchie: Can we give an actual reason for this `sleep`?

lib/tools/apk-utils.js
```diff
@@ -0,0 +1,165 @@
...
+  }
+  log.errorAndThrow(`${pkg}/${activityRelativeName} never` +
+                    (not ? `stopped` : `started`));
```
imurchie: As above.

test/functional/apk-utils-e2e-specs.js
```diff
@@ -0,0 +1,65 @@
...
+                          activity: 'ContactManage'}).should.eventually
+                                                     .be.rejectedWith('Activity');
+    });
+    it('should throw error for wrong wait activity', async () => {
+      await adb.install(contactManagerPath);
+      await adb.startApp({pkg: 'com.example.android.contactmanager',
+                          activity: 'ContactManager',
+                          waitActivity: 'ContactManage',
+                          waitDuration: 1000}).should.eventually
+                          .be.rejectedWith('ContactManage');
```
imurchie: `ContactManager`? 

Also, there are different alignment styles going on here. See line 38 above vs. 46 here.

test/functional/apk-utils-e2e-specs.js
```diff
@@ -0,0 +1,65 @@
...
+    });
+    it('should start activity with wait activity', async () => {
+      await adb.install(contactManagerPath);
+      await adb.startApp({pkg: 'com.example.android.contactmanager',
+                          activity: 'ContactManager',
+                          waitActivity: '.ContactManager',
+                          waitDuration: 1000});
```
imurchie: Is there anything that can be asserted here? Or will the `adb.startApp` call fail the test if it breaks?

test/unit/apk-utils-specs.js
```diff
@@ -0,0 +1,188 @@
...
+          let mocks = {};
+          beforeEach(async () => {
+            mocks.adb = sinon.mock(adb);
+          });
+          afterEach(() => {
+            mocks.adb.restore();
+          });
+          fn(mocks);
+        };
+      };
```
imurchie: Indentation?

imurchie:
In general, looks good!

lib/tools/apk-utils.js
moizjv: @jlipps and I discussed this. It ok to modify startAppOptions. What do you think?

lib/tools/apk-utils.js
moizjv: I have mentioned cool down comment for that.

test/functional/apk-utils-e2e-specs.js
moizjv: That was intentional to have wrong activity name. But let me make it `foo`

lib/tools/apk-utils.js
sebv: I don't think it's ok, makes it complicate to track side effects.

test/functional/apk-utils-e2e-specs.js
moizjv: Yes test would break if `startApp` fails. But it is a good idea to add assertion :+1: I have added assertion to all tests of `startApp`.

lib/tools/apk-utils.js
```diff
@@ -0,0 +1,164 @@
...
+      } else {
+        log.errorAndThrow("Activity used to start app doesn't exist or cannot be " +
+                          "launched! Make sure it exists and is a launchable activity");
+      }
+    } else if (stdout.indexOf("java.lang.SecurityException") !== -1) {
+      // if the app is disabled on a real device it will throw a security exception
+      log.errorAndThrow("Permission to start activity denied.");
+    }
+    if (startAppOptions.waitActivity) {
+      if (startAppOptions.hasOwnProperty("waitDuration")) {
```
jlipps: can't we just get rid of this if/else?

```js
await this.waitForActivity(startAppOptions.waitPkg, startAppOptions.waitActivity, startAppOptions.waitDuration);
```

if startAppOptions doesn't have a waitDuration, then it will pass undefined, which is exactly the same thing as not passing it at all.

lib/tools/apk-utils.js
```diff
@@ -0,0 +1,164 @@
...
+  }
+};
+
+
+apkUtilsMethods.getFocusedPackageAndActivity = async function () {
+  log.debug("Getting focused package and activity");
+  let cmd = ['dumpsys', 'window', 'windows'],
+      nullRe = new RegExp(/mFocusedApp=null/),
+      searchRe = new RegExp(
+        /mFocusedApp.+Record\{.*\s([^\s\/\}]+)\/([^\s\/\}]+)(\s[^\s\/\}]+)*\}/);
```
jlipps: if this is long we can break it across multiple lines using quotes. when you use the `RegExp` constructor, you don't need to use `/.../`, that's just for regex literals

lib/tools/apk-utils.js
jlipps: maybe `// cool down so we're not overloading device with requests`

test/unit/apk-utils-specs.js
```diff
@@ -0,0 +1,188 @@
...
+      startAppOptions = {stopApp: true, action: 'action', category: 'cat',
+                         flags: 'flags', pkg: 'pkg', activity: 'act',
+                         optionalIntentArguments: '-x options'},
+      cmd = ['am', 'start', '-n', 'pkg/act', '-S', '-a', 'action', '-c', 'cat',
+             '-f', 'flags', '-x options'];
+
+describe('Apk-utils', () => {
+  let adb = new ADB();
+  let withExecMock = (fn) => {
+        return () => {
```
jlipps: why so many spaces

jlipps:
made my comments

lib/tools/apk-utils.js
moizjv: yes I tried that but i would need to escape all `/` which was looking but odd with too many '//'

PR #84, opened 2015-06-16
Adding last set of functionalities to ADB. 

moizjv:
- Adding more functionalities to ADB.
- Adding more functional and unit tests.
- Restructuring tests to have fixtures. #81
- @jlipps @sebv @imurchie  please review!


test/functional/adb-commands-e2e-specs.js
```diff
@@ -2,12 +2,17 @@ import chai from 'chai';
...
 import ADB from '../../lib/adb.js';
+import path from 'path';
 
 chai.use(chaiAsPromised);
 // change according to CI
 const apiLevel = '21',
       IME = 'com.example.android.softkeyboard/.SoftKeyboard',
-      defaultIME = 'com.android.inputmethod.latin/.LatinIME';
+      defaultIME = 'com.android.inputmethod.latin/.LatinIME',
+      contactManagerPath = path.resolve(__dirname, '..', '..', '..', 'test',
```
sebv: This going to break coverage, you need to use the pachage root helper.

test/functional/adb-commands-e2e-specs.js
```diff
@@ -33,4 +38,31 @@ describe('adb commands', () => {
...
+    (await adb.getPIDsByName(pkgName)).length.should.equal(0);
+  });
+  it('killProcessByPID should kill process', async () => {
+    await adb.install(contactManagerPath);
+    await adb.startApp({pkg: pkgName,
+                        activity: actName});
+    let pids = await adb.getPIDsByName(pkgName);
+    pids.length.should.be.above(0);
+    await adb.killProcessByPID(pids[0]);
+    (await adb.getPIDsByName(pkgName)).length.should.equal(0);
```
sebv: i believe it is should.have.length(0)

test/functional/adb-commands-e2e-specs.js
```diff
@@ -33,4 +38,31 @@ describe('adb commands', () => {
...
+                        activity: actName});
+    await adb.killProcessesByName(pkgName);
+    (await adb.getPIDsByName(pkgName)).length.should.equal(0);
+  });
+  it('killProcessByPID should kill process', async () => {
+    await adb.install(contactManagerPath);
+    await adb.startApp({pkg: pkgName,
+                        activity: actName});
+    let pids = await adb.getPIDsByName(pkgName);
+    pids.length.should.be.above(0);
```
sebv: try should.have.length.above(0)

test/functional/apk-signing-e2e-specs.js
```diff
@@ -7,13 +7,13 @@ import { system } from 'appium-support';
...
 
 const selendroidTestApp = path.resolve(__dirname, '..', '..', '..', 'test',
-                                       'selendroid-test-app.apk'),
+                                       'fixtures', 'selendroid-test-app.apk'),
       contactManagerPath = path.resolve(__dirname, '..', '..', '..', 'test',
-                                        'ContactManager.apk'),
+                                        'fixtures', 'ContactManager.apk'),
       unsignJar = path.resolve(__dirname, '..', '..', '..', 'jars', 'unsign.jar'),
       tmp = system.isWindows() ? 'C:\\Windows\\Temp' : '/tmp',
       keystorePath = path.resolve(__dirname, '..', '..', '..', 'test',
```
sebv: same as above.

test/functional/apk-utils-e2e-specs.js
```diff
@@ -9,7 +9,7 @@ chai.use(chaiAsPromised);
 describe('apk utils', () => {
   let adb = new ADB();
   const contactManagerPath = path.resolve(__dirname, '..', '..', '..', 'test',
```
sebv: same.

test/functional/helpers-specs-e2e-specs.js
```diff
@@ -5,7 +5,8 @@ import { getAndroidPlatformAndPath, assertZipArchive } from '../../lib/helpers.j
 import path from 'path';
 
 const should = chai.should(),
-      apkPath = path.resolve(__dirname, '..', '..', '..', 'test','ContactManager.apk');
```
sebv: ...

test/functional/syscalls-e2e-specs.js
```diff
@@ -29,4 +31,29 @@ describe('System calls', () => {
...
+  });
+  // Skipping for now. Will unskip depending on how it behaves on CI
+  it.skip('launchAVD should get all connected avds', async function () {
+    this.timeout(240000);
+    let proc = await adb.launchAVD(avdName);
+    (await adb.getConnectedEmulators()).length.should.be.above(0);
+    proc.kill();
+  });
+  it('waitForDevice should get all connected avds', async function () {
+    this.timeout(60000);
```
sebv: wow that's a long test.

test/unit/adb-commands-specs.js
```diff
@@ -21,16 +23,32 @@ const apiLevel = '21',
...
 
 describe('adb commands', () => {
   let adb = new ADB();
+  let logcat = new Logcat({
+    adb: adb
+  , debug: false
+  , debugTrace: false
+  });
   describe('shell', () => {
     let withAdbMock = (fn) => {
```
sebv: Can try to make this a helper. withMocks({adb, logcat, cp}, fn)...

test/functional/adb-commands-e2e-specs.js
jlipps: sebv, what is the "package root helper"? I haven't seen this

test/functional/syscalls-e2e-specs.js
jlipps: launching an emu takes a long time :-/

ultimately I don't think it should be Appium's responsibility to do this and we should deprecate this behavior, but since we don't want breaking changes here, we need to include it in this rewrite.

lib/tools/adb-commands.js
```diff
@@ -4,6 +4,9 @@ import { getIMEListFromOutput, isShowingLockscreen, isCurrentFocusOnKeyguard,
 import path from 'path';
 import { fs } from '../utils.js';
 import net from 'net';
+import Logcat from '../logcat.js';
```
jlipps: for our local imports, let's do `from '../logcat';` not `from '../logcat.js';`, here and everywhere

lib/tools/adb-commands.js
```diff
@@ -262,7 +265,207 @@ methods.push = async function (localPath, remotePath) {
...
+  await this.adbExec('forward', [`tcp:${systemPort}`, `localabstract:${devicePort}`]);
+};
+
+methods.ping = async function () {
+  try {
+    let stdout = await this.shell(["echo", "ping"]);
+    if (stdout.indexOf("ping") === 0) {
+      return true;
+    }
+    log.errorAndThrow(`ADB ping failed, returned ${stdout}`);
```
jlipps: since this will be caught in the catch here, we don't need to `errorAndThrow` it, we just need to throw it, it will be logged in the block below

lib/tools/adb-commands.js
```diff
@@ -262,7 +265,207 @@ methods.push = async function (localPath, remotePath) {
...
 
 methods.pull = async function (remotePath, localPath) {
-  await this.exec(['pull', remotePath, localPath]);
+  await this.adbExec(['pull', remotePath, localPath]);
+};
+
+methods.processExists = async function (processName) {
+  try {
+    if (!this.isValidClass(processName)) {
+      log.errorAndThrow(`Invalid process name: ${processName}`);
```
jlipps: same thing with errorAndThrow here

lib/tools/adb-commands.js
```diff
@@ -262,7 +265,207 @@ methods.push = async function (localPath, remotePath) {
...
+  , debugTrace: false
+  });
+  await this.logcat.startCapture();
+};
+
+methods.stopLogcat = async function () {
+  if (this.logcat !== null) {
+    await this.logcat.stopCapture();
+    this.logcat = null;
+  } else {
```
jlipps: this `else` block doesn't do anything and can be removed

lib/tools/adb-commands.js
```diff
@@ -262,7 +265,207 @@ methods.push = async function (localPath, remotePath) {
...
+    let stdout = await this.shell(["ps", name]);
+    stdout = stdout.trim();
+    let pids = [];
+    for (let line of stdout.split("\n")) {
+      if (line.indexOf(name) !== -1) {
+        let match = /[^\t ]+[\t ]+([0-9]+)/.exec(line);
+        if (match) {
+          pids.push(parseInt(match[1], 10));
+        } else {
+          log.errorAndThrow(`Could not extract PID from ps output: ${line}`);
```
jlipps: same thing with errorAndThrow here

lib/tools/adb-commands.js
```diff
@@ -262,7 +265,207 @@ methods.push = async function (localPath, remotePath) {
...
+    log.errorAndThrow(`Unable to get pids for ${name}. Orginial error: ${e.message}`);
+  }
+};
+
+methods.killProcessesByName = async function (name) {
+  try {
+    log.debug(`Attempting to kill all ${name} processes`);
+    let pids = await this.getPIDsByName(name);
+    if (pids.length < 1) {
+      log.errorAndThrow(`No ${name} process found to kill`);
```
jlipps: same thing with errorAndThrow here

lib/tools/adb-commands.js
```diff
@@ -262,7 +265,207 @@ methods.push = async function (localPath, remotePath) {
...
+  try {
+    while ((Date.now() - start) < timeoutMs) {
+      if (await this.processExists(processName)) {
+        // cool down
+        await sleep(400);
+        continue;
+      }
+      return;
+    }
+    log.errorAndThrow(`Process never died within ${timeoutMs} ms`);
```
jlipps: same thing with errorAndThrow here

lib/tools/adb-commands.js
moizjv: yes will change once I am don rewriting logcat :)

lib/tools/adb-commands.js
```diff
@@ -262,7 +265,207 @@ methods.push = async function (localPath, remotePath) {
...
+  return new Promise(async (resolve, reject) => {
+    let args = this.adb.defaultArgs
+      .concat(['shell', 'am', 'instrument', '-e', 'coverage', 'true', '-w'])
+      .concat([instrumentClass]);
+    log.debug(`Collecting coverage data with: ${[this.adb.path].concat(args).join(' ')}`);
+    try {
+      // am instrument runs for the life of the app process.
+      this.instrumentProc = child_process.spawn(this.adb.path, args);
+      this.instrumentProc.on('error', function (e) {
+        reject(log.errorAndThrow(`Failed to run instrumentation. Original error: ${e.message}`));
```
jlipps: woah this is a bit weird. `log.errorAndThrow` throws an error, which  means it does not return an error, which means `reject` will never get run... it should probably be:

```js
reject(new Error(`Failed to run...`));
```
as well  as below

moizjv: I see, good catch !

lib/tools/adb-commands.js
```diff
@@ -262,7 +265,207 @@ methods.push = async function (localPath, remotePath) {
...
+  }
+  return new Promise(async (resolve, reject) => {
+    let args = this.adb.defaultArgs
+      .concat(['shell', 'am', 'instrument', '-e', 'coverage', 'true', '-w'])
+      .concat([instrumentClass]);
+    log.debug(`Collecting coverage data with: ${[this.adb.path].concat(args).join(' ')}`);
+    try {
+      // am instrument runs for the life of the app process.
+      this.instrumentProc = child_process.spawn(this.adb.path, args);
+      this.instrumentProc.on('error', function (e) {
```
jlipps: if you use `SubProcess` this will be easier:

```js
this.instrumentProc = new SubProcess(this.adb.path, args);
await this.instrumentProc.start(0); // this will take care of any errors so you don't have to listen to that event
this.instrumentProc.on('output', (stdout, stderr) => {
  if (stderr) {
    reject(new Error(`Failed to run...`));
  }
});
await this.waitForActivity(waitPkg, waitActivity);
resolve();
```

of course you'll need to then use `await this.instrumentsProc.stop();` instead of `this.instrumentsProc.kill()` elsewhere in the code

lib/tools/adb-commands.js
```diff
@@ -262,7 +265,207 @@ methods.push = async function (localPath, remotePath) {
...
+        reject(log.errorAndThrow(`Failed to run instrumentation. Original error: ${e.message}`));
+      });
+      this.instrumentProc.stderr.on('data', function (data) {
+        reject(log.errorAndThrow(`Failed to run instrumentation: ${new Buffer(data).toString('utf8')}`));
```
jlipps: same bad use of reject + errorAndThrow

lib/tools/adb-commands.js
```diff
@@ -262,7 +265,207 @@ methods.push = async function (localPath, remotePath) {
...
+      });
+      await this.waitForActivity(waitPkg, waitActivity);
+      resolve();
+    } catch (e) {
+      reject(log.errorAndThrow(`Android coverage failed. Original error: ${e.message}`));
```
jlipps: same here

lib/tools/system-calls.js
```diff
@@ -216,4 +217,178 @@ systemCallMethods.getPortFromEmulatorString = function (emStr) {
...
+      let port = this.getPortFromEmulatorString(device.udid);
+      if (port) {
+        device.port = port;
+        emulators.push(device);
+      }
+    }
+    log.debug(`${emulators.length} emulator(s) connected`);
+    return emulators;
+  } catch (e) {
+    log.errorAndThrow(`Error getting emulators. Original error: ${e.message}`);
```
jlipps: I'm starting to wonder whether this pattern (catching all errors and then wrapping them up in a catch block with a more descriptive error) is really what we want to do.

the value that it adds for the user is they get a message saying "error getting emulators" instead of something potentially harder to decipher.

is this worth it?

lib/tools/system-calls.js
```diff
@@ -216,4 +217,178 @@ systemCallMethods.getPortFromEmulatorString = function (emStr) {
...
+    let start = Date.now();
+    while ((Date.now() - start) < timeoutMs) {
+      try {
+        let runningAVD = await this.getRunningAVD(avdName.replace('@', ''));
+        if (runningAVD) {
+          return runningAVD;
+        }
+      } catch (e) {
+        // Do nothing.
+        log.info('Trying to get running AVD');
```
jlipps: this seems like a strange place for this log. since we log in `getRunningAVD`, I don't think we even need this? Could just do `catch (e) {}`, or:

```js
} catch (e) {
  log.info(`Couldn't get running AVD, will retry. Error was: ${e.message}`);
}
```

lib/tools/system-calls.js
```diff
@@ -216,4 +217,178 @@ systemCallMethods.getPortFromEmulatorString = function (emStr) {
...
+    cmd = 'TASKKILL';
+    args = ['TASKKILL' ,'/IM', 'emulator.exe'];
+  } else {
+    cmd = '/usr/bin/killall';
+    args = ['-m', 'emulator*'];
+  }
+  try {
+    await exec(cmd, args);
+  } catch (e) {
+    log.errorAndThrow(`Error killing emulators. Original error: ${e.message}`);
```
jlipps: the other thing I don't like about this pattern is we lose the original error stack. If `exec` was a deep complicated function, all we would get here is the error message, not a stack of exactly where in `exec` it went wrong.

so i'm kind of on the fence. what do you think @sebv?

lib/tools/system-calls.js
```diff
@@ -216,4 +217,178 @@ systemCallMethods.getPortFromEmulatorString = function (emStr) {
...
+  }
+  if (typeof locale === "string") {
+    log.debug(`Setting Android Device Country to ${locale}`);
+    launchArgs.push("-prop", `persist.sys.country=${locale.toUpperCase()}`);
+  }
+  if (typeof avdArgs === "string") {
+    avdArgs = avdArgs.split(" ");
+    launchArgs = launchArgs.concat(avdArgs);
+  }
+  let proc = child_process.spawn(emulatorBinaryPath, launchArgs);
```
jlipps: perhaps also want to use `SubProcess` here?

lib/tools/system-calls.js
```diff
@@ -216,4 +217,178 @@ systemCallMethods.getPortFromEmulatorString = function (emStr) {
...
+  proc.on("error", function (err) {
+    log.error("Unable to start Emulator: " + err.message);
+    // actual error will get caught by getRunningAVDWithRetry
```
jlipps: how would this error get caught by getRunningAVDWithRetry? it's not thrown anywhere...

lib/tools/system-calls.js
```diff
@@ -216,4 +217,178 @@ systemCallMethods.getPortFromEmulatorString = function (emStr) {
...
+  while ((Date.now() - start) < timeoutMs) {
+    try {
+      let stdout = await this.shell(["getprop", "init.svc.bootanim"]);
+      if (stdout.indexOf('stopped') > -1) {
+        return;
+      }
+    } catch (e) {
+      // do nothing
+    }
+    sleep(3000);
```
jlipps: need to `await` this

test/unit/android-manifest-specs.js
```diff
@@ -0,0 +1,76 @@
...
+import ADB from '../../lib/adb.js';
+import sinon from 'sinon';
+import * as teen_process from 'teen_process';
+
+chai.use(chaiAsPromised);
+
+describe('android-manifest', async () => {
+  let adb = new ADB();
+  let withExecMock = (fn) => {
+        return () => {
```
jlipps: why is this indented so much?

jlipps:
ok, made comments!

lib/tools/system-calls.js
moizjv: In that case can we use `e.stack()` too?

test/functional/adb-commands-e2e-specs.js
sebv: Here: https://github.com/appium/appium-adb/commit/353cbf21c97ed253243f8c855656f9610ad0964e
It's just when you use transpile on the fly (for coverage), the build repo is not used, so you get one less `..` in the PATH.

PR #85, opened 2015-06-17
Porting logcat to es6+ also adding functional and unit tests for it

moizjv:
- Logcat with es6+ and teen_process.
- Added android-22 and android-mnc
- @jlipps @sebv please review

### DIFF
--- a/lib/logcat.js
+++ b/lib/logcat.js
@@ -1,118 +1,82 @@
-"use strict";
+import { SubProcess } from 'teen_process';
+import { getLogger } from 'appium-logger';
 
-var spawn = require('child_process').spawn
-  , through = require('through')
-  , EventEmitter = require('events').EventEmitter
-  , util = require('util')
-  , _ = require('lodash')
-  , logger = require('./logger');
+const log = getLogger('Logcat');
 
-var Logcat = function (opts) {
-  EventEmitter.call(this);
-  this.adb = opts.adb;
-  this.debug = opts.debug;
-  this.debugTrace = opts.debugTrace;
-  this.proc = null;
-  this.onLogcatStart = null;
-  this.logcatStarted = false;
-  this.calledBack = false;
-  this.logs = [];
-  this.logsSinceLastRequest = [];
-};
-
-util.inherits(Logcat, EventEmitter);
-
-Logcat.prototype.startCapture = function (cb) {
-  this.onLogcatStart = cb;
-  logger.debug("Starting logcat capture");
-  this.proc = spawn(this.adb.path, this.adb.defaultArgs.concat(['logcat']));
-  this.proc.stdout.setEncoding('utf8');
-  this.proc.stderr.setEncoding('utf8');
-  this.proc.on('error', function (err) {
-    logger.error('Logcat capture failed: ' + err.message);
-    if (!this.calledBack) {
-      this.calledBack = true;
-      cb(err);
-    }
-  }.bind(this));
-  this.proc.on('exit', function (code, signal) {
-    logger.debug('Logcat terminated with code ' + code + ', signal ' + signal);
-    this.proc = null;
-  }.bind(this));
-  this.proc.stdout.pipe(through(this.onStdout.bind(this)));
-  this.proc.stderr.pipe(through(this.onStderr.bind(this)));
-};
-
-Logcat.prototype.stopCapture = function (cb) {
-  logger.debug("Stopping logcat capture");
-  if (this.proc === null) {
-    logger.debug("Logcat already stopped");
-    cb();
-    return;
+class Logcat {
+  constructor (opts = {}) {
+    this.adb = opts.adb;
+    this.debug = opts.debug;
+    this.debugTrace = opts.debugTrace;
+    this.logs = [];
+    this.logsSinceLastRequest = [];
   }
-  this.proc.on('exit', function () {
-    cb();
-  });
-  this.proc.kill();
-  this.proc = null;
-};
 
-Logcat.prototype.onStdout = function (data) {
-  this.onOutput(data, '');
-};
-
-Logcat.prototype.onStderr = function (data) {
-  if (/execvp\(\)/.test(data)) {
-    logger.error('Logcat process failed to start');
-    if (!this.calledBack) {
-      this.calledBack = true;
-      this.onLogcatStart(new Error("Logcat process failed to start"));
-      return;
-    }
+  startCapture () {
+    return new Promise(async (resolve, reject) => {
+      log.debug("Starting logcat capture");
+      this.proc = new SubProcess(this.adb.path, this.adb.defaultArgs.concat(['logcat']));
+      await this.proc.start(0);
+      this.proc.on('exit', (code, signal) => {
+        log.error(`Logcat terminated with code ${code}, signal ${signal}`);
+        this.proc = null;
+      });
+      this.proc.on('lines-stderr', lines => {
+        for (let line of lines) {
+          if (/execvp\(\)/.test(line)) {
+            log.error('Logcat process failed to start');
+            reject(new Error(`Logcat process failed to start. stderr: ${line}`));
+          }
+          this.outputHandler(line, 'STDERR: ');
+        }
+        resolve();
+      });
+      this.proc.on('lines-stdout', lines => {
+        resolve();
+        for (let line of lines) {
+          this.outputHandler(line);
+        }
+      });
+    });
   }
-  this.onOutput(data, ' STDERR');
-};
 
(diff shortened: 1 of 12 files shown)


lib/logcat.js
```diff
@@ -1,118 +1,80 @@
...
-Logcat.prototype.onStderr = function (data) {
-  if (/execvp\(\)/.test(data)) {
-    logger.error('Logcat process failed to start');
-    if (!this.calledBack) {
-      this.calledBack = true;
-      this.onLogcatStart(new Error("Logcat process failed to start"));
-      return;
-    }
+  startCapture () {
+    return new Promise(async (resolve, reject) => {
```
sebv: Async methods already return promises, no need to do that.

lib/logcat.js
```diff
@@ -1,118 +1,80 @@
...
-    }
-  }.bind(this));
-  this.proc.on('exit', function (code, signal) {
-    logger.debug('Logcat terminated with code ' + code + ', signal ' + signal);
+class Logcat {
+  constructor (opts = {}) {
+    this.adb = opts.adb;
+    this.debug = opts.debug;
+    this.debugTrace = opts.debugTrace;
     this.proc = null;
```
sebv: Most of `this.proc = null;` in the PR are unnecessary,

lib/logcat.js
```diff
@@ -1,118 +1,80 @@
...
-    }
-  }
-  data = data.trim();
-  data = data.replace(/\r\n/g, "\n");
-  var logs = data.split("\n");
-  _.each(logs, function (log) {
-    log = log.trim();
-    if (log) {
-      var logObj = {
+  formatOuput (output, prefix) {
```
sebv: This method do more than format output it may be worth doing some cleanup there.

test/unit/android-manifest-specs.js
```diff
@@ -11,17 +11,17 @@ describe('android-manifest', async () => {
   let adb = new ADB();
   let withExecMock = (fn) => {
```
sebv: Same as last time, we could create a helper to do that in one line.

test/unit/logcat-specs.js
```diff
@@ -0,0 +1,54 @@
...
+import * as teen_process from 'teen_process';
+import events from 'events';
+import Logcat from '../../lib/logcat';
+
+chai.use(chaiAsPromised);
+
+describe('logcat', async () => {
+  let adb = {path: 'dummyPath', defaultArgs: []};
+  let logcat = new Logcat({adb: adb, debug: false, debugTrace: false});
+  let withMocks = (fn) => {
```
sebv: same

test/unit/logcat-specs.js
```diff
@@ -0,0 +1,54 @@
...
+    };
+  };
+  describe('startCapture', withMocks((mocks) => {
+    it('should correctly call subprocess and should resolve promise', async () => {
+      let conn = new events.EventEmitter();
+      conn.start = () => { };
+      mocks.teen_process.expects("SubProcess")
+        .once().withExactArgs('dummyPath', ['logcat'])
+        .returns(conn);
+      setTimeout(function () {
```
sebv: Can use nextTick

test/unit/logcat-specs.js
```diff
@@ -0,0 +1,54 @@
...
+      logs.should.have.length.above(0);
+      mocks.teen_process.verify();
+    });
+    it('should correctly call subprocess and should reject promise', async () => {
+      let conn = new events.EventEmitter();
+      conn.start = () => { };
+      mocks.teen_process.expects("SubProcess")
+        .once().withExactArgs('dummyPath', ['logcat'])
+        .returns(conn);
+      setTimeout(function () {
```
sebv: Can do a promise returning function to do that and call it with  `produceLine().done();`, good promise exercise.

sebv:
Few comments, lgtm.

test/unit/android-manifest-specs.js
moizjv: Yeah I was thinking, how I can make it a utility with more flexibility.

test/unit/logcat-specs.js
moizjv: I tried that, but some how does not work the same :(

lib/logcat.js
```diff
@@ -1,118 +1,80 @@
...
-      this.onLogcatStart(new Error("Logcat process failed to start"));
-      return;
-    }
+  startCapture () {
+    return new Promise(async (resolve, reject) => {
+      log.debug("Starting logcat capture");
+      this.proc = new SubProcess(this.adb.path, this.adb.defaultArgs.concat(['logcat']));
+      await this.proc.start(0);
+      this.proc.on('exit', (code, signal) => {
+        log.info(`Logcat terminated with code ${code}, signal ${signal}`);
```
jlipps: I would say this is worthy of a `log.error`

lib/logcat.js
```diff
@@ -1,118 +1,80 @@
...
+        this.proc = null;
+      });
+      this.proc.on('lines-stderr', lines => {
+        for (let line of lines) {
+          if (/execvp\(\)/.test(line)) {
+            log.error('Logcat process failed to start');
+            reject(new Error(`Logcat process failed to start. stderr: ${line}`));
+          }
+          resolve();
```
jlipps: why is this resolve here? is the one in `lines-stdout` not sufficient?

lib/logcat.js
```diff
@@ -1,118 +1,80 @@
...
-  }
-  data = data.trim();
-  data = data.replace(/\r\n/g, "\n");
-  var logs = data.split("\n");
-  _.each(logs, function (log) {
-    log = log.trim();
-    if (log) {
-      var logObj = {
+  formatOuput (output, prefix) {
+    output.trim();
```
jlipps: don't think this is doing anything. from MDN:

> The trim() method returns the string stripped of whitespace from both ends. trim() does not affect the value of the string itself.

lib/logcat.js
```diff
@@ -1,118 +1,80 @@
-"use strict";
+import { SubProcess } from 'teen_process';
+import log from './logger';
```
jlipps: do we want to do this? or maybe we just do:

```js
import { getLogger } from 'appium-logger';
const log = getLogger('Logcat');
```

that way we don't need to do `[LOGCAT STDERR]` below, it will just be blank for stdout and `[STDERR]` for stderr

jlipps:
made a couple notes, overall looks really good!

test/unit/android-manifest-specs.js
sebv: Call it like this, pass a map:

```js
withMock({adb, something}, fn(mocks) {
   ...
]);
```

lib/logcat.js
moizjv: When we stop it legitimately, it still triggers exit i think

lib/logcat.js
moizjv: either can be called first, so promise can be resolved at the earliest.

lib/logcat.js
moizjv: ohh yes it should have been `output = output.trim();`

lib/logcat.js
moizjv: hmm interesting. It is a choice that we have to make to have `ADB` as prefix or `LOGCAT` as prefix or as `ADB [LOGCAT ...` currently it is the third one.

jlipps: right, I'm saying I think we could just have two prefixes, one for adb and one for logcat. from the user's perspective it doesn't matter that logcat is part of the adb package

lib/logcat.js
moizjv: do you mean i should change the name here? how about `outputHandler`?

lib/logcat.js
jlipps: then maybe before we stop it we should remove the exit listener? we're using SubProcess, so when we `await this.proc.stop();` we know that it has stopped after that, so we don't also need to be listening in that case. so what we could do is this to stop:

```js
this.proc.removeAllListeners('exit');
await this.proc.stop();
```

lib/logcat.js
jlipps: i guess what i don't like is that this is calling resolve() for every single line we ever get from stderr.

lib/logcat.js
moizjv: hmm what this intends to do is calling function successfully `awaits` till we resolve or reject the promise.

test/unit/android-manifest-specs.js
moizjv: Ok, let me make all the changes in another pull request. :)

PR #87, opened 2015-06-24
Adding default timeout for adbExec and refactoring tests for CI

moizjv:
- Adding default timeout to adbExec to 20 sec 
- Fixing helper methods
- Setting babel version to exact 5.5.5
- Setting up CI :) 
- Refactoring tests for CI.
- @jlipps @sebv please review


lib/tools/system-calls.js
```diff
@@ -151,6 +151,10 @@ systemCallMethods.adbExec = async function (cmd, opts = {}) {
   if (!cmd) {
     throw new Error("You need to pass in a command to adbExec()");
   }
+  // setting default timeout for each command to prevent infinite wait.
```
sebv: opts.timeout = opts.timeout || 20000;

sebv:
One comment, otherwise lgtm.

moizjv:
@sebv fixed. Thank for pointing that out.

lib/helpers.js
```diff
@@ -83,11 +83,21 @@ function getIMEListFromOutput (stdout) {
 }
 
 function getJavaForOs () {
-  let java = path.resolve(process.env.JAVA_HOME, 'bin', 'java');
-  if (system.isWindows()) {
-    java = java + '.exe';
+  if (process.env.JAVA_HOME) {
```
jlipps: why doesn't this just use `getJavaHome` instead of duplicating the logic?

jlipps:
one comment otherwise lgtm

moizjv:
@jlipps thanks for pointing it out. Fixed.

Issue #91, opened 2015-07-03
Using "optionalIntentArguments" to pass a JSON string argument to an activity fails due to quoting issues through 2 levels of shells (host -> Android)


jeanbaptistelab:
Hi,

I was trying to pass a JSON string to an Android activity using the "optionalIntentArguments" capabilities and was puzzled by the fact that only the first JSON parameter would end up in the activity intent extras.
```python

# This is Python code
optionalIntentArguments = '-e user \'"\'{"username":"selenium","administrator":true,"objectId":"1","email":""}\'"\''
```
which ends up in the appium log as:

>info: [debug] executing cmd: /home/jbl/adt/sdk/platform-tools/adb -s LGDXXXX shell am start -S -a android.intent.action.MAIN -c android.intent.category.LAUNCHER -f 0x10200000 -n xx.xx.xx/xx.xx.xx.Activity  -e user '"'{"username":"selenium","administrator":true,"objectId":"1","mobileUpdatedAt":"2015-07-03T09:31:08.194Z","hide_customers":false,"fullName":"selenium","email":""}'"'

This looks fine, but the extras getString("user") returns `"username":"selenium"` which is wrong...

I immediately suspected quoting and tried various combinations of backslashes, single vs double quotes and couldn't for the life of me figure it out.

I got to a point where this worked when issuing the "am start" command from the ADB shell, but not from my computer's shell (which is similar to the appium scenario).

I searched around and found this, which put me on the right track to solve my issue:

See http://stackoverflow.com/a/29428061/1872442

I modified the _buildStartCmd function in https://github.com/appium/appium-adb/blob/master/lib/adb.js
to quote the entire adb command:
>adb -s LGDXXXX shell am start -S -a android.intent.action.MAIN -c android.intent.category.LAUNCHER -f 0x10200000 -n xx.xx.xx/xx.xx.xx.Activity  -e user '"'{"username":"selenium","administrator":true,"objectId":"1","email":""}'"'

becomes:
>adb -s LGDXXXX shell "am start -S -a android.intent.action.MAIN -c android.intent.category.LAUNCHER -f 0x10200000 -n xx.xx.xx/xx.xx.xx.Activity  -e user '"'{"username":"selenium","administrator":true,"objectId":"1","email":""}'"'"

which works fine and doesn't seem to cause any other issue.

I'm currently working on a pull request for this...

Cheers,

JB

scottdixon:
Seems like this is outdated. Reopen if you're still experiencing issues.

PR #92, opened 2015-07-03
Fix for https://github.com/appium/appium-adb/issues/91 related to quoting when using "optionalIntentArguments" capability

jeanbaptistelab:
…ting when using "optionalIntentArguments" capability

### DIFF
--- a/lib/adb.js
+++ b/lib/adb.js
@@ -1085,7 +1085,7 @@ ADB.prototype.killProcessByPID = function (pid, cb) {
 };
 
 var _buildStartCmd = function (startAppOptions, apiLevel) {
-  var cmd = "am start ";
+  var cmd = "\"am start ";
 
   cmd += startAppOptions.stopApp && apiLevel >= 15 ? "-S" : "";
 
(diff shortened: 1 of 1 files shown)


jlipps:
@moizjv this should be fixed automatically by using teen_process in 1.5, no?

moizjv:
@jlipps yes i want to think so but i would still test it as this seems to be issue were we want to pass JSON as argument to spawn ( `adb shell "am start ... -option someJSON"` )  I will test and get back

jeanbaptistelab:
@jlipps actually, I don't think using teen_process alone would solve this issue, which is really related to the way the "adb shell am start -option <args>" handles the quoting of args. I believe my fix or one of the sort will still be needed...

jlipps:
@jeanbaptistelab are you able to work on a fix (and test) on the 2.0 branch? you'll notice it's quite different in terms of the JS ^_^

jeanbaptistelab:
@jlipps I'll take a look at the 2.0 branch...

jlipps:
@jeanbaptistelab have you had a chance to take a look at this?

PR #93, opened 2015-07-08
Adding getAdbPath method

moizjv:


test/unit/adb-commands-specs.js
```diff
@@ -460,4 +460,7 @@ describe('adb commands', () => {
     adb.isValidClass('some.package/some.package.Activity').index.should.equal(0);
     should.not.exist(adb.isValidClass('illegalPackage#/adsasd'));
   });
+  it('getAdbPath should correctly return adbPath', () => {
+    adb.getAdbPath().should.equal(adb.adb.path);
```
sebv: Would it be better to rename adb.adb to adb.executable?

moizjv: what do you mean? replacing `adb` object in ADB properties with `executables`?

sebv: Yeah change the name. It's strange `adb.adb`.

jlipps:
makes sense :+1:

PR #95, opened 2015-07-10
Adding npmignore with not including build

moizjv:
How many of these are actually present?


Jonahss:
looks good to me

jlipps:
appium-package-master by default makes an npmignore that doesn't include build

PR #97, opened 2015-08-13
Returning localPath with strigns.json in extractStringsFromApk

moizjv:


lib/tools/apk-utils.js
```diff
@@ -163,6 +163,7 @@ apkUtilsMethods.install = async function (apk, replace = true) {
 apkUtilsMethods.extractStringsFromApk = async function (apk, language, out) {
   log.debug("Extracting strings for language: " + (language || "default"));
```
imurchie: String interpolation?

lib/tools/apk-utils.js
```diff
@@ -180,7 +181,8 @@ apkUtilsMethods.extractStringsFromApk = async function (apk, language, out) {
 
   try {
     log.debug("Reading strings from converted strings.json");
-    fileData = await fs.readFile(path.join(out, stringsJson), 'utf8');
+    localPath = path.join(out, stringsJson);
+    fileData = await fs.readFile(localPath, 'utf8');
```
imurchie: It looks like this is a local, home-grown `fs`. It would be good to shift to `appium-support`'s version.

moizjv: i will create an issue for this right now and we can take it up later :)

moizjv:
https://github.com/appium/appium-adb/issues/98

PR #105, opened 2015-08-27
Pass -v threadtime to adb logcat

Hexxeh:


moizjv:
@Hexxeh thanks for that change. I am wondering how much more verbose would it be, if verbose then we would need to pass it as an optional flag. Also please fix the tests, on your local machine you can run `npm run test`

Hexxeh:
It's minimally more verbose:

"D/AndroidRuntime( 2896): >>>>>> AndroidRuntime START com.android.internal.os.RuntimeInit <<<<<<"

vs

"08-27 00:01:19.873 24517 24517 D AndroidRuntime: >>>>>> START com.android.internal.os.RuntimeInit uid 2000 <<<<<<"

Neater, if anything!

moizjv:
@Hexxeh i would suggest you to add it as an optional parameter, and pass it when `startCapture`.

lib/logcat.js
```diff
@@ -15,7 +15,7 @@ class Logcat {
   startCapture () {
     return new Promise(async (resolve, reject) => {
       log.debug("Starting logcat capture");
-      this.proc = new SubProcess(this.adb.path, this.adb.defaultArgs.concat(['logcat']));
+      this.proc = new SubProcess(this.adb.path, this.adb.defaultArgs.concat(['logcat', '-v threadtime']));
```
moizjv: `['logcat', '-v threadtime']` should be `['logcat', '-v', 'threadtime']`

Hexxeh:
How does that find it's way through to startCapture if I'm using the Appium Python API?

moizjv:
@Hexxeh yeah we would need to pass it as a capability or we can implement this https://github.com/appium/appium/issues/5032. We are working towards getting 1.5 out and you would need to wait until beta comes out, some time this week then it should be pretty straight forward to add a cap. I can help you through a simple code walk through of what needs to be get done to add a cap

Hexxeh:
If you could walk me through that, that'd be great. I'm in #appium and #saucelabs on Freenode, by the way.

I just tried to switch to using Appium logs rather than running adb logcat directly, since running it directly was causing us to miss parts of the logs. Guess Appium was stealing them from us. But we need threadtime because we have a bunch of tools that expect the timestamps to be there.

Hexxeh:
This is the default in Android M and newer, by the way. So you might want to specify for consistency anyway, if not you'll get a different log format depending on what Android version you're using.

jlipps:
@moizjv any reason not to merge this?

moizjv:
Yes if this is new default then lets do it!

PR #107, opened 2015-08-29
Added timeout to installFromDevicePath, fixed getActivityRelativeName, fixed setWifiAndData + added tests

scottdixon:
@moizjv @jlipps @imurchie please review


test/unit/helper-specs.js
```diff
@@ -0,0 +1,19 @@
+//import chai from 'chai';
```
jlipps: if we don't need this code, let's just get rid of it instead of comment it out

jlipps:
one small comment then LGTM

moizjv:
good catch !

lib/tools/adb-commands.js
```diff
@@ -250,10 +251,10 @@ methods.setDataState = async function (on) {
 methods.setWifiAndData = async function ({wifi:w, data:d}) {
```
jlipps: interesting use of ES6 destructured assignment in a parameter list! can we make it conform to our style though and do:

```js
{wifi: w, data: d}
```

Though IMO it would look even more elegant as simply `{wifi, data}`

lib/tools/adb-commands.js
```diff
@@ -250,10 +251,10 @@ methods.setDataState = async function (on) {
...
   let wifiOpts = [],
       dataOpts = [];
-  if (w) {
+  if (!_.isUndefined(w)) {
```
jlipps: yep, good catch

test/unit/adb-commands-specs.js
```diff
@@ -254,7 +254,15 @@ describe('adb commands', () => {
         await adb.setWifiAndData({wifi:true});
         mocks.adb.verify();
       });
-      it('should call shell with correct args when only data', async () => {
+      it('should call shell with correct args when turning only wifi off', async () => {
+        mocks.adb.expects("shell")
+          .once().withExactArgs(['am', 'start', '-n', 'io.appium.settings/.Settings',
+                                  '-e', 'wifi', 'off'])
+          .returns("");
+        await adb.setWifiAndData({wifi:false});
```
jlipps: I realize you're just following the surrounding code here, which is great, but all of these should be `{wifi: <value>}` (i.e., with a space after the colon). mind cleaning it all up while you're in here?

lib/tools/adb-commands.js
jlipps: @scottdixon let's fix the styling on this line here (even though it wasn't part of your change)

jlipps:
merged, thanks!

PR #113, opened 2015-09-09
isRealDevice()

scottdixon:
For emulator sessions, `getprop ro.setupwizard.mode` usually returns 'EMULATOR', but sometimes this isn't the case. Fallback to check for 'generic' (as in generic_x86) in `ro.product.device`.

Example output:
https://gist.github.com/scottdixon/99f263431bcbaba100a0
And:
https://gist.github.com/tetsu-koba/1073308

@moizjv @sebv - I know you guys recently worked on this. Would love your input.

### DIFF
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -48,6 +48,20 @@ methods.isValidClass = function (classString) {
   return new RegExp(/^[a-zA-Z0-9\./_]+$/).exec(classString);
 };
 
+methods.isRealDevice = async function () {
+  // Emulator should return 'EMULATOR' for ro.setupwizard.mode
+  // or contain 'generic' in ro.product.device
+  let setupWizardMode = await this.shell(['getprop', 'ro.setupwizard.mode']);  
+  if (/EMULATOR/.exec(setupWizardMode)) {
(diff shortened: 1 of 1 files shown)


moizjv:
ping @bootstraponline

bootstraponline:
I know we have an old issue that discusses this somewhere. The conclusion was this method of detection is completely unreliable. There are a ton of android emulators now so we should be clear about which are supported and which aren't.

bootstraponline:
I found the issue. https://github.com/appium/appium/issues/4723#issuecomment-81986927

The relevant emulators are:

- [Xamarin Player](https://xamarin.com/android-player)
- [Microsoft Visual Studio Emulator](https://www.visualstudio.com/en-us/features/msft-android-emulator-vs.aspx)
- [Genymotion](https://www.genymotion.com/#!/)
- [Google's Android emulator](https://developer.android.com/sdk/)

Unless the `isRealDevice()` method has been confirmed to work on all 4 then I'm not in support of this change. I have some ideas about implementing fallback detection in the issue linked above.

jlipps:
yeah probably a good idea to check these various emulators

PR #117, opened 2015-09-29
Added ability to delete files

kostechko:


moizjv:
@kostechko thanks for pull request, is  [this] (https://github.com/appium/appium-adb/blob/2.0/lib/tools/adb-commands.js#L264) same thing. I have added all methods already available in our brand new [2.0 branch] (https://github.com/appium/appium-adb/tree/2.0) :).

kostechko:
Should've check the other branches before :) That's great actually. Is there some code to work with these methods in appium and clients?

moizjv:
@kostechko appium uses internally, do you want a way to access this method using appium?

kostechko:
Yep. I looked for something like "delete" in appium's android controller and in java_client but didn't find anything.

artkoshelev:
@moizjv when you expect 2.0 to be released?

PR #123, opened 2015-11-04
Undo unzip selection

imurchie:
Turns out that `gunzip` works entirely differently than `unzip`. Decided to just require `unzip` on the system.


imurchie:
Also it can't unzip an `apk`. Argh.

PR #126, opened 2015-12-10
Fix for getting android getting device property for multiple instances

pearlselvan:
Fix for https://github.com/appium/appium/issues/5396

### DIFF
--- a/lib/adb.js
+++ b/lib/adb.js
@@ -193,6 +193,15 @@ ADB.prototype.shell = function (cmd, cb) {
   this.exec(execCmd, cb);
 };
 
+// This is fix for https://github.com/appium/appium/issues/5396 getting the devices propery when runnning with mulitiple appium instance
+ADB.prototype.shellForgettingDeviceProperty = function (cmd, cb) {
+  if (cmd.indexOf('"') === -1) {
+    cmd = '"' + cmd + '"';
+  }
+  var execCmd = '-s '+this.udid+' shell ' + cmd;
+  this.exec(execCmd, cb);
+};
+
 ADB.prototype.spawn = function (args) {
   logger.debug("spawning: " + [this.adb.path].concat(this.adb.defaultArgs, args).join(" "));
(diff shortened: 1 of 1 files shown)


lib/adb.js
```diff
@@ -193,6 +193,15 @@ ADB.prototype.shell = function (cmd, cb) {
   this.exec(execCmd, cb);
 };
 
+// This is fix for https://github.com/appium/appium/issues/5396 getting the devices propery when runnning with mulitiple appium instance
+ADB.prototype.shellForgettingDeviceProperty = function (cmd, cb) {
```
imurchie: `shellForgettingDeviceProperty` ought to be `shellForGettingDeviceProperty` (i.e., "for getting" is different than "forgetting").

lib/adb.js
```diff
@@ -193,6 +193,15 @@ ADB.prototype.shell = function (cmd, cb) {
...
+  if (cmd.indexOf('"') === -1) {
+    cmd = '"' + cmd + '"';
+  }
+  var execCmd = '-s '+this.udid+' shell ' + cmd;
```
imurchie: Add spaces around the `+` signs.

imurchie:
Thanks for this! A few style issues, otherwise it looks good.

jlipps:
thanks @pearlselvan however i think we need to implement it in the 2.0 branch if we want it to be a part of appium moving forward.

pearlselvan:
@jlipps , Thanks for your comments , could your please share is there any things should i take care to implement it in the 2.0 branch because this required to push for running appium in multiple real devices .

@imurchie : Thanks for your comment , currently i am working fixing your comment and will push in case if i get any further comment from @jlipps  regarding my above questions

jlipps:
@pearlselvan can you try this with appium 1.5? i believe appium 1.5 already has fixed this issue because it always includes the `-s <id>` when running adb shell. if so then your fix won't be required there.

imurchie:
All ADB calls have this built in now.

PR #132, opened 2016-01-26
Use multiple possible activities

imurchie:
Instead of trying to guess if the activity needs to be made relative, use a list of possible activity names to check if it has started.

Also improve logging so we have more visibility into what is going on.

See https://github.com/appium/appium/issues/5936


moizjv:
:ship:

PR #134, opened 2016-02-09
Add method to send text to the device

imurchie:
This will be used to set the value immediately on an element, without Ui Automator.


bootstraponline:
:+1:

PR #140, opened 2016-02-18
only pay attention to the 'adbPort' opt

jlipps:
not the 'ANDROID_ADB_SERVER_PORT' env var

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,7 @@
 // transpile:main
 
 import ADB from './lib/adb';
+import { DEFAULT_ADB_PORT } from './lib/adb';
 
 export default ADB;
+export { DEFAULT_ADB_PORT };
--- a/lib/adb.js
+++ b/lib/adb.js
@@ -3,9 +3,9 @@ import path from 'path';
 import methods from './tools/index.js';
 import { rootDir} from './helpers';
 
-
+const DEFAULT_ADB_PORT = 5037;
 const JAR_PATH = path.resolve(rootDir, 'jars');
-const DEFAULTS_OPTS = {
+const DEFAULT_OPTS = {
   sdkRoot: null,
   udid: null,
   appDeviceReadyTimeout: null,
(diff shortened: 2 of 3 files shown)


jlipps:
looks like i broke it.

jlipps:
retest this please

lib/adb.js
```diff
@@ -24,7 +24,8 @@ const DEFAULTS_OPTS = {
   javaVersion: null,
   suppressKillServer: null,
   jars: {},
-  helperJarPath: JAR_PATH
+  helperJarPath: JAR_PATH,
+  adbPort: 5037
```
imurchie: I'd prefer `5037` to be pulled out as a `const`, so it's obvious what the constants are.

imurchie:
Looks good to me, other than the one comment.

jlipps:
updated. however not sure how to get our e2e tests to rerun, any ideas?

imurchie:
They ought to magically start. It takes a little while, though.

jlipps:
ok tests passed. what do you think?

imurchie:
Looks good to me. I think, though, that since the default port is referenced elsewhere (i.e., in android driver) it ought to be exported from this package.

jlipps:
you're right, i was being lazy of course :-)

index.js
```diff
@@ -1,5 +1,7 @@
 // transpile:main
 
 import ADB from './lib/adb';
+import { DEFAULT_ADB_PORT } from './lib/adb';
 
 export default ADB;
+export { DEFAULT_ADB_PORT };
```
imurchie: Cool!

I like the pattern, however, of putting the default into the named exports as well. That way if you just want the default you can use

```js
import ADB from 'appium-adb';
```

but if you want the default and something else you can use

```js
import { ADB, DEFAULT_ADB_PORT } from 'appium-adb';
```

and not 

```js
import { default as ADB, DEFAULT_ADB_PORT } from 'appium-adb';
```

jlipps:
oops, i didn't realized you'd added a new comment and I merged. will add ADB to the named exports too

jlipps:
fixed and published as 2.4.2

jsdevel:
I have a need to pass in a different `adbPort` in capabilities, but it looks like it doesn't do anything right now.  Is there a willingness to add that as the `-P` param?

PR #141, opened 2016-02-18
corrected jar path

scottdixon:
Resolves https://github.com/appium/appium/issues/6102

- Windows was unable to launch Selendroid Server
- Both OSX and Windows were using the same APK
- The APK, however, produced different package names
- Further inspection of each APK showed differences in the AndroidManifest.xml file
- Windows wasn't inserting the modified AndroidManifest.xml because it couldn't find the move_manifest.jar file

CC @imurchie


imurchie:
Good catch! Looks good to me.

PR #143, opened 2016-03-08
Fix adb language methods to work with API 23

rgonalo:
Android 6.0 (API 23) uses 'persist.sys.locale' instead of 'persist.sys.language' and 'persist.sys.country' properties to store device language settings.
Besides, in some devices, 'persist.sys.*' properties could be empty. In these cases, device language settings could be found in 'ro.product.*' properties.

So, I have added some new adb methods to get and set these properties and a new getDeviceLanguage method that returns device language in any API.


imurchie:
Nice! Thank you!

PR #145, opened 2016-03-09
Update getRunningAVD to allow launching Android Emulator

rgonalo:
To fix https://github.com/appium/appium/issues/6233, getRunningAVD method should return null when no avd is connected, instead of throwing an exception.
I have also added locale property in launchAVD method to launch API23 emulators with proper language settings.


imurchie:
LGTM. Thanks!

PR #160, opened 2016-05-31
Correct Regex for deviceNotFoundError

crashbell:
The exact error is `error: device '3801b560347ac200' not found`. So I change the Regex to match with that error.

### DIFF
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -180,7 +180,7 @@ systemCallMethods.adbExec = async function (cmd, opts = {}) {
       return stdout;
     } catch (e) {
       let protocolFaultError = new RegExp("protocol fault \\(no status\\)", "i").test(e);
-      let deviceNotFoundError = new RegExp("error: device not found", "i").test(e);
+      let deviceNotFoundError = new RegExp("error: device ('.+' )?not found", "i").test(e);
       if (protocolFaultError || deviceNotFoundError) {
         log.info(`error sending command, reconnecting device and retrying: ${cmd}`);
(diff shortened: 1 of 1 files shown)


imurchie:
Thanks for this! The regex does seem to need to be that way. Are we sure that for earlier versions of adb this was always the case? The regex could be made to handle the old behaviour as well as the new (i.e., catch `error: device 'blah-blah-blah' not found` as well as `error: device not found`)

https://regex101.com/r/rQ1pI7/1

crashbell:
@imurchie your regex makes more sense. I update the code.

imurchie:
Excellent! There is no need, however, for the escaping around the single quotes.

crashbell:
I have removed them.

imurchie:
Weird. jshint clearly is confused. But since it seems that it will only be a space after the device name, the `\s` could just be a single space (i.e., `let deviceNotFoundError = new RegExp("error: device ('.+' )?not found", "i").test(e);`) You can run, on the command line, `gulp jshint` to see if it passes linting.

Sorry for the mix up. :(

crashbell:
I updated it, now it passed the jshint checker.

imurchie:
Functional tests are good to go, too: https://team-appium.ci.cloudbees.com/view/Appium%20Libs/job/adb-e2e-android/159/console

imurchie:
Thanks for fixing this! And for your patience!

crashbell:
@imurchie Thanks for reviewing this :)

imurchie:
Published as `appium-adb@2.6.3`

PR #164, opened 2016-06-14
Add method to clear a text field using adb

imurchie:
Android text field clearing is a problem. `adb` can send a key event to delete, and can do it quickly and reliably. Add a method to do that so we can offer a better way to clear, if possible.


imurchie:
Build: https://team-appium.ci.cloudbees.com/view/Appium%20Libs/job/adb-e2e-android/163/

jlipps:
LGTM

PR #168, opened 2016-08-17
Use adbPort capability.

jsdevel:


lib/tools/system-calls.js
```diff
@@ -90,7 +90,7 @@ systemCallMethods.getBinaryFromPath = async function (binaryName) {
 systemCallMethods.getConnectedDevices = async function () {
   log.debug("Getting connected devices...");
   try {
-    let {stdout} = await exec(this.executable.path, ['devices']);
+    let {stdout} = await exec(this.executable.path, this.executable.defaultArgs.concat('devices'));
```
imurchie: If this is an adb command we are executing, why not make use of `adbExec`?

imurchie:
You can run the tests, to get them working, with `gulp once`.

lib/tools/system-calls.js
jsdevel: Sounds good.  I'll look at it further tonight.

PR #171, opened 2016-09-06
Fixed issues with Android 7 (issue #6665 in Appium)

cschabetsberger-micro-focus:
* Adjust reading of the process id with a given name to also make it work on Android 7 by using a combination of "ps" and "grep" commands because the api of "ps" has changed
* If an adb exec command fails, pass the stderr or the stdout (if possible) to the caller (because failing adb exec commands on Android 7 may throw an exception while on Android 6 they did not; instead, on Android 6 the exception text was returned in stdout)

### DIFF
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -424,7 +424,7 @@ methods.getPIDsByName = async function (name) {
     if (name.length > 15) {
       name = name.substr(name.length - 15);
     }
-    let stdout = await this.shell(["ps", name]);
+    let stdout = await this.shell(["set `ps | grep '" + name + "'`;echo $2"]);
     stdout = stdout.trim();
     let pids = [];
     for (let line of stdout.split("\n")) {
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -165,6 +165,7 @@ systemCallMethods.adbExec = async function (cmd, opts = {}) {
   // setting default timeout for each command to prevent infinite wait.
   opts.timeout = opts.timeout || DEFAULT_ADB_EXEC_TIMEOUT;
   let execFunc = async () => {
+    let linkerWarningRe = /^WARNING: linker.+$/m;
     try {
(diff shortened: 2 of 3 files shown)


imurchie:
One unit test is failing. Can you run `gulp once` locally to ry to fix?

imurchie:
How great that the API changes! Thanks for taking this on. We really appreciate it!

lib/tools/adb-commands.js
```diff
@@ -424,7 +424,7 @@ methods.getPIDsByName = async function (name) {
     if (name.length > 15) {
       name = name.substr(name.length - 15);
     }
-    let stdout = await this.shell(["ps", name]);
+    let stdout = await this.shell(["set `ps | grep '" + name + "'`;echo $2"]);
```
imurchie: This can use string interpolation to be cleaner:

```js
`set \`ps | grep '${name}'\`;echo $2`
```

cschabetsberger-micro-focus:
I was trying to use string interpolation as suggested but I am receiving syntax errors in jshint when running gulp once (so it seems that \` does not work to escape the string interpolation here). Any suggestions?

For now, I have updated the unit test for adb-commands to work with the version that does not use string interpolation.

imurchie:
Ah. Jshint has some issues of its own. :| 

You can leave as is, or add a comment to skip the line (http://stackoverflow.com/a/19635321/6214184).

Either way, can you squash into a single commit, and we will get it in. Thanks for all the work on this!

jlipps:
@cschabetsberger-micro-focus thanks for your efforts with this fix!

cschabetsberger-micro-focus:
I also wanted to thank you for the assistance on this!

imurchie:
Published in `appium-adb@2.6.13`.

Kalyan1989:
We have a new GUI tool that you can download at: https://github.com/appium/appium-desktop/releases/latest

jassid1:
Hi I am still getting this above error. any help?

PR #176, opened 2016-10-05
Remove -g from install, which breaks stuff

imurchie:
The work done to add the `-g` flag to install breaks more than it fixes (which seems to be nothing).


jlipps:
😢 🐼

PR #186, opened 2016-11-30
grant and revoke app permission for Android 23 and above

sravanmedarapu:
Fix for #https://github.com/appium/appium/issues/5561

cc: @jlipps,  @imurchie

PR #187, opened 2016-12-03
refactored targetSdkVersionFromManifest 

sravanmedarapu:
PR https://github.com/appium/appium-adb/pull/186 not working with some applications.
Issue: #https://github.com/appium/appium/issues/7349, #https://github.com/appium/appium/issues/7353
- Granting permissions at run time works only when both below 2 cases matched
    
1.  Device API should be 23 and above (From device point of view)
2.  Application's `targetSdkVersion` version should be 23 and above (From Application point of view)

But there is no direct way to fetch `targetSdkVersion` from application's APK. Here we are relying on dumping apk info using (`aapt dump badging APK`). But unfortunately in legacy applications we do not find targetSdkVersion info. 

Till we find the better alternatives to figure out `targetSdkVersion`, removing the `targetSdkVersion` to grant permissions.

cc: @jlipps, @imurchie

### DIFF
--- a/lib/tools/android-manifest.js
+++ b/lib/tools/android-manifest.js
@@ -96,9 +96,12 @@ manifestMethods.targetSdkVersionFromManifest = async function (localApk) {
     let args = ['dump', 'badging', localApk];
     let {stdout} = await exec(this.binaries.aapt, args);
     let targetSdkVersion = new RegExp(/targetSdkVersion:'([^']+)'/g).exec(stdout);
+    if (!targetSdkVersion) {
+      throw new Error(`targetSdkVersion is not specified in the application.`);
+    }
     return parseInt(targetSdkVersion[1], 10);
   } catch (e) {
-    log.errorAndThrow(`targetSdkVersionFromManifest failed. Original error: ${e.message}`);
+    log.errorAndThrow(`fetching targetSdkVersion from local APK failed. Original error: ${e.message}`);
   }
 };
 
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -193,8 +193,14 @@ apkUtilsMethods.install = async function (apk, replace = true, timeout = 60000) 
   }
   if (apk.includes('.apk')) {
     let apiLevel = await this.getApiLevel();
-    let targetSdk = await this.targetSdkVersionFromManifest(apk);
-    if (apiLevel >= 23 && targetSdk >= 23) {
+    let targetSdk = null;
+    try {
+      targetSdk = await this.targetSdkVersionFromManifest(apk);
+    } catch (e) {
+      //avoiding logging error stack, as calling library function would have logged
+      log.warn(`Ran into problem getting target SDK version; ignoring...`);
+    }
+    if (apiLevel >= 23 && (!targetSdk || targetSdk >= 23)) {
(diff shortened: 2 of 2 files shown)


lib/tools/android-manifest.js
```diff
@@ -89,18 +89,6 @@ manifestMethods.packageAndLaunchActivityFromManifest = async function (localApk)
   }
 };
 
-manifestMethods.targetSdkVersionFromManifest = async function (localApk) {
```
mykola-mokhnach: I'd rather keep it. This method should be simply fixed by checking the the _android:minSdkVersion_ value (it is mandatory to set) in case _ targetSdkVersion_ is omitted in the manifest file. Check https://developer.android.com/guide/topics/manifest/uses-sdk-element.html for more details.

sravanmedarapu: @mykola-mokhnach, Actually this method add to implement to grant permissions.

As I mentioned in this [PR](https://github.com/appium/appium-adb/pull/187), grant permission feature depends on API versions. 

In legacy developed applications do not have these properties respected.

jlipps:
@sravanmedarapu how old does a target application have to be for this to fail?

sravanmedarapu:
@jlipps, typically it is about best practices. All manifest would typically have either of the properties(targetSdkVersion, minSdkVersion). But there are cases where these properties neglected. 

In Google docs mentioned that, it is certain to declare these properties and generally all applications does have these.

sravanmedarapu:
What we could do in the absence of either of these properties, we can continue to grant the permissions. But need sometime to test this different apks and API combinations. As module broken in current state, I think we can continue with condition skip as a hot fix.

lib/tools/android-manifest.js
jlipps: another option would be to attempt to retrieve the target sdk version, but simply ignore the error in the calling function. something like this:

```js
let targetSdk = null;
try {
  targetSdk = await this.targetSdkVersionFromManifest(apk);
} catch (e) {
  logger.warn(`Ran into problem getting target SDK; ignoring: ${e.message}`);
}
if (apiLevel >= 23 && (!targetSdk || targetSdk >= 23)) {
```

lib/tools/android-manifest.js
```diff
@@ -98,7 +98,8 @@ manifestMethods.targetSdkVersionFromManifest = async function (localApk) {
     let targetSdkVersion = new RegExp(/targetSdkVersion:'([^']+)'/g).exec(stdout);
     return parseInt(targetSdkVersion[1], 10);
   } catch (e) {
-    log.errorAndThrow(`targetSdkVersionFromManifest failed. Original error: ${e.message}`);
+    log.warn(`Ran into problem getting target SDK; ignoring: ${e.message}`);
```
jlipps: This isn't exactly what I had in mind.

What you've done here is to make `targetSdkVersionFromManifest` emit a warning instead of throwing an error.

But I think this function _should_ throw an error if something fails.

It's in the _calling_ function below that I think we should explicitly ignore the error. In other words, `targetSdkVersionFromManifest` is a library function, and as such it should indeed throw any errors it encounters. But based on how we are using it in `install` here, we may or may not care about the error. That in my mind is a more flexible structure.

What do you think?

jlipps:
published in 2.9.1

PR #189, opened 2016-12-09
refactor - granting permissions

sravanmedarapu:


lib/tools/adb-commands.js
```diff
@@ -69,11 +69,30 @@ methods.clear = async function (pkg) {
...
+methods.grantAllPermissions = async function (pkg, apk) {
+    let apiLevel = await this.getApiLevel();
+    let targetSdk = null;
+    try {
+      targetSdk = await this.targetSdkVersionFromManifest(apk);
+    } catch (e) {
+      //avoiding logging error stack, as calling library function would have logged
+      log.warn(`Ran into problem getting target SDK version; ignoring...`);
+    }
+    if (apiLevel >= 23 && targetSdk >= 23) {
```
jlipps: doesn't this need to have `(!targetSdk || targetSdk >= 23)`?

sravanmedarapu: Tried with different combinations with and with out having target sdk specified in app, grant permission respected only when application defined with targetSdk.

Official Google documentation also mention same
https://developer.android.com/training/permissions/requesting.html

lib/tools/adb-commands.js
```diff
@@ -73,10 +73,14 @@ methods.grantAllPermissions = async function (pkg, apk) {
...
-      targetSdk = await this.targetSdkVersionFromManifest(apk);
+      if (apk === undefined || apk === null) {
+        targetSdk = await this.targetSdkVersionUsingPKG(pkg);
+      } else {
+        targetSdk = await this.targetSdkVersionFromManifest(apk);
+      }
     } catch (e) {
       //avoiding logging error stack, as calling library function would have logged
-      log.warn(`Ran into problem getting target SDK version; ignoring...`);
+      log.warn(`Ran into problem getting target SDK version; ignoring...`+e.message);
```
jlipps:
```
log.warn(`Ran into problem getting target SDK version; ignoring... ${e.message}`);
```

lib/tools/adb-commands.js
```diff
@@ -73,10 +73,14 @@ methods.grantAllPermissions = async function (pkg, apk) {
     let apiLevel = await this.getApiLevel();
     let targetSdk = null;
     try {
-      targetSdk = await this.targetSdkVersionFromManifest(apk);
+      if (apk === undefined || apk === null) {
```
jlipps:
```js
if (!apk) {
```

lib/tools/android-manifest.js
```diff
@@ -105,6 +105,17 @@ manifestMethods.targetSdkVersionFromManifest = async function (localApk) {
...
 
+manifestMethods.targetSdkVersionUsingPKG = async function (pkg) {
+  let stdout =  await this.shell(['dumpsys', 'package', pkg]);
+  let targetSdkVersion = new RegExp(/targetSdk=([^\s\s]+)/g).exec(stdout);
+  if (targetSdkVersion && targetSdkVersion.length >= 2) {
+    targetSdkVersion = targetSdkVersion[1];
+  } else {
+    targetSdkVersion = null;
+  }
+  return parseInt(targetSdkVersion, 10);
```
jlipps: `parseInt(null, 10)` returns `NaN` which is probably not what we mean. I'd either return `null` explicitly or return 0.

lib/tools/adb-commands.js
jlipps: ok makes sense

mykola-mokhnach: shouldn't this look like 

```
apiLevel >= 23 && targetSdk && targetSdk >= 23
```

in such case? What is the result of null >= 23 comparison?

jlipps: `null >= 23` is `false` so we should be ok

jlipps:
@sravanmedarapu do functional and unit tests pass with these changes?

jlipps:
aha, I had the feeling we might need a test update :-)

jlipps:
@sravanmedarapu if you've run all the unit and functional tests locally, i'm fine to not wait for travis.

sravanmedarapu:
travis execution from my forked [repo](https://travis-ci.org/sravanmedarapu/appium-adb/builds/182727085) passing  and functional tests are also passing .
I think we can proceed.

jlipps:
ok, publishing in 2.9.2

sravanmedarapu:
will update in other drivers.

sravanmedarapu:
#https://github.com/appium/appium/issues/7493

PR #190, opened 2016-12-27
Correct the getConnectedDevices function

trantrungtin:
Handle a new case that there is a message "adb server version (32) doesn't match this client (36); killing..." when the running adb doesn't match the appium-adb.

Ex:
adb.exe devices
List of devices attached
adb server version (32) doesn't match this client (36); killing...
\* daemon started successfully \*
0816f3699208ee6d        device


trantrungtin:
CC: @jlipps, @imurchie

imurchie:
Published in `appium-adb@2.9.3`.

PR #191, opened 2017-01-03
Add new platforms to list

imurchie:
Platform list now includes latest platforms. Also correct a few log formatting issues that have been bugging me.


lib/tools/system-calls.js
```diff
@@ -460,7 +460,9 @@ systemCallMethods.reboot = async function () {
     if (booted === '1') {
       return;
     } else {
-      log.errorAndThrow('Waiting for reboot this takes time');
+      let msg = 'Waiting for reboot. This takes time';
+      log.debug(msg);
+      throw new Error(msg);
```
mykola-mokhnach: I think we have something like logAndThrow

lib/tools/apk-utils.js
```diff
@@ -18,7 +18,7 @@ apkUtilsMethods.isAppInstalled = async function (pkg) {
     let apkInstalledRgx = new RegExp(`^package:${pkg.replace(/(\.)/g, "\\$1")}$`,
                                      'm');
     installed = apkInstalledRgx.test(stdout);
-    log.debug(`App is ${!installed ? " not" : ""} installed`);
+    log.debug(`App is${!installed ? " not" : ""} installed`);
```
mykola-mokhnach: Is there any common rule about single/double quotes usage?

lib/tools/system-calls.js
imurchie: This is explicitly _not_ using that (see the diff). Otherwise you get a stack trace for every throw, and the log looks like it's full of errors instead of expected behaviour..

lib/tools/apk-utils.js
imurchie: No.

PR #193, opened 2017-01-06
Make tests runnable in all API levels

imurchie:
So we can begin to test fully.


test/functional/setup.js
```diff
@@ -0,0 +1,38 @@
+const avdName = process.env.ANDROID_AVD || 'NEXUS_S_18_X86';
+const platformVersion = process.env.PLATFORM_VERSION || '4.3';
+
+let apiLevel = process.env.API_LEVEL;
+if (!apiLevel) {
+  // https://developer.android.com/guide/topics/manifest/uses-sdk-element.html
+  switch (parseFloat(platformVersion)) {
```
jlipps: this seems a little verbose, maybe something lke:

```js
if (!apiLevel) {
  let levelMap = {
    '4.1': '16',
    '4.2': '17',
    '4.3': '18',
    '4.4': '19',
    '5': '21',
    '5.1': '22',
    '6': '23',
    '7': '24',
    '7.1': '25',
  };
  apiLevel = levelMap[parseFloat(platformVersion).toString()];
}
```

?

test/functional/apk-utils-e2e-specs.js
```diff
@@ -95,18 +95,4 @@ describe('apk utils', function () {
     let {apkStrings} = await adb.extractStringsFromApk(contactManagerPath, null, '/tmp');
     apkStrings.save.should.equal('Save');
   });
-  it('should get device language and country', async () => {
```
jlipps: these tests are duplicated above?

imurchie: Yes.

test/functional/setup.js
imurchie: Makes sense. I've updated a little differently, but same.

PR #195, opened 2017-01-17
added utility methods for fetching model, manufacturer and screen size 

sravanmedarapu:
Req Ticket: #https://github.com/appium/appium/issues/3210

cc: @jlipps, @imurchie

### DIFF
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -674,4 +674,21 @@ methods.getDeviceProductLocale = async function () {
   return await this.getDeviceProperty("ro.product.locale");
 };
 
+methods.getModel = async function () {
+  return await this.getDeviceProperty("ro.product.model");
+};
+
+methods.getManufacturer = async function () {
+  return await this.getDeviceProperty("ro.product.manufacturer");
+};
+
+methods.getScreenSize = async function() {
+  let stdout = await this.shell(['wm', 'size']);
+  let size = new RegExp(/Physical size: ([^\r?\n]+)*/g).exec(stdout);
+  if (size && size.length >= 2) {
+    return size[1].trim();
+  }
+  return null;
+};
+
 export default methods;
--- a/test/unit/adb-commands-specs.js
+++ b/test/unit/adb-commands-specs.js
@@ -31,7 +31,10 @@ const apiLevel = '21',
       flags=0x0`,
       psOutput = `USER     PID   PPID  VSIZE  RSS     WCHAN    PC   NAME
 u0_a101   5078  3129  487404 37044 ffffffff b76ce565 S com.example.android.contactmanager`,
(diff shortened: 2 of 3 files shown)


lib/tools/adb-commands.js
```diff
@@ -674,4 +674,37 @@ methods.getDeviceProductLocale = async function () {
   return await this.getDeviceProperty("ro.product.locale");
 };
 
+methods.getModel = async function () {
+  let stdout = await this.shell('getprop');
+  let model = new RegExp(/ro.product.model]:([^\r?\n]+)*/g).exec(stdout);
+  if (model && model.length >= 2) {
```
mykola-mokhnach: I think it would be more convenient to rewrite the condition this way:

```javascript
if (model && model.length >= 2) {
  return model[1].trim().replace(/[\[\]]/g, '');
}
return null;
```

Same to the methods below

sravanmedarapu: Sure, Will update 👍

imurchie: Can you also add a comment about what the replace is doing? Otherwise one needs to examine it to know that you're removing square brackets.

lib/tools/adb-commands.js
```diff
@@ -674,4 +674,31 @@ methods.getDeviceProductLocale = async function () {
   return await this.getDeviceProperty("ro.product.locale");
 };
 
+methods.getModel = async function () {
+  let stdout = await this.shell('getprop');
```
imurchie: Why not be more specific, so the regexp is not necessary? `adb shell getprop ro.product.model` will just return the model name, no?

lib/tools/adb-commands.js
```diff
@@ -674,4 +674,31 @@ methods.getDeviceProductLocale = async function () {
...
+  let model = new RegExp(/ro.product.model]:([^\r?\n]+)*/g).exec(stdout);
+  if (model && model.length >= 2) {
+    return model[1].trim().replace(/[\[\]]/g, '');
+  }
+  return null;
+};
+
+methods.getManufacturer = async function () {
+  let stdout = await this.shell('getprop');
```
imurchie: Also could use full command.

test/unit/adb-commands-specs.js
```diff
@@ -576,6 +576,26 @@ describe('adb commands', () => {
       });
     }));
   });
+  describe('device info', withMocks({adb}, (mocks) => {
+    it('should get device model', async () => {
+      mocks.adb.expects("shell")
+          .once().withArgs('getprop');
+      await adb.getModel();
```
imurchie: For all of these, the mock ought to return expected results, so the method's logic is tested.

imurchie:
Published in `appium-adb@2.10.0`

PR #196, opened 2017-01-20
Enable to set a wildcard value in appWaitActivity.

gleemitsu:
Fixes appium/appium#7607
When an app contains multiple activities, it's difficult to target correct activity names.
This PR adds an ability to set appWaitActivity value containing wildcards.

### DIFF
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -1,4 +1,4 @@
-import { buildStartCmd, getPossibleActivityNames } from '../helpers.js';
+import { buildStartCmd } from '../helpers.js';
 import { exec } from 'teen_process';
 import log from '../logger.js';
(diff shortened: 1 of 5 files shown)


lib/tools/apk-utils.js
```diff
@@ -1,4 +1,4 @@
-import { buildStartCmd, getPossibleActivityNames } from '../helpers.js';
```
imurchie: Since `getPossibleActivityNames` is no longer used here, it needs to be pulled from the codebase altogether. This includes in `lib/helpers.js` and `test/unit/helper-specs.js`.

gleemitsu:
Thank you for review!
I've cleaned up.

imurchie:
Published in `appium-adb@2.11.0`.

PR #198, opened 2017-01-25
Change 'get AVD list command' in systemCallMethods.checkAvdExist

whirosan:
When I want to E2E test Android app with Google Play Services, appium-adb can't recognize AVD with Google API, test failed.
checkAvdExist `android list avd -c` command does not return AVD list with Google API.
So I change 'android list avd -c' to 'emulator -list-avds'.

### DIFF
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -411,11 +411,25 @@ systemCallMethods.launchAVD = async function (avdName, avdArgs, language, countr
 };
 
 systemCallMethods.checkAvdExist = async function (avdName) {
-  let cmd = await this.getSdkBinaryPath('emulator');
-  let args = ['-list-avds'];
-  let {stdout} = await exec(cmd, args);
-  if (stdout.indexOf(avdName) === -1) {
-    let existings = `(${stdout.trim().replace(/[\n]/g, '), (')})`;
+  let cmd, args, result;
+  try {
+    cmd = await this.getSdkBinaryPath('emulator');
+    args = ['-list-avds'];
+    result = await exec(cmd, args);
+  } catch (e) {
+    let unknownOptionError = new RegExp("unknown option: -list-avds", "i").test(e.stderr);
+    if (!unknownOptionError) {
+      log.errorAndThrow(`Error executing checkAvdExist. Original error: '${e.message}'; ` +
+                        `Stderr: '${(e.stderr || '').trim()}'; Code: '${e.code}'`);
+
+    }
+    // If -list-avds option is not available, use android command as an alternative
+    cmd = await this.getSdkBinaryPath('android');
+    args = ['list', 'avd', '-c'];
+    result = await exec(cmd, args);
+  }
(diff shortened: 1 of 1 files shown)


jlipps:
@vrunoa were you running into this recently too?

I think this change is OK, but I'm not entirely sure of the ramifications or if there are any crucial differences between these two commands. Is this supported in `emulator` binaries for all the different SDK levels?

vrunoa:
@jlipps yes I got that issue on osx recently but apparently was a problem with some google sdk version release that they fix in a newer version since it's working for me now.
 
I always use `emulator -list-avds`, no idea what's the difference between those to commands and 
I think it's supported in older sdk versions. I can download the most recent not deprecated version of the android sdk platforms-tools and confirm this if you want.

jlipps:
Thanks! @whirosan let's keep this open for a bit and meanwhile can you try the workaround @vrunoa suggested to see if helps?

vrunoa:
@jlipps @whirosan I thinks this might be a feature of android emulators 2.0 since I try to run this command with an older version of android tools with emulators 1.0 and it's not working. 
```bash
./emulator -list-avds
unknown option: -list-avds
please use -help for a list of valid options
```

This command `emulator -list-avds` it's in the official command-line (documentation)[https://developer.android.com/studio/run/emulator-commandline.html] so maybe my theory is right, but I cannot find any release notes that con confirm this. [https://developer.android.com/studio/releases/sdk-tools.html](https://developer.android.com/studio/releases/sdk-tools.html)

Emus 2.0 were release on December 10, 2015
[https://android-developers.googleblog.com/2015/12/android-studio-20-preview-android.html](https://android-developers.googleblog.com/2015/12/android-studio-20-preview-android.html)

If a user did not move to emus 2.0 this change will probably affect their tests.

jlipps:
Thanks for the research @vrunoa. I was afraid of something like this. It means I can't accept this PR as is, since it would probably be a breaking change for some subset of users. @whirosan if you wanted to implement a fallback strategy that tried `emulator` first and `adb` second, I would consider that!

whirosan:
Thanks! @jlipps @vrunoa 
I understand that this change will have the effect of some users failing the test.
I will consider implementing a fallback strategy with this PR.

whirosan:
@jlipps @vrunoa I implements fallback strategy.
If `emulator -list-avds` is not available, use `android list avd -c` as alternative.
Please review.

whirosan:
@imurchie Thanks!
@jlipps I downgraded sdk tools to version 24.0.1(`emulator -list-avds` unknown option), tested it and confirmed that the test passed.
However, I have not tested the emulator binary to another location.
If you move the binaries of the emulator to another location, `checkAvdExist` and `launchAvd` will also fail.

jlipps:
Riiiiight. Makes sense. OK, I think we're good to go here then. Thanks!

jlipps:
published in `appium-adb@2.11.1`

PR #201, opened 2017-03-03
Adding ability to check SDK version to user avdmanager or android to …

dmendes-daitan:
…find AVD

Since android SDK 25 the avd search now is made with avdmanager instead
of android. Me and @aquintiliano made a method to check the SDK version
and if it’s > 25 will use the avdmanager instead of android

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -4,7 +4,8 @@ import log from './logger.js';
 import AdmZip from 'adm-zip';
 import { exec } from 'teen_process';
 import _ from 'lodash';
-
+import readline from 'readline';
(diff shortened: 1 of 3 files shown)


mykola-mokhnach:
Dear @dmendes-daitan Thanks for your PR, but there is already [one](https://github.com/appium/appium-adb/pull/200), which is very similar to yours. Could you please join to that one so we don't do the same job twice?

dmendes-daitan:
@mykola-mokhnach , the guy who I added on my comment, is the same who did the PR, I've made without knowing he already did it. Let's keep with his. Thanks !

PR #202, opened 2017-03-06
Modify the comments desctiption of system-calls.js

harlentan:
The comments use the word "stupid" will not make many people feel very well.
It seems that there is something wrong with the adb


imurchie:
The "stupid" comment has nothing to do with people. It is commenting on `adb`. But thanks for the submission.

PR #204, opened 2017-03-10
Use avdmanager for Android SDK 25+

mykola-mokhnach:
People wanted it - we did it.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -209,8 +209,28 @@ function buildStartCmd (startAppOptions, apiLevel) {
   return cmd;
 }
 
+async function getSdkToolsVersion () {
+  const androidHome = process.env.ANDROID_HOME;
+  if (!androidHome) {
+    log.errorAndThrow('ANDROID_HOME environment varibale is expected to be set');
+  }
+  const propertiesPath = path.resolve(androidHome, 'tools', 'source.properties');
+  if (!await fs.exists(propertiesPath)) {
+    log.warn(`Cannot find ${propertiesPath} file to read SDK version from`);
+    return;
+  }
+  const propertiesContent = await fs.readFile(propertiesPath, 'utf8');
+  const versionMatcher = new RegExp(/Pkg\.Revision=(\d+)\.?(\d+)?\.?(\d+)?/);
+  const match = versionMatcher.exec(propertiesContent);
+  if (match) {
+    return {major: parseInt(match[1], 10),
+            minor: match[2] ? parseInt(match[2], 10) : 0,
+            build: match[3] ? parseInt(match[3], 10) : 0};
+  }
+  log.warn(`Cannot parse "Pkg.Revision" value from ${propertiesPath}`);
+}
 
 export { getDirectories, getAndroidPlatformAndPath, unzipFile, assertZipArchive,
          getIMEListFromOutput, getJavaForOs, isShowingLockscreen, isCurrentFocusOnKeyguard,
          getSurfaceOrientation, isScreenOnFully, buildStartCmd, getJavaHome,
-         rootDir, androidPlatforms };
+         rootDir, androidPlatforms, getSdkToolsVersion };
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -2,7 +2,7 @@ import path from 'path';
 import log from '../logger.js';
 import B from 'bluebird';
 import { system, fs } from 'appium-support';
-import { getDirectories } from '../helpers';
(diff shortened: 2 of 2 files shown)


mykola-mokhnach:
@xuanzhaopeng can you please incorporate this patch to your local source and try if it works for you?

lib/helpers.js
```diff
@@ -209,8 +209,29 @@ function buildStartCmd (startAppOptions, apiLevel) {
...
+  if (!await fs.exists(propertiesPath)) {
+    log.warn(`Cannot find ${propertiesPath} file to read SDK version from`);
+    return;
+  }
+  const propertiesContent = await fs.readFile(propertiesPath, 'utf8');
+  const versionMatcher = new RegExp(/Pkg\.Revision=(\d+)\.?(\d+)?\.?(\d+)?/);
+  const match = versionMatcher.exec(propertiesContent);
+  // _.toInteger(undefined) returns zero, which is OK
+  if (match) return {
+    major: _.toInteger(match[1]),
```
xuanzhaopeng: I get an error here : Encountered internal error running command: TypeError: _lodash2.default.toInteger is not a function. 

I use lodash 3.10.1

mykola-mokhnach: added fix. lodash version should be at least 4.0.0

imurchie:
You can just use `gulp transpile` to transpile and skip running the tests.

lib/helpers.js
imurchie: That will require more work than just bumping the version. For instance, https://github.com/appium/appium-android-driver/blob/495d954348f41564cfd4139f31d59bdaff1e9653/lib/driver.js#L56 will fail (because `_.pairs` was removed and replaced with `_.toPairs`).

xuanzhaopeng:
Cool! Now works!

Following is my test: 

> 
> [debug] [ADB] Emulator Nexus_5X_API_23 not running
> [debug] [ADB] Launching Emulator with AVD Nexus_5X_API_23, launchTimeout60000 ms and readyTimeout 60000 ms
> [ADB] Checking whether emulator is present
> [ADB] Using emulator from /Users/robin/Library/Android/sdk/tools/emulator
> [ADB] Checking whether emulator is present
> [ADB] Using emulator from /Users/robin/Library/Android/sdk/tools/emulator
> **=> PS: Here is my debug :) . [debug] [ADB] Robin!!!!!!  Your Android SDK Version is 25**
> [ADB] Checking whether avdmanager is present
> **[ADB] Using avdmanager from /Users/robin/Library/Android/sdk/tools/bin/avdmanager**
> [debug] [ADB] Running '/Users/robin/Library/Android/sdk/tools/emulator' with args: ["-avd","Nexus_5X_API_23"]

mykola-mokhnach:
@xuanzhaopeng many thanks for your help. Lem'me know if there's something else to fix :)

lib/helpers.js
```diff
@@ -209,8 +209,28 @@ function buildStartCmd (startAppOptions, apiLevel) {
...
+  }
+  const propertiesPath = path.resolve(androidHome, 'tools', 'source.properties');
+  if (!await fs.exists(propertiesPath)) {
+    log.warn(`Cannot find ${propertiesPath} file to read SDK version from`);
+    return;
+  }
+  const propertiesContent = await fs.readFile(propertiesPath, 'utf8');
+  const versionMatcher = new RegExp(/Pkg\.Revision=(\d+)\.?(\d+)?\.?(\d+)?/);
+  const match = versionMatcher.exec(propertiesContent);
+  if (match) return {
```
jlipps: nit: technically we do full braces for ifs:

```js
if (match) {
  return {...};
}
log.warn('...');
```

mykola-mokhnach: Thanks for the hint. I just follow [this](https://github.com/appium/appium/blob/master/docs/en/contributing-to-appium/style-guide-2.0.md) document and it's stated there, that one may use return without brackets for short-circuiting. I though this rule might be applied here as well.

jlipps: good point, it is a form of short-circuiting. but the exception to the rule was designed specifically for this boilerplate:

```js
if (err) {
  return cb(err);
}
```

which is super common when using callbacks. now we don't use callbacks anymore so no real need for an exception.

mykola-mokhnach:
Squashed the stuff and enclosed if block.

jlipps:
excellent, looks great now

jlipps:
published in `appium-adb@2.12.0`

dmendes-daitan:
How can I update my current 1.6.3 to use this version ?

PR #205, opened 2017-03-20
Add helper methods for getting packages info

mykola-mokhnach:
There method will be then useful together with https://github.com/appium/io.appium.settings/pull/7


imurchie:
Published in `appium-adb@2.13.0`.

PR #206, opened 2017-03-21
Fix getting non-granted permissions

mykola-mokhnach:
This should speed up granting permissions for apps, since not all permissions should be granted explicitly. Some of them are considered as "normal" and are granted by default. See https://developer.android.com/guide/topics/permissions/requesting.html#normal-dangerous for more details.


imurchie:
Published in `appium-adb@2.13.1`.

PR #207, opened 2017-03-23
Add installOrUpgrade method to improve utilities setup

mykola-mokhnach:
The currently present _install_ method is unable to properly upgrade a package if an older version is already installed

### DIFF
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -209,7 +209,7 @@ apkUtilsMethods.installOrUpgrade = async function (apk, pkg = null, timeout = 60
     return;
   }
   if (!await this.isAppInstalled(pkg)) {
-    await this.installApk(apk, false, timeout);
+    await this.install(apk, false, timeout);
     return;
   }
   const pkgInfo = this.getPackageInfo(pkg);
@@ -229,13 +229,13 @@ apkUtilsMethods.installOrUpgrade = async function (apk, pkg = null, timeout = 60
   log.debug(`The installed "${pkg}" package is older than ${apk} (${pkgVersionCode} < ${apkVersionCode}). ` +
             `Executing upgrade`);
   try {
-    await this.installApk(apk, true, timeout);
+    await this.install(apk, true, timeout);
   } catch (err) {
     log.warn(`Cannot upgrade ${pkg} because of "${err.message}". Trying full reinstall`);
     if (!await this.uninstallApk(pkg)) {
       log.errorAndThrow(`"${pkg}" package cannot be uninstalled`);
     }
-    await this.installApk(apk, false, timeout);
+    await this.install(apk, false, timeout);
   }
 };
 
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -198,6 +198,47 @@ apkUtilsMethods.install = async function (apk, replace = true, timeout = 60000) 
   }
 };
 
+apkUtilsMethods.installOrUpgrade = async function (apk, pkg = null, timeout = 60000) {
+  let apkInfo = null;
+  if (!pkg) {
+    apkInfo = await this.getApkInfo(apk);
+    pkg = apkInfo.name;
+  }
+  if (!pkg) {
+    log.warn(`Cannot read the package name of ${apk}. Assuming correct app version is already installed`);
+    return;
+  }
+  if (!await this.isAppInstalled(pkg)) {
+    await this.adbExec(['install', apk], {timeout});
+    return;
+  }
+  const pkgInfo = this.getPackageInfo(pkg);
+  const pkgVersionCode = pkgInfo.versionCode;
+  if (!apkInfo) {
+    apkInfo = await this.getApkInfo(apk);
+  }
+  const apkVersionCode = apkInfo.versionCode;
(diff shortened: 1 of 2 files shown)


lib/tools/apk-utils.js
```diff
@@ -198,6 +198,47 @@ apkUtilsMethods.install = async function (apk, replace = true, timeout = 60000)
...
+    log.warn(`Cannot read the package name of ${apk}. Assuming correct app version is already installed`);
+    return;
+  }
+  if (!await this.isAppInstalled(pkg)) {
+    await this.adbExec(['install', apk], {timeout});
+    return;
+  }
+  const pkgInfo = this.getPackageInfo(pkg);
+  const pkgVersionCode = pkgInfo.versionCode;
+  if (!apkInfo) {
```
imurchie: Has anything changed by this point in terms of what `this.getApkInfo(apk)` will return? If not, why not just immediately do
```js
let apkInfo = await this.getApkInfo(apk);
pkg = pkg || apkInfo.name;
```

lib/tools/apk-utils.js
```diff
@@ -198,6 +198,47 @@ apkUtilsMethods.install = async function (apk, replace = true, timeout = 60000)
...
+  if (!pkg) {
+    apkInfo = await this.getApkInfo(apk);
+    pkg = apkInfo.name;
+  }
+  if (!pkg) {
+    log.warn(`Cannot read the package name of ${apk}. Assuming correct app version is already installed`);
+    return;
+  }
+  if (!await this.isAppInstalled(pkg)) {
+    await this.adbExec(['install', apk], {timeout});
```
imurchie: Seems like we could just `await this.install(apk, false, timeout);`. Then if installing changed sometime down the line it would only be in one place.

lib/tools/apk-utils.js
```diff
@@ -198,6 +198,47 @@ apkUtilsMethods.install = async function (apk, replace = true, timeout = 60000)
...
+    return;
+  }
+  const pkgInfo = this.getPackageInfo(pkg);
+  const pkgVersionCode = pkgInfo.versionCode;
+  if (!apkInfo) {
+    apkInfo = await this.getApkInfo(apk);
+  }
+  const apkVersionCode = apkInfo.versionCode;
+  if (!_.isUndefined(apkVersionCode) && !_.isUndefined(pkgVersionCode)) {
```
imurchie: Can this conditional be switched so it short-circuits?
```js
if (_.isUndefined(apkVersionCode) || _.isUndefined(pkgVersionCode)) {
  log.warn(`Cannot read version codes of ${apk} and/or ${pkg}. Assuming correct app version is already installed`);
  return;
}
```

lib/tools/apk-utils.js
```diff
@@ -198,6 +198,47 @@ apkUtilsMethods.install = async function (apk, replace = true, timeout = 60000)
...
+    if (pkgVersionCode >= apkVersionCode) {
+      log.debug(`The installed "${pkg}" package does not require upgrade (${pkgVersionCode} >= ${apkVersionCode})`);
+      return;
+    }
+    log.debug(`The installed "${pkg}" package is older than ${apk} (${pkgVersionCode} < ${apkVersionCode}). ` +
+              `Executing upgrade`);
+    try {
+      await this.adbExec(['install', '-r', apk], {timeout});
```
imurchie: Again, just `await this.install(apk, true, timeout);`?

test/unit/apk-utils-specs.js
```diff
@@ -491,4 +491,111 @@ describe('Apk-utils', () => {
       }
     });
   }));
+  describe('installOrUpgrade', withMocks({adb}, (mocks) => {
+    const pkgId = 'io.appium.settings';
+    const apkPath = '/path/to/my.apk';
+
+    it('should execute install if the package is not present', async () => {
+      mocks.adb.expects('getApkInfo').withExactArgs([apkPath]).once().returns({
```
imurchie: Why is `apkPath` expected in an array. The code above just passes the variable through to `getApkInfo`, and you're passing in a string, not an array. Same goes for all the rest of the mocks.

lib/tools/apk-utils.js
mykola-mokhnach: this is done intentionally to avoid execution of extra shell command in case the app is not installed.

lib/tools/apk-utils.js
mykola-mokhnach: ok

lib/tools/apk-utils.js
mykola-mokhnach: yup, this looks better

lib/tools/apk-utils.js
mykola-mokhnach: ok

test/unit/apk-utils-specs.js
mykola-mokhnach: yes, I see now. Thanks for pointing this out.

lib/tools/apk-utils.js
imurchie: Ok. That makes sense.

imurchie:
Published in `appium-adb@2.14.0`.

PR #208, opened 2017-03-25
Update network switch procedure and separate simulator calls from real device calls

mykola-mokhnach:
Please do not merge this PR before https://github.com/appium/io.appium.settings/pull/7 is merged and published

### DIFF
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -441,6 +441,7 @@ methods.setGeoLocation = async function (location, isEmulator = false) {
   }
   latitude = `${_.ceil(latitude, 5)}`;
   if (isEmulator) {
+    this.resetTelnetAuthToken();
     this.adbExec(['emu', 'geo', 'fix', longitude, latitude]);
     // A workaround for https://code.google.com/p/android/issues/detail?id=206180
     this.adbExec(['emu', 'geo', 'fix', longitude.replace('.', ','), latitude.replace('.', ',')]);
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -439,7 +439,7 @@ methods.setGeoLocation = async function (location, isEmulator = false) {
   if (isNaN(latitude)) {
     log.errorAndThrow(`location.latitude is expected to be a valid float number. '${location.latitude}' is given instead`);
   }
-  let latitude = `${_.ceil(latitude, 5)}`;
+  latitude = `${_.ceil(latitude, 5)}`;
   if (isEmulator) {
     this.adbExec(['emu', 'geo', 'fix', longitude, latitude]);
   } else {
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -9,6 +9,7 @@ import Logcat from '../logcat';
 import { sleep, retryInterval } from 'asyncbox';
 import { SubProcess } from 'teen_process';
 
+const SETTINGS_HELPER_ID = 'io.appium.settings';
 
 let methods = {};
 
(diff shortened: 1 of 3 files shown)


mykola-mokhnach:
@vrunoa FYI

lib/tools/adb-commands.js
```diff
@@ -390,8 +391,13 @@ methods.isWifiOn = async function () {
 /*
  * on: true (to turn on) or false (to turn off)
  */
-methods.setWifiState = async function (on) {
-  await this.shell(['settings', 'put', 'global', 'wifi_on', on ? 1 : 0]);
+methods.setWifiState = async function (on, isSimulator = false) {
+  if (isSimulator) {
```
vrunoa: @mykola-mokhnach could this _var_ name be changed to `isEmulator` since android are emulators not simulators

mykola-mokhnach:
No, this is called from android driver, where we just verify whether _opts.avd_ is set or not

vrunoa:
Awesome. What else is set with _io.appium.settings_ app, mock locations ? We can also do that with *adb* and maybe avoid using this app for emulators. I'll create a PR

mykola-mokhnach:
@vrunoa For location mocking I'd rather create a separate class, which can send all supported commands via emulator console:
http://stackoverflow.com/questions/2279647/how-to-emulate-gps-location-in-the-android-emulator
https://developer.android.com/studio/run/emulator-console.html

Please let me know if any support from my side is needed there.

vrunoa:
@mykola-mokhnach yeah, my idea was to add it in here [https://github.com/appium/appium-adb/blob/master/lib/tools/adb-emu-commands.js](https://github.com/appium/appium-adb/blob/master/lib/tools/adb-emu-commands.js)

mykola-mokhnach:
Yep. There is also _sendTelnetCommand_ method in _tools/adb-commands.js_. You may try to reuse it.

vrunoa:
@mykola-mokhnach we can use `adb emu geo fix 50 50`

mykola-mokhnach:
yep, that's even simpler then

mykola-mokhnach:
@vrunoa Added "adb emu geo fix" call. Please take a look now.
This will allow us to get rid of settings app installation on emulator

vrunoa:
@mykola-mokhnach I have code ready to PR, but I would like to add some functional tests, do you guys have an app for testing over emulators ?

mykola-mokhnach:
@vrunoa Functional tests are currently broken here. It's better to ask @imurchie about the current situation on Travis.
And regarding location mocking - I think it would be good to merge it in single PR, since then I can exclude settings app installation for simulator in android driver. @vrunoa Would you mind to merge your changes into this PR?

vrunoa:
sure @mykola-mokhnach I'll add them tonight

lib/tools/adb-commands.js
```diff
@@ -402,24 +408,34 @@ methods.isDataOn = async function () {
...
   }
   if (!_.isUndefined(data)) {
-    dataOpts = ['-e', 'data', (data ? 'on' : 'off')];
+    this.setDataState(data, isEmulator);
+  }
+};
+
+methods.setGeoLocation = async function (location, isEmulator = false) {
+  if (isEmulator) {
+    this.adbExec(['emu', 'geo', 'fix', ''+ location.longitude, '' + location.latitude]);
```
jlipps: template strings please!

jlipps: this is nice, we don't need telnet anymore? that's how we've been doing it but also it was flakey.

mykola-mokhnach: This was a recommendation from @vrunoa. If you say it is not very stable to do it via adb them we can anytime change it to telnet if we get complains from users.

jlipps: No I think this is great! I just didn't know about it, and think it is something that didn't exist when we did the telnet option. Maybe this is much more stable.

mykola-mokhnach: @jlipps I've checked what people say about that and found the bug
https://code.google.com/p/android/issues/detail?id=206180
Is there any method in Node to set decimal splitter based on location?

vrunoa: @jlipps @mykola-mokhnach  I tried this yesterday and I notice some weird issue, I could set some locations, for ex. I put my city and Vancouver and it didnt change, but it was working with Google example. I wonder if this was an issue with the Google Maps app or the emu. I would like to create an app to tests this before removing the .settings helpers app to set the mock location.

vrunoa: nice catch @mykola-mokhnach

mykola-mokhnach: ok, I'll execute the command twice - once with dots and the second time with commas. Then it'll work in all cases even without knowing the actual locale.

vrunoa:
@mykola-mokhnach can you add `this.resetTelnetAuthToken();` to the `emuMethods.fingerprint` as well ? That might also fail if the token was erased

imurchie:
Published in `appium-adb@2.16.0`.

PR #209, opened 2017-03-27
Try to restart adbd as root on reboot failure

imurchie:
Fixes https://github.com/appium/appium/issues/6894

### DIFF
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -477,21 +477,36 @@ systemCallMethods.waitForDevice = async function (appDeviceReadyTimeout = 30) {
 };
 
 systemCallMethods.reboot = async function () {
-  await this.shell(['stop']);
-  await B.delay(2000); // let the emu finish stopping;
-  await this.setDeviceProperty('sys.boot_completed', 0);
-  await this.shell(['start']);
-  await retryInterval(90, 1000, async () => {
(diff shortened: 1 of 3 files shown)


lib/tools/system-calls.js
```diff
@@ -477,7 +477,18 @@ systemCallMethods.waitForDevice = async function (appDeviceReadyTimeout = 30) {
...
+  try {
+    await this.shell(['stop']);
+  } catch (err) {
+    if (err.message.indexOf('must be root') === -1) {
+      throw err;
+    }
+    // this device needs adb to be running as root to stop.
+    // so try to restart the daemon
+    log.debug('Device requires adb to be running as root in order to reboot. Restarting daemon');
+    await exec(this.executable.path, ['root']);
```
mykola-mokhnach: does it work properly even without executing _adb kill-server_?

imurchie: Yes. You can try, just run `adb root` on the command line.

vrunoa: @imurchie should we run `await exec(this.executable.path, ['unroot']);` after the device reboots to avoid mean people trying to do bad things with root privileges?

imurchie: Yes, that seems like a good idea. I didn't actually know that existed!

PR #210, opened 2017-04-01
Add method to rotate emulators

vrunoa:
We can rotate the emulators using `adb emu rotate` command, i'm adding this method here so we can call it later from base android drivers.

### DIFF
--- a/lib/tools/adb-emu-commands.js
+++ b/lib/tools/adb-emu-commands.js
@@ -18,40 +18,33 @@ emuMethods.isEmulatorConnected = async function (udid) {
   return false;
 };
 
+emuMethods.checkEmulatorConnected = async function (udid) {
+  if (!_.isUndefined(udid) && !(await this.isEmulatorConnected(udid)) ) {
+    log.errorAndThrow(`Device '${udid}' is not available.`);
+  } else if ((await this.getConnectedEmulators()).length === 0) {
+    log.errorAndThrow('No devices connected');
+  }
+  if (!_.isUndefined(udid)) {
+    await this.setDeviceId(udid);
+  }
+};
+
 emuMethods.fingerprint = async function (fingerprintId, udid = undefined) {
   if (!fingerprintId) {
     log.errorAndThrow('Fingerprint id parameter must be defined');
   }
-
   // the method used only works for API level 23 and above
   let level = await this.getApiLevel();
   if (parseInt(level, 10) < 23) {
     log.errorAndThrow(`Device API Level must be >= 23. Current Api level '${level}'`);
   }
-
-  if (!_.isUndefined(udid) && !(await this.isEmulatorConnected(udid)) ) {
-    log.errorAndThrow(`Device '${udid}' is not available.`);
-  } else if ((await this.getConnectedEmulators()).length === 0) {
-    log.errorAndThrow('No devices connected');
-  }
-
-  if (udid) {
-    await this.setDeviceId(udid);
-  }
+  await this.checkEmulatorConnected(udid);
   this.resetTelnetAuthToken();
   await this.adbExec(['emu', 'finger', 'touch', fingerprintId]);
 };
 
 emuMethods.rotate = async function (udid) {
-  if (!_.isUndefined(udid) && !(await this.isEmulatorConnected(udid)) ) {
-    log.errorAndThrow(`Device '${udid}' is not available.`);
-  } else if ((await this.getConnectedEmulators()).length === 0) {
-    log.errorAndThrow('No devices connected');
-  }
-
-  if (!_.isUndefined(udid)) {
-    await this.setDeviceId(udid);
-  }
+  await this.checkEmulatorConnected(udid);
   this.resetTelnetAuthToken();
   this.adbExec(['emu', 'rotate']);
 };
--- a/lib/tools/adb-emu-commands.js
+++ b/lib/tools/adb-emu-commands.js
@@ -19,6 +19,12 @@ emuMethods.isEmulatorConnected = async function () {
   return !!_.find(emulators, (x) => x && x.udid === this.curDeviceId);
 };
 
+emuMethods.checkEmulatorConnected = async function () {
+  if (!(await this.isEmulatorConnected())) {
+    log.errorAndThrow(`Emulator ${this.curDeviceId} disconnected!`);
+  }
+};
+
 emuMethods.fingerprint = async function (fingerprintId) {
   if (!fingerprintId) {
     log.errorAndThrow('Fingerprint id parameter must be defined');
@@ -28,13 +34,13 @@ emuMethods.fingerprint = async function (fingerprintId) {
   if (parseInt(level, 10) < 23) {
     log.errorAndThrow(`Device API Level must be >= 23. Current Api level '${level}'`);
   }
-  await this.isEmulatorConnected();
+  await this.checkEmulatorConnected();
   await this.resetTelnetAuthToken();
   await this.adbExec(['emu', 'finger', 'touch', fingerprintId]);
 };
 
 emuMethods.rotate = async function () {
-  await this.isEmulatorConnected();
+  await this.checkEmulatorConnected();
   await this.resetTelnetAuthToken();
   await this.adbExec(['emu', 'rotate']);
 };
--- a/lib/tools/adb-emu-commands.js
+++ b/lib/tools/adb-emu-commands.js
@@ -28,8 +28,12 @@ emuMethods.isEmulatorConnected = async function (udid) {
 };
 
 emuMethods.checkEmulatorConnected = async function (udid) {
-  if (!_.isUndefined(udid) && !(await this.isEmulatorConnected(udid)) ) {
-    log.errorAndThrow(`Device '${udid}' is not available.`);
+  if (!_.isUndefined(udid)) {
+    if (await this.isEmulatorConnected(udid)) {
+      await this.setDeviceId(udid);
+    } else {
+      log.errorAndThrow(`Device '${udid}' is not available.`);
+    }
   } else if ((await this.getEmulators()).length === 0) {
     log.errorAndThrow('No devices connected');
   }
--- a/lib/tools/adb-emu-commands.js
+++ b/lib/tools/adb-emu-commands.js
@@ -14,27 +14,12 @@ emuMethods.getEmulators = async function() {
   return this.emulators;
 };
 
-emuMethods.isEmulatorConnected = async function (udid) {
+emuMethods.isEmulatorConnected = async function () {
   let emulators = await this.getEmulators();
-  if (!emulators.length) {
-    return false;
-  }
-  return !!_.find(emulators, (x) => x && x.udid === udid);
+  return !!_.find(emulators, (x) => x && x.udid === this.curDeviceId);
 };
 
-emuMethods.checkEmulatorConnected = async function (udid) {
-  if (!_.isUndefined(udid)) {
-    if (await this.isEmulatorConnected(udid)) {
-      await this.setDeviceId(udid);
-    } else {
-      log.errorAndThrow(`Device '${udid}' is not available.`);
-    }
-  } else if ((await this.getEmulators()).length === 0) {
-    log.errorAndThrow('No devices connected');
-  }
-};
-
-emuMethods.fingerprint = async function (fingerprintId, udid = undefined) {
+emuMethods.fingerprint = async function (fingerprintId) {
   if (!fingerprintId) {
     log.errorAndThrow('Fingerprint id parameter must be defined');
   }
@@ -43,13 +28,13 @@ emuMethods.fingerprint = async function (fingerprintId, udid = undefined) {
   if (parseInt(level, 10) < 23) {
     log.errorAndThrow(`Device API Level must be >= 23. Current Api level '${level}'`);
   }
-  await this.checkEmulatorConnected(udid);
+  await this.isEmulatorConnected();
   await this.resetTelnetAuthToken();
   await this.adbExec(['emu', 'finger', 'touch', fingerprintId]);
 };
 
-emuMethods.rotate = async function (udid) {
-  await this.checkEmulatorConnected(udid);
+emuMethods.rotate = async function () {
+  await this.isEmulatorConnected();
   await this.resetTelnetAuthToken();
   await this.adbExec(['emu', 'rotate']);
 };
--- a/lib/tools/adb-emu-commands.js
+++ b/lib/tools/adb-emu-commands.js
@@ -21,7 +21,7 @@ emuMethods.isEmulatorConnected = async function () {
 
 emuMethods.checkEmulatorConnected = async function () {
   if (!(await this.isEmulatorConnected())) {
-    log.errorAndThrow(`Emulator ${this.curDeviceId} disconnected!`);
+    log.errorAndThrow(`The emulator "${this.curDeviceId}" was unexpectedly disconnected`);
   }
 };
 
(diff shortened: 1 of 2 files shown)


lib/tools/adb-emu-commands.js
```diff
@@ -38,9 +38,15 @@ emuMethods.fingerprint = async function (fingerprintId, udid = undefined) {
   if (udid) {
     await this.setDeviceId(udid);
   }
+  this.resetTelnetAuthToken();
   await this.adbExec(['emu', 'finger', 'touch', fingerprintId]);
 };
 
+emuMethods.rotate = async function () {
```
mykola-mokhnach: where this method is expected to be used?

vrunoa: when users set the deviceOrientation capability or even during tests to verify their apps are rotating(or not) according to what the want to give as user experience.

mykola-mokhnach: does this call happen in appium-android-driver? I'm just trying to build the full picture. 

And there is also one more thing, which bothers me a bit - we always use 'adb emu' call without any additional arguments. Will it work as expected in case there are multiple simulators running? Probably, one has to specify it explicitly to which simulator the command is going to be sent (like -s option for adb with real device)?

vrunoa: @mykola-mokhnach you're right I forgot to add 
```
if (udid) {
      await this.setDeviceId(udid);
}
```
yeah, the idea is to call it from `appium-android-driver`

lib/tools/adb-emu-commands.js
```diff
@@ -38,9 +38,24 @@ emuMethods.fingerprint = async function (fingerprintId, udid = undefined) {
   if (udid) {
     await this.setDeviceId(udid);
   }
+  this.resetTelnetAuthToken();
   await this.adbExec(['emu', 'finger', 'touch', fingerprintId]);
 };
 
+emuMethods.rotate = async function (udid) {
+  if (!_.isUndefined(udid) && !(await this.isEmulatorConnected(udid)) ) {
```
mykola-mokhnach: probably it makes sense to refactor the stuff and move setting of the id/token reset into separate method. Also, there is no need to call the stuff twice, so we could also cache results and avoid calling the save some time if an emulator command is invoked more than once.

vrunoa: I was just think about caching the results, what if the emulators disconnect between commands ? I think we would have unhandled errors

mykola-mokhnach: I would sacrifice the possibility of simulator disconnection to the overall test speed. Each terminal command invocation is expensive from performance point of view

lib/tools/adb-emu-commands.js
```diff
@@ -18,29 +18,37 @@ emuMethods.isEmulatorConnected = async function (udid) {
...
-  }
+  await this.checkEmulatorConnected(udid);
+  this.resetTelnetAuthToken();
   await this.adbExec(['emu', 'finger', 'touch', fingerprintId]);
 };
 
+emuMethods.rotate = async function (udid) {
+  await this.checkEmulatorConnected(udid);
+  this.resetTelnetAuthToken();
+  this.adbExec(['emu', 'rotate']);
```
mykola-mokhnach: await?

lib/tools/adb-emu-commands.js
```diff
@@ -18,29 +18,37 @@ emuMethods.isEmulatorConnected = async function (udid) {
...
-    await this.setDeviceId(udid);
-  }
+  await this.checkEmulatorConnected(udid);
+  this.resetTelnetAuthToken();
   await this.adbExec(['emu', 'finger', 'touch', fingerprintId]);
 };
 
+emuMethods.rotate = async function (udid) {
+  await this.checkEmulatorConnected(udid);
+  this.resetTelnetAuthToken();
```
mykola-mokhnach: await?

lib/tools/adb-emu-commands.js
```diff
@@ -18,29 +18,37 @@ emuMethods.isEmulatorConnected = async function (udid) {
...
-    log.errorAndThrow(`Device '${udid}' is not available.`);
-  } else if ((await this.getConnectedEmulators()).length === 0) {
-    log.errorAndThrow('No devices connected');
-  }
-
-  if (udid) {
-    await this.setDeviceId(udid);
-  }
+  await this.checkEmulatorConnected(udid);
+  this.resetTelnetAuthToken();
```
mykola-mokhnach: await?

vrunoa: yes, `checkEmulatorConnected` calls to `systemCallMethods.getConnectedEmulators` which is async

lib/tools/adb-emu-commands.js
```diff
@@ -18,29 +18,37 @@ emuMethods.isEmulatorConnected = async function (udid) {
   return false;
 };
 
+emuMethods.checkEmulatorConnected = async function (udid) {
```
jlipps: based on how I read the logic here, if you fail to pass in a `udid` to this method, this method will do precisely nothing. is that intended? maybe we should error instead?

lib/tools/adb-emu-commands.js
jlipps: looks like `resetTelnetAuthToken` needs some `await` love here and below as @mykola-mokhnach pointed out.

vrunoa: oh sorry @mykola-mokhnach I didnt understand I was missing an `await`

lib/tools/adb-emu-commands.js
vrunoa: @jlipps I was wondering that, should we always ask for an udid when using emulators ? I think so, right?

jlipps: i think so. i think it's better for this function to error if the parameters you call it with would otherwise cause this function to do nothing. in this case it means erroring if you fail to pass in a udid.

if you wind up in a position where you don't have a udid, and want nothing to happen, then you just choose not to call this function. rather than call it and have it do nothing.

lib/tools/adb-emu-commands.js
```diff
@@ -5,8 +5,17 @@ import path from 'path';
 
 let emuMethods = {};
 
+emuMethods.getEmulators = async function() {
+  if (_.isArray(this.emulators) &&
+      this.emulators.length !== 0) {
```
mykola-mokhnach: if (this.emulators.length) === if (this.emulators.length !== 0)

vrunoa: good point

lib/tools/adb-emu-commands.js
```diff
@@ -5,8 +5,17 @@ import path from 'path';
...
+    return this.emulators;
+  }
+  this.emulators = await this.getConnectedEmulators();
+  return this.emulators;
+};
+
 emuMethods.isEmulatorConnected = async function (udid) {
-  let emulators = await this.getConnectedEmulators();
+  let emulators = await this.getEmulators();
   if (!emulators.length) {
```
mykola-mokhnach: this all could be replaced with 

```
return !!_.find(emulators, (x) => x && x.udid === udid);
```

vrunoa: I like that, would we this more readable
`return Boolean(_.find(emulators, (x) => x && x.udid === udid));`

mykola-mokhnach: Double exclamation mark is a quite common practice in JS to have variables casted to boolean - everyone should understand what this means.

lib/tools/adb-emu-commands.js
jlipps: still think this should be changed

vrunoa: @jlipps I clean up and refactor the code a bit, let me know what you think
https://github.com/vrunoa/appium-adb/blob/add-emu-rotate/lib/tools/adb-emu-commands.js#L25-L35

lib/tools/adb-emu-commands.js
```diff
@@ -5,42 +5,55 @@ import path from 'path';
...
```
mykola-mokhnach: This _if_ might be also removed. the _find_ will work  as expected for the empty array case.

lib/tools/adb-emu-commands.js
```diff
@@ -5,42 +5,55 @@ import path from 'path';
...
     return false;
   }
-  for (let emulator of emulators) {
-    if (emulator.udid === udid) {
-      return true;
+  return !!_.find(emulators, (x) => x && x.udid === udid);
+};
+
+emuMethods.checkEmulatorConnected = async function (udid) {
```
jlipps: still a bit confused about this method. what do we want to do if `udid` is not defined? just check that _some_ emulator(s) are connected? how does that help?

vrunoa: `just check that some emulator(s) are connected?` YES. 
Are we always setting the emulator *udid*? Is that managed by the `appium-android-driver`?

vrunoa: I'm seeing that we dont set the udid in the `appium-android-driver` call https://github.com/appium/appium-android-driver/blob/e073b913f3deefdb94cb3ea0a4c9e6f9721b0a0b/lib/commands/actions.js#L179-L181 and the devideId is already set after prepareEmulator https://github.com/appium/appium-android-driver/blob/master/lib/android-helpers.js#L119 so I guess we can remove the udid validation. what do you think @jlipps @mykola-mokhnach ?

lib/tools/adb-emu-commands.js
```diff
@@ -43,13 +28,13 @@ emuMethods.fingerprint = async function (fingerprintId, udid = undefined) {
   if (parseInt(level, 10) < 23) {
     log.errorAndThrow(`Device API Level must be >= 23. Current Api level '${level}'`);
   }
-  await this.checkEmulatorConnected(udid);
+  await this.isEmulatorConnected();
```
mykola-mokhnach: what is the point to invoke this method if the returned result is not checked?

vrunoa: my bad! that's for doing things before having coffee. The idea was to throw an exception, I forgot to add it. Fixing it.

lib/tools/adb-emu-commands.js
```diff
@@ -5,42 +5,46 @@ import path from 'path';
...
+};
+
+emuMethods.isEmulatorConnected = async function () {
+  let emulators = await this.getEmulators();
+  return !!_.find(emulators, (x) => x && x.udid === this.curDeviceId);
+};
+
+emuMethods.checkEmulatorConnected = async function () {
+  if (!(await this.isEmulatorConnected())) {
+    log.errorAndThrow(`Emulator ${this.curDeviceId} disconnected!`);
```
mykola-mokhnach: better to change to:

```
`The emulator "${this.curDeviceId}" was unexpectedly disconnected`
```

vrunoa: should we cache emulators list ?

lib/tools/adb-emu-commands.js
```diff
@@ -5,42 +5,46 @@ import path from 'path';
 
 let emuMethods = {};
 
-emuMethods.isEmulatorConnected = async function (udid) {
-  let emulators = await this.getConnectedEmulators();
-  if (!emulators.length) {
-    return false;
+emuMethods.getEmulators = async function() {
```
jlipps: an easier way to handle caching the value without needing to store state on the object would be to use `_.memoize`:

```js
emuMethods.getEmulators = _.memoize(async function() {
  return await this.getConnectedEmulators();
});
```

jlipps: also, we might want to consider `debounce` instead of `memoize` because what if the person is running appium for a long time? the emulator state might reasonably change, right?

vrunoa: It shouldnt disconnect but it might happen. I have never use `debounce` or `memoize`, no idea what difference between them and which one it's better. 
Should we cache the emulators list ? Or verify it's connected on every emu command ?

jlipps: `memoize` caches forever, `debounce` caches for a certain amount of time.

@vrunoa i guess we have to make an estimate of the likelihood that emulator state will change during the lifetime of an ADB instance.

debouncing to like 5s might be a good idea just to avoid making lots of calls at the same time? you're better equipped to judge than i am.

vrunoa: @jlipps I would use debounce then. It's quite unlikely for an emulator to disconnect but It might happen on large tests. I would use `debounce` with 5 minutes. what do you guys think?

jlipps: hmm hmm. we can start with that and see. i would have said lower? what's the cost of running the call?

vrunoa: This are the results of running get connected emulators 30 times on my local laptop
```javascript
let times = [];
for(let i = 0;i<30;i++) {
      let t1 = (new Date()).getTime();
      await adb.getConnectedEmulators();
      let t2 = (new Date()).getTime();
      times.push([t1, t2, t2-t1]);
 }
```

```json
[ [ 1492484336620, 1492484336647, 27 ],
  [ 1492484336647, 1492484336663, 16 ],
  [ 1492484336664, 1492484336683, 19 ],
  [ 1492484336683, 1492484336697, 14 ],
  [ 1492484336697, 1492484336719, 22 ],
  [ 1492484336719, 1492484336739, 20 ],
  [ 1492484336739, 1492484336753, 14 ],
  [ 1492484336753, 1492484336770, 17 ],
  [ 1492484336770, 1492484336792, 22 ],
  [ 1492484336792, 1492484336805, 13 ],
  [ 1492484336805, 1492484336819, 14 ],
  [ 1492484336819, 1492484336836, 17 ],
  [ 1492484336836, 1492484336857, 21 ],
  [ 1492484336857, 1492484336871, 14 ],
  [ 1492484336871, 1492484336887, 16 ],
  [ 1492484336887, 1492484336904, 17 ],
  [ 1492484336904, 1492484336925, 21 ],
  [ 1492484336925, 1492484336939, 14 ],
  [ 1492484336939, 1492484336952, 13 ],
  [ 1492484336952, 1492484336975, 23 ],
  [ 1492484336975, 1492484336991, 16 ],
  [ 1492484336991, 1492484337004, 13 ],
  [ 1492484337004, 1492484337019, 15 ],
  [ 1492484337019, 1492484337047, 28 ],
  [ 1492484337047, 1492484337068, 21 ],
  [ 1492484337068, 1492484337087, 19 ],
  [ 1492484337087, 1492484337107, 20 ],
  [ 1492484337107, 1492484337128, 21 ],
  [ 1492484337128, 1492484337141, 13 ],
  [ 1492484337141, 1492484337155, 14 ] ]
```

jlipps: @vrunoa is that in ms? it only takes 27 ms to run it the first time? in that case i don't think we need any caching at all.

vrunoa:
@jlipps yes it's 27 miliseconds. I rather not to cache, so we get the correct error in case the emulator got disconnected(or crashes). it's up to you guys.

lib/tools/adb-emu-commands.js
```diff
@@ -5,42 +5,37 @@ import path from 'path';
...
-    return false;
-  }
-  for (let emulator of emulators) {
-    if (emulator.udid === udid) {
-      return true;
-    }
+  return !!_.find(emulators, (x) => x && x.udid === this.curDeviceId);
+};
+
+emuMethods.checkEmulatorConnected = async function () {
```
jlipps: since this throws, maybe we should call it `verifyEmulatorConnected` rather than `check`, since I'd expect a verification to throw but not a check?

imurchie:
Published in `appium-adb@2.19.0`.

PR #211, opened 2017-04-03
Add proxy settings

vrunoa:
Add a method to set proxy settings via adb for the emulators, the idea is user can set proxy settings in the capabilities so they can use tunnels. They will also have to add the `--http-proxy` param in `avdArgs`.

### DIFF
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -728,19 +728,19 @@ methods.getScreenSize = async function() {
   return null;
 };
 
-methods.setHttpProxy = async function (proxy_host, proxy_port) {
-  let proxy = `${proxy_host}:${proxy_port}`;
-  if (_.isUndefined(proxy_host)) {
+methods.setHttpProxy = async function (proxyHost, proxyPort) {
+  let proxy = `${proxyHost}:${proxyPort}`;
+  if (_.isUndefined(proxyHost)) {
     log.errorAndThrow(`Call to setHttpProxy method with undefined proxy_host: ${proxy}`);
   }
-  if (_.isUndefined(proxy_port)) {
+  if (_.isUndefined(proxyPort)) {
     log.errorAndThrow(`Call to setHttpProxy method with undefined proxy_port ${proxy}`);
   }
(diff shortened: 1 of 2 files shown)


lib/tools/adb-commands.js
```diff
@@ -728,4 +728,19 @@ methods.getScreenSize = async function() {
   return null;
 };
 
+methods.setHttpProxy = async function (proxy_host, proxy_port) {
```
jlipps: we use `camelCase`

lib/tools/adb-commands.js
```diff
@@ -728,4 +728,19 @@ methods.getScreenSize = async function() {
   return null;
 };
 
+methods.setHttpProxy = async function (proxyHost, proxyPort) {
```
mykola-mokhnach: I'd rather rename the variable to proxyUrl or proxyAddress to reflect the fact that it expects not just a simple host name

vrunoa: @mykola-mokhnach I would leave proxyHost since the adb setting is called `global_http_proxy_host`

lib/tools/adb-commands.js
```diff
@@ -728,4 +728,19 @@ methods.getScreenSize = async function() {
...
+  let proxy = `${proxyHost}:${proxyPort}`;
+  if (_.isUndefined(proxyHost)) {
+    log.errorAndThrow(`Call to setHttpProxy method with undefined proxy_host: ${proxy}`);
+  }
+  if (_.isUndefined(proxyPort)) {
+    log.errorAndThrow(`Call to setHttpProxy method with undefined proxy_port ${proxy}`);
+  }
+  await this.shell(['settings', 'put', 'global', 'http_proxy', proxy]);
```
mykola-mokhnach: or with more functional style

```javascript
for (let suffix of [['global', 'http_proxy', proxy],
                            ['secure', 'http_proxy', proxy],
                            ['system', 'http_proxy', proxy],
                            ['system', 'global_http_proxy_host', proxyHost],
                            ['system', 'global_http_proxy_port', proxyPort]]) {
  await this.shell(['settings', 'put'].concat(suffix));
}
```

PR #213, opened 2017-04-11
Set the default temp folder path

mykola-mokhnach:
Fixes the case when package activity name cannot be extracted from manifest file. See https://github.com/appium/appium/issues/8221 for more details.


imurchie:
Published in `appium-adb@2.16.1`

PR #214, opened 2017-04-11
Fix the wrong APK path reference ( #8221)

NozomiIto:
Please see https://github.com/appium/appium/issues/8221 for detail.


imurchie:
Published in `appium-adb@2.16.2`.

PR #215, opened 2017-04-13
Log error from unroot but continue

imurchie:
Fixes https://github.com/appium/appium/issues/8234

### DIFF
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -505,7 +505,11 @@ systemCallMethods.reboot = async function () {
       }
     });
   } finally {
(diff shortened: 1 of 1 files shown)


lib/tools/system-calls.js
```diff
@@ -505,7 +505,11 @@ systemCallMethods.reboot = async function () {
       }
     });
   } finally {
-    await exec(this.executable.path, ['unroot']);
+    try {
+      await exec(this.executable.path, ['unroot']);
+    } catch (err) {
+      log.debug(`Unable to unroot adb daemon: '${err.message}'`);
```
mykola-mokhnach: debug or warn?

PR #218, opened 2017-04-21
Fix behavor with spaces

crazyk2:
I had to create fork from scratch to assign my email with commit.


imurchie:
Published in `appium-adb@2.17.2`

PR #220, opened 2017-04-26
Remove adm-zip from appium-adb

dpgraham:
* Refactored the `unzipFile` method to use `zip.toInMemoryZip`
* Refactored `checkApkKeystoreMatch` to use `zip.readEntries`
* Added a functional test for `unzipFile` that passed with adm-zip, failed when adm-zip was removed and then passed again when `unzipFile` was refactored


lib/tools/apk-signing.js
```diff
@@ -152,11 +151,14 @@ apkSigningMethods.checkApkKeystoreMatch = async function (keytool, md5re, keysto
     log.debug(`keystore MD5: ${keystoreHash}`);
     let matchesKeystore = entryHash && entryHash === keystoreHash;
     log.debug(`Matches keystore? ${matchesKeystore}`);
+
+    // If we have a keystore match, stop iterating
     if (matchesKeystore) {
-      return true;
+      foundKeystoreMatch = true;
+      return false;
```
mykola-mokhnach: will this stop the iteration? I thought it is necessary to send the **end** event to the object to do it

imurchie: https://github.com/appium/appium-support/pull/50

mykola-mokhnach: false flag does not look very obvious IMHO. It would be more readable to have an explicit argument for  such purpose like in objectiveC blocks:

```objc
[array enumerateObjectsUsingBlock:^(id obj, NSUInteger idx, BOOL *stop) {
    //…
    if(idx == 1000) 
        *stop = YES;
}];
```

mykola-mokhnach: OK, JQuery's **each** also follows the "return false" rule. Please ignore the previous comment.

dpgraham:
Tagged and published to 2.17.3 (https://www.npmjs.com/package/appium-adb)

PR #221, opened 2017-04-26
Add getAdbVersion method

vrunoa:
Add a method to get adb version. Newer versions of adb added a lot of methods for emulators so in order to use those methods we should verify a minimum version of adb. This can also help us in the future to deprecate old version of adb to use with appium.

### DIFF
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -412,19 +412,20 @@ systemCallMethods.launchAVD = async function (avdName, avdArgs, language, countr
   return proc;
 };
 
-systemCallMethods.getAdbVersion = async function () {
-  if (!_.isUndefined(this.adbVersion)) {
-    return this.adbVersion;
-  }
+systemCallMethods.getAdbVersion = _.memoize(async function () {
   try {
-    let version = await this.adbExec('version');
-    this.adbVersion = version.replace(/Android\sDebug\sBridge\sversion\s([\d\.]*)[\s\w\-]*/, "$1");
-    return this.adbVersion;
+    let adbVersion = (await this.adbExec('version'))
+      .replace(/Android\sDebug\sBridge\sversion\s([\d\.]*)[\s\w\-]*/, "$1");
+    return {
+      versionString: adbVersion,
+      versionFloat: parseFloat(adbVersion).toFixed(1)
+    };
   } catch (e) {
     log.errorAndThrow(`Error executing adb version. Original error: '${e.message}'; ` +
-                      `Stderr: '${(e.stderr || '').trim()}'; Code: '${e.code}'`);
+                        `Stderr: '${(e.stderr || '').trim()}'; Code: '${e.code}'`);
   }
-};
+});
+
 
 systemCallMethods.checkAvdExist = async function (avdName) {
   let cmd, result;
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -418,7 +418,7 @@ systemCallMethods.getAdbVersion = _.memoize(async function () {
       .replace(/Android\sDebug\sBridge\sversion\s([\d\.]*)[\s\w\-]*/, "$1");
     return {
       versionString: adbVersion,
-      versionFloat: parseFloat(adbVersion).toFixed(1)
+      versionFloat: parseFloat(adbVersion.replace(/.([^.]*)$/, '$1'))
     };
   } catch (e) {
     log.errorAndThrow(`Error executing adb version. Original error: '${e.message}'; ` +
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -425,7 +425,7 @@ systemCallMethods.getAdbVersion = _.memoize(async function () {
       patch: parts[2] ? parseInt(parts[2], 10) : undefined,
     };
   } catch (e) {
-    log.errorAndThrow(`Error executing adb version. Original error: '${e.message}'; ` +
+    log.errorAndThrow(`Error getting adb version. Original error: '${e.message}'; ` +
                         `Stderr: '${(e.stderr || '').trim()}'; Code: '${e.code}'`);
   }
 });
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -418,7 +418,7 @@ systemCallMethods.getAdbVersion = async function () {
   }
   try {
     let version = await this.adbExec('version');
-    this.adbVersion = version.replace(/(Android\sDebug\sBridge\sversion\s)([\d\.]*)([\s\w\-])*/, "$2");
+    this.adbVersion = version.replace(/Android\sDebug\sBridge\sversion\s([\d\.]*)[\s\w\-]*/, "$1");
     return this.adbVersion;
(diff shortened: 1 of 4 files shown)


lib/tools/system-calls.js
```diff
@@ -411,6 +412,20 @@ systemCallMethods.launchAVD = async function (avdName, avdArgs, language, countr
   return proc;
 };
 
+systemCallMethods.getAdbVersion = async function () {
+  if (!_.isUndefined(this.adbVersion)) {
+    return this.adbVersion;
+  }
+  try {
+    let version = await this.adbExec('version');
+    this.adbVersion = version.replace(/(Android\sDebug\sBridge\sversion\s)([\d\.]*)([\s\w\-])*/, "$2");
```
imurchie: Why capture so many groups? Would not `version.replace(/Android\sDebug\sBridge\sversion\s([\d\.]*)[\s\w\-]*/, "$1");` give the same result?

vrunoa: @imurchie yes. Changing it.

lib/tools/system-calls.js
```diff
@@ -411,6 +412,20 @@ systemCallMethods.launchAVD = async function (avdName, avdArgs, language, countr
   return proc;
 };
 
+systemCallMethods.getAdbVersion = async function () {
```
mykola-mokhnach: I think it would be better to return the version number as an object, like

```
 {major: 1,
  minor: 1,
  build: 2}
```

in this case it will be easier to compare it.

imurchie: I was just about to ask how it was envisioned this would be used. If we need to access it piecemeal, I agree this would be better, so the parsing doesn't have to happen all over the place. I would add `versionFloat: 1.1` and `versionString: "1.1.2"`.

jlipps: rather than implementing our own caching can't we use lodash to just wrap the method? then we don't need an instance variable, too

jlipps: agree with these reflections about version return

test/unit/syscalls-specs.js
```diff
@@ -107,6 +107,18 @@ describe('System calls', withMocks({teen_process}, (mocks) => {
...
 
 describe('System calls',  withMocks({adb, B, teen_process}, (mocks) => {
+  it('should return adb version', async () => {
+    mocks.adb.expects("adbExec")
+      .once()
+      .withExactArgs('version')
+      .returns("Android Debug Bridge version 1.0.39\nRevision 5943271ace17-android");
+    let adbVersion = await adb.getAdbVersion();
+    adbVersion.versionString.should.equal("1.0.39");
+    adbVersion.versionFloat.should.equal("1.0");
```
jlipps: shouldn't you be testing equality against a float instead of a string?

mykola-mokhnach: It might be a bad idea to compare two floats for equality. Better to use **within**

vrunoa: I try to use `adbVersion.versionFloat.should.equal(1.0)` but it's converting the value to _1_ and the test fail

lib/tools/system-calls.js
```diff
@@ -411,6 +412,21 @@ systemCallMethods.launchAVD = async function (avdName, avdArgs, language, countr
   return proc;
 };
 
+systemCallMethods.getAdbVersion = _.memoize(async function () {
+  try {
+    let adbVersion = (await this.adbExec('version'))
+      .replace(/Android\sDebug\sBridge\sversion\s([\d\.]*)[\s\w\-]*/, "$1");
+    return {
+      versionString: adbVersion,
+      versionFloat: parseFloat(adbVersion).toFixed(1)
```
mykola-mokhnach: why use toFixed? I think all the digits there are important

vrunoa: @mykola-mokhnach i'll change the `parseFloat` and remove `toFixed` since it's not getting the last numbers

lib/tools/system-calls.js
```diff
@@ -411,6 +412,21 @@ systemCallMethods.launchAVD = async function (avdName, avdArgs, language, countr
   return proc;
 };
 
+systemCallMethods.getAdbVersion = _.memoize(async function () {
+  try {
+    let adbVersion = (await this.adbExec('version'))
+      .replace(/Android\sDebug\sBridge\sversion\s([\d\.]*)[\s\w\-]*/, "$1");
+    return {
+      versionString: adbVersion,
+      versionFloat: parseFloat(adbVersion.replace(/.([^.]*)$/, '$1'))
```
imurchie: I don't think this is what we want. The point of having a `versionFloat` is to be able to quickly and easily check major-minor versions together. In my opinion this object should be return something like:
```js
let abdVersion = '1.2.3';
let parts = abdVersion.split('.');
return {
  versionString: abdVersion,
  versionFloat: parseFloat(adbVersion),
  major: parseInt(parts[0], 10),
  minor: parseInt(parts[1], 10),
  patch: parts[2] ? parseInt(parts[2], 10) : undefined,
```
Then
```js
let version = getAbdVersion();
version.versionString === '1.2.3';
version.versionFloat === 1.2;
version.major === 1;
version.minor === 2;
version.patch === 3;
```

Akin to https://github.com/appium/appium-xcode/blob/master/lib/xcode.js#L138-L154

vrunoa: oh got it!

test/unit/syscalls-specs.js
```diff
@@ -107,6 +107,21 @@ describe('System calls', withMocks({teen_process}, (mocks) => {
...
+      .once()
+      .withExactArgs('version')
+      .returns("Android Debug Bridge version 1.0.39\nRevision 5943271ace17-android");
+    let adbVersion = await adb.getAdbVersion();
+    adbVersion.versionString.should.equal("1.0.39");
+    adbVersion.versionFloat.should.be.within(1.0, 1.0);
+    adbVersion.major.should.equal(1);
+    adbVersion.minor.should.equal(0);
+    adbVersion.patch.should.equal(39);
+    // verify adb exec is called once, since we are caching the result
```
imurchie: This should be a second spec.
```js
it('should cache adb results', async function () {
  mocks.adb.expects("adbExec")
    .once()
    .withExactArgs('version')
    .returns("Android Debug Bridge version 1.0.39\nRevision 5943271ace17-android");
  await adb.getAdbVersion();
  await adb.getAdbVersion();
  mocks.adb.verify();
});

vrunoa: @imurchie any idea how to remove memoize cache ?

imurchie: Oh! Good point.

https://lodash.com/docs/4.17.4#memoize
"The cache is exposed as the cache property on the memoized function. Its creation may be customized by replacing the _.memoize.Cache constructor with one whose instances implement the Map method interface of clear, delete, get, has, and set."

imurchie: I would try just setting the `cache` property to `null`. If that doesn't work, `delete`ing it.

imurchie:
Published in `appium-adb@2.18.0`.

PR #223, opened 2017-04-27
Handle telnet connection error

imurchie:
I ran across an error locally. The current implementation will crash the server full stop.run


imurchie:
Hah! I thought a new tab had opened and was searching the upcoming "run for all women" event in Harlem.

PR #224, opened 2017-04-27
Support specifying multiple wait packages with a comma separated list

vikki:
I found an issue from last year - appium/appium#7390 about specifying multiple activities, but as I understand it they still have to be in the same package to match correctly. I have an app that can be launched into one of two activities in different packages, so I added it support for multiple packages.

It basically just works the same way as appWaitActivity but on the package instead of activity. It then combines each package name with each activity name and allows any of those MxN fully qualified activity names to be the wait package. Wasn't totally sure about test coverage but I've tried to keep to the style of the existing ones, though happy to tweak them if that helps.


test/unit/apk-utils-specs.js
```diff
@@ -216,6 +216,61 @@ describe('Apk-utils', () => {
       await adb.waitForActivityOrNot(pkg, '.Settings$AppDrawOverlaySettingsActivity', false);
       mocks.adb.verify();
     });
+    it('should be able to get an activity that is an inner class', async () => {
```
imurchie: This test is already in, right above.

vikki:
Argh sorry, I messed up pulling the merge of my other branch so duplicated the test! Shouldn't have done the PR so late yesterday 😪
Anyway, it should be fixed now, I can squash the second commit if you like?
Thanks for responding so quickly BTW!

imurchie:
No need to manually squash, we can do that on merge in most cases.

imurchie:
Published in `appium-adb@2.17.4`.

PR #226, opened 2017-05-02
Add send sms method

vrunoa:
Add method to send sms message to the emulator, this would be useful for apps that has sms validation. 

# Usage 

```javascript
await adb.sendSMS(4509, "Hello Appium")
```

screen shot 2017-05-02 at 1 17 27 am


lib/tools/adb-emu-commands.js
```diff
@@ -36,6 +36,25 @@ emuMethods.rotate = async function () {
...
+emuMethods.sendSMS = async function (phoneNumber, message) {
+  if (_.isUndefined(phoneNumber)) {
+    log.errorAndThrow('sendSMS phoneNumber param is undefined');
+  }
+  phoneNumber = (phoneNumber+"").replace(/\s*/, "");
+  if (!/^[\+]?[(]?[0-9]*[)]?[-\s\.]?[0-9]*[-\s\.]?[0-9]{2,}$/im.test(phoneNumber)) {
+    log.errorAndThrow(`Invalid phoneNumber param ${phoneNumber}`);
+  }
+
+  if (_.isUndefined(message) && _.isEmpty(message)) {
```
imurchie: Better to give a default and skip the `undefined` check:
```js
emuMethods.sendSMS = async function (phoneNumber, message = '') {
  // ...
  if (message === '') {
    log.errorAndThrow('Sending an SMS requires a message');
  }
  // ...
}
```

This could also go at the very top, before processing anything.

lib/tools/adb-emu-commands.js
```diff
@@ -36,6 +36,25 @@ emuMethods.rotate = async function () {
...
+    log.errorAndThrow('sendSMS message param is undefined');
+  }
+  message = _.trim(message);
```
imurchie: `message = message.trim();` will do the same thing.

lib/tools/adb-emu-commands.js
```diff
@@ -36,6 +36,25 @@ emuMethods.rotate = async function () {
   await this.adbExec(['emu', 'rotate']);
 };
 
+emuMethods.sendSMS = async function (phoneNumber, message) {
+  if (_.isUndefined(phoneNumber)) {
```
mykola-mokhnach: this verification is probably no needed - you anyway check if the number matches to regexp

lib/tools/adb-emu-commands.js
```diff
@@ -36,6 +36,25 @@ emuMethods.rotate = async function () {
...
+    log.errorAndThrow('sendSMS phoneNumber param is undefined');
+  }
+  phoneNumber = (phoneNumber+"").replace(/\s*/, "");
```
mykola-mokhnach: (phoneNumber+"") -> `${phoneNumber}`

lib/tools/adb-emu-commands.js
```diff
@@ -36,6 +36,25 @@ emuMethods.rotate = async function () {
...
+  if (!/^[\+]?[(]?[0-9]*[)]?[-\s\.]?[0-9]*[-\s\.]?[0-9]{2,}$/im.test(phoneNumber)) {
```
mykola-mokhnach: it would be more readable to move this pattern into a constant and give it appropriate name

lib/tools/adb-emu-commands.js
mykola-mokhnach: _.isEmpty returns true for undefined ;)

vrunoa:
@imurchie @mykola-mokhnach yes, this only work for emulators. On the method we have a call to `await this.verifyEmulatorConnected()`. We also have to add a validation on the android driver to avoid this call when testing on real devices.

mykola-mokhnach:
yeah, less code === less bugs

mykola-mokhnach:
@jlipps token reset is necessary in order to make sure the 'adb emu' command can be executed successfully. It's a cheap operation, but I would consider moving 
```
await this.verifyEmulatorConnected();
await this.resetTelnetAuthToken();
await this.adbExec(['emu', 'sms', 'send', phoneNumber, message]);
```
into a common method, where it's only executed once for the very first method call. There is definitely no need to execute this operation for each invocation of 'adb emu'.

jlipps:
Yes, why don't we make an `adbExecEmu` command which runs the verification and telnet reset only when necessary, and perhaps caches the last time it ran so it knows when it doesn't need to run anymore?

vrunoa:
yeah, thats a great idea, I'll create a`adbExecEmu` method for you guys to review.

lib/tools/system-calls.js
```diff
@@ -165,6 +165,31 @@ systemCallMethods.restartAdb = async function () {
...
+  } catch (e) {
+    log.warn(`Error ${e.message} while resetting the content of ${dstPath}. Ignoring resetting of emulator\'s telnet authentication token`);
+    return false;
+  }
+  return true;
+});
+
+systemCallMethods.adbExecEmu = async function (cmd) {
+  await this.verifyEmulatorConnected();
+  await this.resetTelnetAuthToken();
```
mykola-mokhnach: I think it would be better to memoize the function here

test/unit/adb-emu-commands-specs.js
```diff
@@ -108,6 +108,6 @@ describe('adb emulator commands', () => {
         await adb.sendSMS(phoneNumber, message);
         mocks.adb.verify();
       });
-    }));
+    }));
```
mykola-mokhnach: extra spaces

lib/tools/system-calls.js
vrunoa: @mykola-mokhnach how would that be ?

mykola-mokhnach: Nevermind - I was just thinking about creating a separate instance of memoized function. But this won't look good.

docs/COMMANDS.md
```diff
@@ -10,3 +15,11 @@
   "patch": 39
 }
 ```
+
+# sendSMS
+
+```javascript
+await adb.senSMS(4509, "Hello Appium");
```
imurchie: `sendSMS`

lib/tools/system-calls.js
```diff
@@ -165,6 +165,31 @@ systemCallMethods.restartAdb = async function () {
...
+  await this.adbExec(_.union(['emu'], cmd));
```
imurchie: No need for the lodash function.
```js
['emu', ...cmd]
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_operator

vrunoa: cool! didnt new about the spread operator

imurchie:
Published in `appium-adb@2.20.0`.

mykola-mokhnach:
@imurchie  @jlipps  Clients could potentially get access to these emulator methods via **mobile: emulator** interface like we do for iOS gestures

jlipps:
@mykola-mokhnach that's true, however I have always viewed the executeScript hack as a hack, a temporary way to try out functionality that we're not sure we want to keep or we don't quite know how to make a real endpoint yet. so this could be a good candidate for that but it seems pretty useful and like we will probably keep it. it's a documentation pain to make something a "mobile: " method first and then change it later. but we could do both at the same time, I suppose!

vrunoa:
@mykola-mokhnach @jlipps i'm going to add a few more methods for the emulators, do we want to make this `mobile:` approach now or should I keep adding PR for this methods one by one.

imurchie:
I think we should try to add endpoints within the `/appium/` vendor domain, and sync the clients as much as possible. This can be batched, or course. The best thing is to create an issue, and add to backlog, for the creation of the rest of the apparatus. 

Whatever the case, anything created at the `appium-adb` level will require wiring into the android and/or uiautomator2 drivers before anything can be done client-side.

PR #227, opened 2017-05-03
Support start activities that contain an inner class

vikki:
Hopefully fixing appium/appium#8359 (more details there), whereby you can't specify an inner class as the start activity, because every thing from the dollar onwards gets clipped. Fix is just to escape the dollar (awfully similar to my last PR!)

### DIFF
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -41,6 +41,8 @@ apkUtilsMethods.startApp = async function (startAppOptions = {}) {
       log.errorAndThrow("activity and pkg is required for launching application");
     }
     startAppOptions = _.clone(startAppOptions);
+    startAppOptions.activity = startAppOptions.activity.replace('$', '\\$');
+
     // initializing defaults
     _.defaults(startAppOptions, {
       waitPkg: startAppOptions.pkg,
--- a/test/functional/apk-utils-e2e-specs.js
+++ b/test/functional/apk-utils-e2e-specs.js
@@ -131,7 +131,6 @@ describe('apk utils', function () {
         .be.rejectedWith('Activity');
     });
   });
-
   it('should start activity when start activity is an inner class', async () => {
     await adb.install(contactManagerPath);
     await adb.startApp({pkg: 'com.android.settings',
@@ -140,9 +139,7 @@ describe('apk utils', function () {
     let {appPackage, appActivity} = await adb.getFocusedPackageAndActivity();
     appPackage.should.equal('com.android.settings');
     appActivity.should.equal('.Settings$NotificationAppListActivity');
-
   });
-
   it('getFocusedPackageAndActivity should be able get package and activity', async () => {
     await adb.install(contactManagerPath);
     await adb.startApp({pkg: 'com.example.android.contactmanager',
--- a/test/functional/apk-utils-e2e-specs.js
+++ b/test/functional/apk-utils-e2e-specs.js
@@ -131,6 +131,18 @@ describe('apk utils', function () {
         .be.rejectedWith('Activity');
     });
   });
+
+  it('should start activity when start activity is an inner class', async () => {
+    await adb.install(contactManagerPath);
+    await adb.startApp({pkg: 'com.android.settings',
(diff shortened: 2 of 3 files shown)


lib/tools/apk-utils.js
```diff
@@ -41,6 +41,8 @@ apkUtilsMethods.startApp = async function (startAppOptions = {}) {
       log.errorAndThrow("activity and pkg is required for launching application");
     }
     startAppOptions = _.clone(startAppOptions);
+    startAppOptions.activity = startAppOptions.activity.replace('$', '\\$');
```
mykola-mokhnach: shouldn't we check that activity property is defined first?

test/functional/apk-utils-e2e-specs.js
```diff
@@ -131,6 +131,18 @@ describe('apk utils', function () {
...
+
+  it('should start activity when start activity is an inner class', async () => {
+    await adb.install(contactManagerPath);
+    await adb.startApp({pkg: 'com.android.settings',
+      activity: '.Settings$NotificationAppListActivity'});
+
+    let {appPackage, appActivity} = await adb.getFocusedPackageAndActivity();
+    appPackage.should.equal('com.android.settings');
+    appActivity.should.equal('.Settings$NotificationAppListActivity');
+
```
mykola-mokhnach: this empty line is not needed

test/unit/apk-utils-specs.js
```diff
@@ -354,6 +354,19 @@ describe('Apk-utils', () => {
...
+      const startAppOptionsWithInnerClass = { pkg: 'pkg', activity: 'act$InnerAct'},
+            cmdWithInnerClass = ['am', 'start', '-W', '-n', 'pkg/act\\$InnerAct', '-S'];
+
+      mocks.adb.expects('getApiLevel')
+        .once().withExactArgs()
+        .returns('17');
+      mocks.adb.expects('shell')
+        .once().withExactArgs(cmdWithInnerClass)
+        .returns('');
+      (await adb.startApp(startAppOptionsWithInnerClass));
```
mykola-mokhnach: extra brackets here

test/unit/apk-utils-specs.js
```diff
@@ -354,6 +354,19 @@ describe('Apk-utils', () => {
       (await adb.startApp(startAppOptions));
       mocks.adb.verify();
     });
+    it('should call getApiLevel and shell with correct arguments when activity is inner class', async () => {
+      const startAppOptionsWithInnerClass = { pkg: 'pkg', activity: 'act$InnerAct'},
+            cmdWithInnerClass = ['am', 'start', '-W', '-n', 'pkg/act\\$InnerAct', '-S'];
```
mykola-mokhnach: I'd rather define this var in a separate line for better readability

vikki: I think I did it that way to start with but the rest of the file uses this style, e.g. line 10. Still want me to change it (I don't mind, just trying to be consistent)?

test/unit/apk-utils-specs.js
vikki: Ahh yes, they are unnecessary, but again consistent with the other tests in the block, e.g. l342 - should I remove?

lib/tools/apk-utils.js
vikki: It's covered on l40, right? So it'll throw and never get this far if we don't have an activity defined.

test/unit/apk-utils-specs.js
mykola-mokhnach: yup, remove pls

test/unit/apk-utils-specs.js
mykola-mokhnach: Up to you

lib/tools/apk-utils.js
mykola-mokhnach: yes, it's ok

imurchie:
Thanks for this!

imurchie:
Published in `appium-android-driver@1.19.0`.

PR #229, opened 2017-05-07
Add emulator power methods

vrunoa:
Add methods to change power in emulators
* _powerAC_ sets the state of the battery charger to connected or not
* _powerCapacity_ sets the battery percentage
* _powerOFF_ drains the battery so the emulator turns off. Another way of killing the emulator besides `adb emu kill`

### DIFF
--- a/lib/tools/adb-emu-commands.js
+++ b/lib/tools/adb-emu-commands.js
@@ -2,6 +2,9 @@ import log from '../logger.js';
 import _ from 'lodash';
 
 const PHONE_NUMBER_PATTERN = /^[\+]?[(]?[0-9]*[)]?[-\s\.]?[0-9]*[-\s\.]?[0-9]{2,}$/im;
+const POWER_AC_ON = 'on';
+const POWER_AC_OFF = 'off';
+const POWER_AC_STATES = [POWER_AC_ON, POWER_AC_OFF];
 let emuMethods = {};
 
 emuMethods.isEmulatorConnected = async function () {
@@ -31,6 +34,25 @@ emuMethods.rotate = async function () {
   await this.adbExecEmu(['rotate']);
 };
 
+emuMethods.powerAC = async function (state = 'on') {
+  if (POWER_AC_STATES.indexOf(state) === -1) {
+    log.errorAndThrow(`Wrong power AC state sent ${state}`);
+  }
+  await this.adbExecEmu(['power', 'ac', state]);
+};
+
+emuMethods.powerCapacity = async function (percent = 100) {
+  if (!parseInt(percent) || percent < 0 || percent > 100) {
+    log.errorAndThrow(`Wrong power percent: ${percent}`);
+  }
+  await this.adbExecEmu(['power', 'capacity', parseInt(percent)]);
+};
+
+emuMethods.powerOFF = async function () {
+  await this.powerAC(POWER_AC_OFF);
+  await this.powerCapacity(0);
+};
+
 emuMethods.sendSMS = async function (phoneNumber, message = '') {
   message = message.trim();
   if (message === "") {
@@ -42,5 +64,8 @@ emuMethods.sendSMS = async function (phoneNumber, message = '') {
   }
   await this.adbExecEmu(['sms', 'send', phoneNumber, message]);
 };
+
+emuMethods.POWER_AC_ON = POWER_AC_ON;
+emuMethods.POWER_AC_OFF = POWER_AC_OFF;
 
 export default emuMethods;
--- a/lib/tools/adb-emu-commands.js
+++ b/lib/tools/adb-emu-commands.js
@@ -4,11 +4,9 @@ import _ from 'lodash';
 const PHONE_NUMBER_PATTERN = /^[\+]?[(]?[0-9]*[)]?[-\s\.]?[0-9]*[-\s\.]?[0-9]{2,}$/im;
 
 let emuMethods = {};
-emuMethods.POWER_AC_ON = 'on';
-emuMethods.POWER_AC_OFF = 'off';
 emuMethods.POWER_AC_STATES = {
-  'POWER_AC_ON': emuMethods.POWER_AC_ON,
-  'POWER_AC_OFF': emuMethods.POWER_AC_OFF
+  'POWER_AC_ON': 'on',
+  'POWER_AC_OFF': 'off'
 };
 
 emuMethods.isEmulatorConnected = async function () {
@@ -40,7 +38,7 @@ emuMethods.rotate = async function () {
 
 emuMethods.powerAC = async function (state = 'on') {
   if (_.values(emuMethods.POWER_AC_STATES).indexOf(state) === -1) {
-    log.errorAndThrow(`Wrong power AC state sent '${state}', possible values: [${emuMethods.POWER_AC_ON}, ${emuMethods.POWER_AC_OFF}]`);
+    log.errorAndThrow(`Wrong power AC state sent '${state}', possible values: ${JSON.stringify(this.POWER_AC_STATES)}]`);
   }
   await this.adbExecEmu(['power', 'ac', state]);
 };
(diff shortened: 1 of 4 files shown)


lib/tools/adb-emu-commands.js
```diff
@@ -31,6 +34,25 @@ emuMethods.rotate = async function () {
   await this.adbExecEmu(['rotate']);
 };
 
+emuMethods.powerAC = async function (state = 'on') {
+  if (POWER_AC_STATES.indexOf(state) === -1) {
+    log.errorAndThrow(`Wrong power AC state sent ${state}`);
```
mykola-mokhnach: please add the list of supported states to error messages

lib/tools/adb-emu-commands.js
```diff
@@ -31,6 +34,25 @@ emuMethods.rotate = async function () {
...
+  }
+  await this.adbExecEmu(['power', 'ac', state]);
+};
+
+emuMethods.powerCapacity = async function (percent = 100) {
+  if (!parseInt(percent) || percent < 0 || percent > 100) {
```
mykola-mokhnach: parseInt method should always contain radix value as the second parameter, which is **10** in this particular case.

lib/tools/adb-emu-commands.js
```diff
@@ -31,6 +34,25 @@ emuMethods.rotate = async function () {
...
+    log.errorAndThrow(`Wrong power percent: ${percent}`);
+  }
+  await this.adbExecEmu(['power', 'capacity', parseInt(percent)]);
```
mykola-mokhnach: can we parseInt only once?

lib/tools/adb-emu-commands.js
```diff
@@ -43,4 +65,7 @@ emuMethods.sendSMS = async function (phoneNumber, message = '') {
   await this.adbExecEmu(['sms', 'send', phoneNumber, message]);
 };
 
+emuMethods.POWER_AC_ON = POWER_AC_ON;
```
mykola-mokhnach: these constant could be defined only once

lib/tools/adb-emu-commands.js
```diff
@@ -31,6 +34,25 @@ emuMethods.rotate = async function () {
...
+emuMethods.powerAC = async function (state = 'on') {
+  if (POWER_AC_STATES.indexOf(state) === -1) {
+    log.errorAndThrow(`Wrong power AC state sent ${state}`);
+  }
+  await this.adbExecEmu(['power', 'ac', state]);
+};
+
+emuMethods.powerCapacity = async function (percent = 100) {
+  if (!parseInt(percent) || percent < 0 || percent > 100) {
+    log.errorAndThrow(`Wrong power percent: ${percent}`);
```
mykola-mokhnach: it would be better to say: The percentage value should be valid integer between 0 and 100

lib/tools/adb-emu-commands.js
imurchie: `parseInt(0, 10)` will be `false`. Do we want to be able to set it to `0`? 

I think you want something like `if (_.isNaN(percent) || percent < 0 || percent > 100) {` (assuming that `percent` has already been parsed).

lib/tools/adb-emu-commands.js
imurchie: I agree. `percent = parseInt(percent, 10)` at the beginning of the function would be better.

lib/tools/adb-emu-commands.js
```diff
@@ -2,7 +2,14 @@ import log from '../logger.js';
 import _ from 'lodash';
 
 const PHONE_NUMBER_PATTERN = /^[\+]?[(]?[0-9]*[)]?[-\s\.]?[0-9]*[-\s\.]?[0-9]{2,}$/im;
+
 let emuMethods = {};
+emuMethods.POWER_AC_ON = 'on';
```
vrunoa: @imurchie @mykola-mokhnach how can I declare this to be const ?

imurchie: Within the object? There is no way, other than to make it a getter with no setter:
```js
let emuMethods = {
  get POWER_AC_ON {
    return 'on';
  },
};
```

But in JavaScript const is not much use. It is fine just to have it there and assume no one is going to modify it
```js
let emuMethods = {
  POWER_AC_ON: 'on',
};
```

imurchie: In Appium we have no usually kept constants on objects. Instead we just export the constants from the module and allow folks to import them when they need them.
```js
const POWER_AC_ON = 'on';

let emuMethods = {};

// ...

export default emuMethods;
export { emuMethods, POWER_AC_ON };
```

lib/tools/adb-emu-commands.js
```diff
@@ -31,6 +38,26 @@ emuMethods.rotate = async function () {
   await this.adbExecEmu(['rotate']);
 };
 
+emuMethods.powerAC = async function (state = 'on') {
+  if (_.values(emuMethods.POWER_AC_STATES).indexOf(state) === -1) {
+    log.errorAndThrow(`Wrong power AC state sent '${state}', possible values: [${emuMethods.POWER_AC_ON}, ${emuMethods.POWER_AC_OFF}]`);
```
mykola-mokhnach: the enumeration can be replaced with

```javascript
_.pairs(emuMethods).filter((x) => x[0].startsWith('POWER_AC_')).map((x) => x[1])
```

lib/tools/adb-emu-commands.js
```diff
@@ -31,6 +38,26 @@ emuMethods.rotate = async function () {
...
+  }
+  await this.adbExecEmu(['power', 'ac', state]);
+};
+
+emuMethods.powerCapacity = async function (percent = 100) {
+  percent = parseInt(percent, 10);
+  if (_.isNaN(percent) || percent < 0 || percent > 100) {
+    log.errorAndThrow(`The percentage value should be valid integer between 0 and 100`);
+  }
+  await this.adbExecEmu(['power', 'capacity', percent]);
```
mykola-mokhnach: `${percent}` ?

imurchie: I don't think that is necessary. A number should eventually get coerced into a string.

lib/tools/adb-emu-commands.js
```diff
@@ -2,7 +2,12 @@ import log from '../logger.js';
 import _ from 'lodash';
 
 const PHONE_NUMBER_PATTERN = /^[\+]?[(]?[0-9]*[)]?[-\s\.]?[0-9]*[-\s\.]?[0-9]{2,}$/im;
+
 let emuMethods = {};
+emuMethods.POWER_AC_STATES = {
+  'POWER_AC_ON': 'on',
```
imurchie: No need for Python-ish quotes around keys, unless they are invalid JS variable names.

lib/tools/adb-emu-commands.js
vrunoa: we dont need to set it as string with quotes

lib/tools/adb-emu-commands.js
```diff
@@ -31,6 +36,26 @@ emuMethods.rotate = async function () {
...
+emuMethods.powerAC = async function (state = 'on') {
+  if (_.values(emuMethods.POWER_AC_STATES).indexOf(state) === -1) {
+    log.errorAndThrow(`Wrong power AC state sent '${state}'. Supported values: ${_.values(emuMethods.POWER_AC_STATES)}]`);
+  }
+  await this.adbExecEmu(['power', 'ac', state]);
+};
+
+emuMethods.powerCapacity = async function (percent = 100) {
+  percent = parseInt(percent, 10);
+  if (_.isNaN(percent) || percent < 0 || percent > 100) {
```
mykola-mokhnach: nit: It would be better here to use the standard JS's **isNaN** instead of the one from lodash lib.

imurchie:
Published in `appium-adb@2.21.0`.

PR #230, opened 2017-05-10
use newest sdk emulator cli

vrunoa:
The newest android sdk have separated the emulator cli from tools, so we should check if this is available under the own emulator folder before checking on tools.


vrunoa:
No, if the emulator folder is not there it takes the one inside the `tools` folder, I just try it by removing the emulator folder from my laptop.

imurchie:
Published in `appium-adb@2.21.1`.

PR #232, opened 2017-05-11
move root, unroot into their own methods

vrunoa:
Move `root`, `unroot` into their own methods so we can call them before running _adb shell_ commands that need to be root.

### DIFF
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -572,8 +572,8 @@ systemCallMethods.root = async function () {
 systemCallMethods.unroot = async function () {
   try {
     await exec(this.executable.path, ['unroot']);
-  } catch (err) {
     this.adbdRooted = false;
+  } catch (err) {
     log.warn(`Unable to unroot adb daemon: '${err.message}'. Continuing`);
   }
 };
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -557,21 +557,19 @@ systemCallMethods.reboot = async function (retries = DEFAULT_ADB_REBOOT_RETRIES)
 };
 
 systemCallMethods.root = async function () {
-  try {
-    let process = await exec(this.executable.path, ['root']);
-    return !!process.code;
-  } catch (err) {
-    log.warn(`Unable to root adb daemon: '${err.message}'. Continuing`);
+  let process = await exec(this.executable.path, ['root']);
+  if (process.code !== 0) {
+    log.warn(`Unable to root adb daemon: '${process.stderr.message}'. Continuing`);
   }
+  return !process.code;
 };
 
 systemCallMethods.unroot = async function () {
-  try {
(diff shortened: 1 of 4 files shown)


lib/tools/system-calls.js
```diff
@@ -554,16 +554,30 @@ systemCallMethods.reboot = async function (retries = DEFAULT_ADB_REBOOT_RETRIES)
...
+    if (this.adbdRooted) {
+      this.unroot();
     }
   }
 };
 
+systemCallMethods.root = async function () {
+  try {
+    await exec(this.executable.path, ['root']);
+    this.adbdRooted = true;
```
mykola-mokhnach: I think there is no point to maintain **this.adbdRooted** flag in these method if we ignore errors. I's rather leave it there where it was
Also, it might be useful to return true/false from these methods to be able to understand whether the operation was successful or not from the client code

vrunoa: agreed, I'll check the exec ouput to verify the emu(or device) got rooted

lib/tools/system-calls.js
```diff
@@ -554,14 +552,24 @@ systemCallMethods.reboot = async function (retries = DEFAULT_ADB_REBOOT_RETRIES)
...
-      } catch (err) {
-        log.warn(`Unable to unroot adb daemon: '${err.message}'. Continuing`);
-      }
-    }
+    this.unroot();
+  }
+};
+
+systemCallMethods.root = async function () {
+  let process = await exec(this.executable.path, ['root']);
```
mykola-mokhnach: this will just throw an exception if the exit code is not zero

test/unit/syscalls-specs.js
```diff
@@ -164,7 +164,8 @@ describe('System calls',  withMocks({adb, B, teen_process}, (mocks) => {
   });
   it('reboot should restart adbd as root if necessary', async () => {
     mocks.teen_process.expects("exec")
-      .once().withExactArgs(adb.executable.path, ['root']);
+      .once().withExactArgs(adb.executable.path, ['root'])
+      .returns({stdout:"restarting adbd as root", stderr: undefined, code:0});
```
jlipps: you'd think lint would barf on `code:0` (should be `code: 0`)

lib/tools/system-calls.js
```diff
@@ -554,13 +552,23 @@ systemCallMethods.reboot = async function (retries = DEFAULT_ADB_REBOOT_RETRIES)
...
-        log.warn(`Unable to unroot adb daemon: '${err.message}'. Continuing`);
-      }
-    }
+    this.unroot();
+  }
+};
+
+systemCallMethods.root = async function () {
+  try {
+    await exec(this.executable.path, ['root']);
```
mykola-mokhnach: how about returning _true_ in case of success and false if _failed_?

imurchie:
Published in `appium-adb@2.22.0`.

PR #233, opened 2017-05-16
Split long permission granting queries into chunks to avoid adb failures

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/8461

### DIFF
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -121,8 +121,19 @@ methods.grantAllPermissions = async function (pkg, apk) {
     }
     log.debug(`Got the following command chunks to execute: ${cmds}`);
     let result = true;
+    let lastError = null;
     for (let cmd of cmds) {
-      result = await this.shell(cmd) && result;
+      try {
+        result = await this.shell(cmd) && result;
+      } catch (e) {
+        // this is to give the method a chance to assign all the requestsed permissions
+        // before to quit in case we'd like to ignore the error on the higher level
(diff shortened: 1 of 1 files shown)


lib/tools/adb-commands.js
```diff
@@ -97,13 +98,33 @@ methods.grantAllPermissions = async function (pkg, apk) {
...
+      }
+      cmdChunk = cmdChunk.concat(nextCmd);
+    }
+    if (cmdChunk.length) {
+      cmds.push(cmdChunk);
+    }
+    log.debug(`Got the following command chunks to execute: ${cmds}`);
+    let result = true;
+    for (let cmd of cmds) {
+      result = await this.shell(cmd) && result;
```
vrunoa: @mykola-mokhnach would it be good to know which permission failed, what do you think ?

mykola-mokhnach: this will be visible anyway, because shell method throws an exception if the return code from adb is not equal to 0

vrunoa: oh ok. Nevermind then.

mykola-mokhnach: I have added error handling there for a situation when adb returns non-zero code for one of chunks located closer to the head of the queue, but there are more at the tail. This gives a possibility to process the whole permissions queue and ignore the error on the higher level if necessary.

imurchie:
Published in `appium-adb@2.23.0`

PR #234, opened 2017-05-26
throw error on avd failing to launch

vrunoa:
Hey guys I've been playing around with avdArgs and I notice that when I put a wrong param the avd fail to launch with for ex. `unknown option --no-window` but the script continues without caring if the process exited with 1. Why are we waiting for the retry to throw an exception and not doing it immediately ? Is this a good approach to exit when the avd fails to launch ?

### DIFF
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -436,7 +436,7 @@ systemCallMethods.launchAVD = async function (avdName, avdArgs, language, countr
   });
   proc.on('exit', (code, signal) => {
     if (code !== 0) {
-      log.errorAndThrow(`[AVD OUTPUT] avd exit with code ${code}, signal ${signal}`);
+      log.errorAndThrow(`Emulator avd ${avdName} exit with code ${code}, signal ${signal}`);
     }
   });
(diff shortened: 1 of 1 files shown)


lib/tools/system-calls.js
```diff
@@ -434,6 +434,11 @@ systemCallMethods.launchAVD = async function (avdName, avdArgs, language, countr
   proc.on('output', (stdout, stderr) => {
     log.info(`[AVD OUTPUT] ${stdout || stderr}`);
   });
+  proc.on('exit', (code, signal) => {
+    if (code !== 0) {
+      log.errorAndThrow(`[AVD OUTPUT] avd exit with code ${code}, signal ${signal}`);
```
imurchie: Not sure about the `[AVD OUTPUT]` part here, since this is not technically AVD output.

vrunoa: maybe `[AVD ERROR]` ?

imurchie: Or no prefix. In other places we just log and throw.

imurchie:
Published in `appium-adb@2.23.4`.

PR #235, opened 2017-05-30
Pass default arguments to adb devices

serggl:
When adb is started on a different host/port `getConnectedDevices` will fail.
Now it should be fixed


imurchie:
Published in `appium-adb@2.23.1`.

PR #236, opened 2017-05-30
Handle error from running as root

imurchie:
Sometimes the `root` command fails but doesn't error out. So check the output to get the logging right.

### DIFF
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -225,6 +225,7 @@ systemCallMethods.adbExec = async function (cmd, opts = {}) {
         stdout = stdout.replace(linkerWarningRe, '').trim();
         return stdout;
       }
+
       throw new Error(`Error executing adbExec. Original error: '${e.message}'; ` +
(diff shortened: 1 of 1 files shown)


lib/tools/system-calls.js
```diff
@@ -558,7 +559,13 @@ systemCallMethods.reboot = async function (retries = DEFAULT_ADB_REBOOT_RETRIES)
 
 systemCallMethods.root = async function () {
   try {
-    await exec(this.executable.path, ['root']);
+    let {stdout} = await exec(this.executable.path, ['root']);
+
+    // on real devices in some situations we get an error in the stdout
+    if (stdout.indexOf('adbd cannot run as root') !== -1) {
```
mykola-mokhnach: strange that exec does not return non-zero return code (and throws an exception) in such case

imurchie: It is `adb` that is not returning a non-zero code. All `exec` does is look at what is execed. It cannot, _a priori_, know that any particular output indicates a failure.

PR #237, opened 2017-06-26
Add killServer and fileSize methods

imurchie:
Both are being used as _ad hoc_ commands elsewhere, so should be part of the API.

### DIFF
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -160,10 +160,15 @@ systemCallMethods.restartAdb = async function () {
 
   log.debug('Restarting adb');
   try {
-    await exec(this.executable.path, ['kill-server']);
+    await this.killServer();
   } catch (e) {
     log.error("Error killing ADB server, going to see if it's online anyway");
   }
+};
+
+systemCallMethods.killServer = async function () {
+  log.debug(`Killing adb server on port ${this.adbPort}`);
+  await exec(this.executable.path, ['kill-server']);
 };
 
 systemCallMethods.resetTelnetAuthToken = _.memoize(async function () {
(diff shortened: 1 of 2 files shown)


lib/tools/system-calls.js
```diff
@@ -608,4 +613,28 @@ systemCallMethods.ls = async function (remotePath) {
   }
 };
 
+systemCallMethods.fileSize = async function (remotePath) {
+  try {
+    let stdout = await this.shell(['ls', '-la', remotePath]);
+    let lines = stdout.split('\n');
+    lines = lines.map((l) => l.trim())
```
mykola-mokhnach: I see some code duplication here %)

lib/tools/system-calls.js
```diff
@@ -608,4 +613,28 @@ systemCallMethods.ls = async function (remotePath) {
...
+      throw new Error(`Remote path is not a file`);
+    }
+    // https://regex101.com/r/fOs4P4/1
+    let match = /[-rwxd]{10}\s*\S*\s*\S*\s*\S*\s*(\d*)\s*\d{4}-\d{2}-\d{2}\s\d{2}:\d{2}\s[\S\.]*/.exec(lines[0]);
+    if (!match || _.isNaN(parseInt(match[1], 10))) {
+      throw new Error(`Unable to parse size from list output: '${lines[0]}'`);
+    }
+    return parseInt(match[1], 10);
+  } catch (err) {
+    if (err.message.indexOf('No such file or directory') === -1) {
```
mykola-mokhnach: Can we somehow improve the flow where we throw and catch custom errors? Probably, replace them with if conditions where possible?

lib/tools/system-calls.js
```diff
@@ -615,24 +616,17 @@ systemCallMethods.ls = async function (remotePath) {
 
 systemCallMethods.fileSize = async function (remotePath) {
   try {
-    let stdout = await this.shell(['ls', '-la', remotePath]);
-    let lines = stdout.split('\n');
-    lines = lines.map((l) => l.trim())
-      .filter(Boolean)
-      .filter((l) => l.indexOf('No such file') === -1);
-    if (lines.length !== 1) {
+    let files = await this.ls(remotePath, ['-la']);
```
mykola-mokhnach: should this call be also inside try/catch block?

lib/tools/system-calls.js
```diff
@@ -615,24 +616,17 @@ systemCallMethods.ls = async function (remotePath) {
...
+    if (files.length !== 1) {
       throw new Error(`Remote path is not a file`);
```
mykola-mokhnach: should the actual path value be present in this error message?

imurchie: No, this gets wrapped in an error with that information.

lib/tools/system-calls.js
```diff
@@ -608,4 +614,21 @@ systemCallMethods.ls = async function (remotePath) {
   }
 };
 
+systemCallMethods.fileSize = async function (remotePath) {
+  try {
+    let files = await this.ls(remotePath, ['-la']);
```
mykola-mokhnach: We could also use the following cmdline to extract file size: `adb shell ls -la /sdcard/view.xml | tr -s ' ' | cut -d' ' -f 5`
All these shell commands are available even in Android 4.2

imurchie: I guess we could? But this makes less sense to me than a regular expression, and also `exec` doesn't handle pipes very well.

mykola-mokhnach: this pipe should be handled by android shell rather than exec itself, so should not be a problem. 
But yeah, we the current implementation has its own pros

PR #238, opened 2017-07-03
Set root, unroot for api 26

vrunoa:
On Android O(api 26) i'm seeing setprop is failing is adb is not root. I modified setDeviceProperty method to set root, unroot in order to use setprop.

### DIFF
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -707,8 +707,16 @@ methods.getDeviceProperty = async function (property) {
 };
 
 methods.setDeviceProperty = async function (prop, val) {
+  if (this.getApiLevel() >= 26 ) {
+    log.debug(`Android O need to be on adb root to run setprop`);
+    await this.root();
+  }
   log.debug(`Setting device property '${prop}' to '${val}'`);
   await this.shell(['setprop', prop, val]);
+  if (this.getApiLevel() >= 26 ) {
+    log.debug(`Removing adb root for setprop`);
+    await this.unroot();
+  }
 };
 
 methods.getDeviceSysLanguage = async function () {
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -709,7 +709,7 @@ methods.getDeviceProperty = async function (property) {
 methods.setDeviceProperty = async function (prop, val) {
   let apiLevel = await this.getApiLevel(), err;
   if (apiLevel >= 26 ) {
-    log.debug(`Android O need to be on adb root to run setprop`);
+    log.debug(`Running adb root, Android O needs adb to be rooted to setDeviceProperty`);
     await this.root();
   }
   log.debug(`Setting device property '${prop}' to '${val}'`);
@@ -719,7 +719,7 @@ methods.setDeviceProperty = async function (prop, val) {
     err = e;
   }
   if (apiLevel >= 26 ) {
-    log.debug(`Removing adb root for setprop`);
+    log.debug(`Removing adb root for setDeviceProperty`);
     await this.unroot();
   }
   if (!_.isUndefined(err)) {
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -707,13 +707,14 @@ methods.getDeviceProperty = async function (property) {
 };
 
 methods.setDeviceProperty = async function (prop, val) {
-  if (this.getApiLevel() >= 26 ) {
+  let apiLevel = await this.getApiLevel();
+  if (apiLevel >= 26 ) {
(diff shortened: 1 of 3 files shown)


lib/tools/adb-commands.js
```diff
@@ -707,8 +707,17 @@ methods.getDeviceProperty = async function (property) {
 };
 
 methods.setDeviceProperty = async function (prop, val) {
+  let apiLevel = await this.getApiLevel();
+  if (apiLevel >= 26 ) {
```
mykola-mokhnach: probably, it makes sense to enclose this into try .. finally

vrunoa: @mykola-mokhnach what do you think of this approach ? I would raise an exception after running `.unroot`.

lib/tools/adb-commands.js
```diff
@@ -707,8 +707,24 @@ methods.getDeviceProperty = async function (property) {
 };
 
 methods.setDeviceProperty = async function (prop, val) {
+  let apiLevel = await this.getApiLevel(), err;
```
mykola-mokhnach: Why cannot one simply write `let err;` to make it more readable?

mykola-mokhnach: also, there is a general recommendation to define variables close to the place where they are first used

lib/tools/adb-commands.js
```diff
@@ -707,8 +707,24 @@ methods.getDeviceProperty = async function (property) {
...
+    await this.shell(['setprop', prop, val]);
+  } catch (e) {
+    err = e;
+  }
+  if (apiLevel >= 26 ) {
+    log.debug(`Removing adb root for setDeviceProperty`);
+    await this.unroot();
+  }
+  if (!_.isUndefined(err)) {
+    log.errorAndThrow(`Error trying to set device property. Original error: ${err.message}`);
```
mykola-mokhnach: I'd rather keep the stack trace. `throw err` should suffice

lib/tools/adb-commands.js
```diff
@@ -707,8 +707,24 @@ methods.getDeviceProperty = async function (property) {
...
+  try {
+    await this.shell(['setprop', prop, val]);
+  } catch (e) {
+    err = e;
+  }
+  if (apiLevel >= 26 ) {
+    log.debug(`Removing adb root for setDeviceProperty`);
+    await this.unroot();
+  }
+  if (!_.isUndefined(err)) {
```
mykola-mokhnach: nit: there is `isError` test in lodash lib

lib/tools/adb-commands.js
vrunoa: adbExec already throws an error that we catch here, in order to make sure `.unroot` is executed. Should I use throw err instead of log.errorAndThrow ?

mykola-mokhnach: throwing the error here via `log.errorAndThrow` will show you the current function as the the stacktrace target, but the actual exception happens inside `this.shell` call. So it will be complicated to find the real problem source. On the other side `throw` preserves the original stracktrace, so it will be visible where the error actually happened

lib/tools/adb-commands.js
```diff
@@ -707,8 +707,25 @@ methods.getDeviceProperty = async function (property) {
...
+    await this.shell(['setprop', prop, val]);
+  } catch (e) {
+    err = e;
+  }
+  if (apiLevel >= 26 ) {
+    log.debug(`Removing adb root for setDeviceProperty`);
+    await this.unroot();
+  }
+  if (_.isError(err)) {
+    throw new Error(err);
```
mykola-mokhnach: `throw err` is enough

lib/tools/adb-commands.js
```diff
@@ -707,8 +707,25 @@ methods.getDeviceProperty = async function (property) {
...
+  try {
+    await this.shell(['setprop', prop, val]);
+  } catch (e) {
+    err = e;
+  }
+  if (apiLevel >= 26 ) {
+    log.debug(`Removing adb root for setDeviceProperty`);
+    await this.unroot();
+  }
+  if (_.isError(err)) {
```
imurchie: Just checking for the truthiness of `err` is enough here. If `err` is anything, it is an `Error` object.

mykola-mokhnach: @imurchie it was me who asked for this change. IMHO positive conditions are easier to read than negative (it was `!._isUndefined(err)`). But yeah, `if (err)` is also very readable

imurchie: I would prefer `if (err) throw err; // eslint-disable-line curly` myself.

imurchie:
Published in `appium-adb@2.24.0`

PR #239, opened 2017-07-04
Fix returned value for getAndroidPlatformAndPath to avoid deconstruction error

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/8773


imurchie:
What is wrong with updating `getAndroidPlatformAndPath` to return `{platform: null, platformPath: null}`, and leaving the calling code like it is?

mykola-mokhnach:
@imurchie It will require more changes, since functional and unit tests rely on it. I've just selected the shortest way ;)

imurchie:
I've done that work under https://github.com/appium/appium-adb/pull/241

mykola-mokhnach:
Thanks @imurchie

PR #240, opened 2017-07-07
Simplify filesize regexp

imurchie:
Different versions of Android apparently have different formats of `ls -la` output, so reduce the regexp to checking file size and date, which appear to always be next to each other.

In regard to https://github.com/appium/appium/issues/8800

### DIFF
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -620,8 +620,8 @@ systemCallMethods.fileSize = async function (remotePath) {
     if (files.length !== 1) {
       throw new Error(`Remote path is not a file`);
     }
-    // https://regex101.com/r/fOs4P4/2
-    let match = /\s(\d+)\s\d{4}-\d{2}-\d{2}/.exec(files[0]);
+    // https://regex101.com/r/fOs4P4/3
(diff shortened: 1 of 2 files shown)


lib/tools/system-calls.js
```diff
@@ -620,8 +620,8 @@ systemCallMethods.fileSize = async function (remotePath) {
     if (files.length !== 1) {
       throw new Error(`Remote path is not a file`);
     }
-    // https://regex101.com/r/fOs4P4/1
-    let match = /[-rwxd]{10}\s*\S*\s*\S*\s*\S*\s*(\d*)\s*\d{4}-\d{2}-\d{2}\s\d{2}:\d{2}\s[\S\.]*/.exec(files[0]);
+    // https://regex101.com/r/fOs4P4/2
+    let match = /\s(\d+)\s\d{4}-\d{2}-\d{2}/.exec(files[0]);
```
mykola-mokhnach: is there always 1 space? I'd rather replace with `\s+`

imurchie: Between the two? Sure.

imurchie:
Published in `appium-adb@2.23.8`.

PR #241, opened 2017-07-12
Fix return value on getAndroidPlatformAndPath

imurchie:
Rather than returning `null`, which causes the calling code to fail, return an object whose properties are null.

Also fix the tests, since setting `process.env.ANDROID_HOME` to `null` is the wrong technique (from the [docs](https://nodejs.org/api/process.html#process_process_env): "Assigning a property on process.env will implicitly convert the value to a string.").

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -32,8 +32,7 @@ async function getDirectories (rootPath) {
 async function getAndroidPlatformAndPath () {
   const androidHome = process.env.ANDROID_HOME;
   if (!_.isString(androidHome)) {
-    log.error("ANDROID_HOME environment variable was not exported!");
-    return {platform: null, platformPath: null};
+    log.errorAndThrow("ANDROID_HOME environment variable was not exported");
   }
 
   // get the latest platform and path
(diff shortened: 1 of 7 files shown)


lib/helpers.js
```diff
@@ -32,8 +32,8 @@ async function getDirectories (rootPath) {
 async function getAndroidPlatformAndPath () {
   const androidHome = process.env.ANDROID_HOME;
   if (!_.isString(androidHome)) {
-    log.error("ANDROID_HOME was not exported!");
-    return null;
+    log.error("ANDROID_HOME environment variable was not exported!");
+    return {platform: null, platformPath: null};
```
mykola-mokhnach: Can we throw the error here immediately instead of retuning null values? Further testing does not have much sense if this verification fails

imurchie: We could. I have no idea why it is like this. The function is not exported from the package, so nothing else depends on it.

mykola-mokhnach: Also, we can add better verification and check if the path set in ANDROID_HOME really exists on the filesystem

lib/helpers.js
```diff
@@ -44,7 +43,7 @@ async function getAndroidPlatformAndPath () {
       return {platform, platformPath};
     }
   }
-  return null;
+  return {platform: null, platformPath: null};
```
mykola-mokhnach: can be also `{}` for totally lazy people :)

imurchie: Yeah. I opted to be explicit, for self-documenting purposes. :)

PR #243, opened 2017-07-18
add await to  getPackageInfo method

cool-firer:
it is necessasry to add await symbol, cause it is asynchronous


mykola-mokhnach:
good catch

imurchie:
Published in `appium-adb@2.24.2`.

PR #245, opened 2017-07-21
Add netspeed method for emulators

vrunoa:
Add a method to change the emulator network speed emulation.

### DIFF
--- a/lib/tools/adb-emu-commands.js
+++ b/lib/tools/adb-emu-commands.js
@@ -25,6 +25,18 @@ emuMethods.GSM_VOICE_STATES = {
 };
 emuMethods.GSM_SIGNAL_STRENGTHS = [0, 1, 2, 3, 4];
 
+emuMethods.NETWORK_SPEED = {
+  GSM: 'gsm', // GSM/CSD (up: 14.4, down: 14.4).
+  SCSD: 'scsd', // HSCSD (up: 14.4, down: 57.6).
+  GPRS: 'gprs', // GPRS (up: 28.8, down: 57.6).
+  EDGE: 'edge', // EDGE/EGPRS (up: 473.6, down: 473.6).
+  UMTS: 'umts', // UMTS/3G (up: 384.0, down: 384.0).
+  HSDPA: 'hsdpa', // HSDPA (up: 5760.0, down: 13,980.0).
+  LTE: 'lte', // LTE (up: 58,000, down: 173,000).
+  EVDO: 'evdo', // EVDO (up: 75,000, down: 280,000).
+  FULL: 'full' // No limit, the default (up: 0.0, down: 0.0).
+};
+
 emuMethods.isEmulatorConnected = async function () {
   let emulators = await this.getConnectedEmulators();
   return !!_.find(emulators, (x) => x && x.udid === this.curDeviceId);
@@ -110,6 +122,14 @@ emuMethods.gsmVoice = async function (state = 'on') {
     log.errorAndThrow(`Invalid gsm voice state param ${state}. Supported values: ${_.values(emuMethods.GSM_VOICE_STATES)}`);
   }
   await this.adbExecEmu(['gsm', 'voice', state]);
+};
+
+emuMethods.networkSpeed = async function (speed = 'full') {
+  // network speed <speed> allows you to set the network speed emulation.
+  if (_.values(emuMethods.NETWORK_SPEED).indexOf(speed) === -1) {
+    log.errorAndThrow(`Invalid network speed param ${speed}. Supported values: ${_.values(emuMethods.NETWORK_SPEED)}`);
+  }
+  await this.adbExecEmu(['network', 'speed', speed]);
 };
 
 export default emuMethods;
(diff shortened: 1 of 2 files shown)


lib/tools/adb-emu-commands.js
```diff
@@ -112,4 +124,12 @@ emuMethods.gsmVoice = async function (state = 'on') {
   await this.adbExecEmu(['gsm', 'voice', state]);
 };
 
+emuMethods.networkSpeed = async function (speed = 'full') {
```
mykola-mokhnach: shouldn't it be called setNetworkSpeed?

test/unit/adb-emu-commands-specs.js
```diff
@@ -256,6 +256,129 @@ describe('adb emulator commands', () => {
         mocks.adb.verify();
       });
     }));
+    describe("network speed method", withMocks({adb}, (mocks) => {
+      it("should throw exception on invalid speed", async () => {
+        await adb.networkSpeed('light').should.eventually.be.rejectedWith("Invalid network speed");
+        mocks.adb.verify();
+      });
+      it("should set network speed to gsm", async () => {
```
mykola-mokhnach: these all verifications can be done in a single loop as for me

lib/tools/adb-emu-commands.js
vrunoa: we called the other methods without the _set_ prefix; `gsmVoice`, `powerAC`...

test/unit/adb-emu-commands-specs.js
```diff
@@ -256,6 +256,38 @@ describe('adb emulator commands', () => {
...
+        'scsd': adb.NETWORK_SPEED.SCSD,
+        'gprs': adb.NETWORK_SPEED.GPRS,
+        'edge': adb.NETWORK_SPEED.EDGE,
+        'umts': adb.NETWORK_SPEED.UMTS,
+        'hsdpa': adb.NETWORK_SPEED.HSDPA,
+        'lte': adb.NETWORK_SPEED.LTE,
+        'evdo': adb.NETWORK_SPEED.EVDO,
+        'full': adb.NETWORK_SPEED.FULL
+      };
+      _.keys(speeds).forEach( (s) => {
```
mykola-mokhnach: for-of is preferred over forEach

test/unit/adb-emu-commands-specs.js
```diff
@@ -256,6 +256,38 @@ describe('adb emulator commands', () => {
...
+        it("should set network speed correctly", async () => {
```
mykola-mokhnach: I'm not sure whether dynamic testcases generation is supported this way. @imurchie ?
And even if it is then it would be useful to have the corresponding tested option name present in the title

vrunoa: @mykola-mokhnach I add the network speed to the test title, if it fails it shows:

``` javascript
1) adb emulator commands emu network speed method should set network speed(scsd) correctly:
     ExpectationError: Unexpected call: adbExec(["emu", "network", "speed", "scsd"])
    Expected adbExec(["emu", "network", "speed", "gsm"]) once (never called)
```

mykola-mokhnach: yep, that is why it's better to have the loop inside it(), but not outside of it.

test/unit/adb-emu-commands-specs.js
```diff
@@ -256,6 +256,38 @@ describe('adb emulator commands', () => {
...
+        'scsd': adb.NETWORK_SPEED.SCSD,
+        'gprs': adb.NETWORK_SPEED.GPRS,
+        'edge': adb.NETWORK_SPEED.EDGE,
+        'umts': adb.NETWORK_SPEED.UMTS,
+        'hsdpa': adb.NETWORK_SPEED.HSDPA,
+        'lte': adb.NETWORK_SPEED.LTE,
+        'evdo': adb.NETWORK_SPEED.EVDO,
+        'full': adb.NETWORK_SPEED.FULL
+      };
+      for (let s of _.keys(speeds)) {
```
imurchie: This can be something like
```js
for (let [key, value] of _.pairs(speeds)) {
  it(`should set network speed(${key}) correctly`, async () => {
    // ...
    mocks.adb.expects("adbExec")
      .once().withExactArgs(["emu", "network", "speed", value])
      .returns();
    // ...
  });
}

File: test/unit/adb-emu-commands-specs.js [hunk 6]
```diff
@@ -256,6 +256,38 @@ describe('adb emulator commands', () => {
         mocks.adb.verify();
       });
     }));
+    describe("network speed method", withMocks({adb}, (mocks) => {
+      it("should throw exception on invalid speed", async () => {
+        await adb.networkSpeed('light').should.eventually.be.rejectedWith("Invalid network speed");
+        mocks.adb.verify();
+      });
+      let speeds = {
+        'gsm': adb.NETWORK_SPEED.GSM,
```
imurchie: But I also don't really understand what this is doing. Doesn't `adb.NETWORK_SPEED.GSM` equal 'gsm'? Would that not mean that `speeds['gsm']` would equal `gsm`? So why not just have the values from `adb.NETWORK_SPEED`?

Here is the `speeds` object:
```js
{ gsm: 'gsm',
  scsd: 'scsd',
  gprs: 'gprs',
  edge: 'edge',
  umts: 'umts',
  hsdpa: 'hsdpa',
  lte: 'lte',
  evdo: 'evdo',
  full: 'full' 
}
```

imurchie:
Published in `appium-adb@2.25.0`.

PR #246, opened 2017-08-01
Make sure adb logcat resolves

imurchie:
Currently, if `adb logcat` fails with a startup error like "execvp", we reject the promise and throw in the calling code. But if startup fails for any other reason we do nothing, and the promise is never resolved or rejected, so the session hangs.

So, keep track of whether we've started, so that we can log that it failed in this condition, and resolve the promise. Things will work fine but there will be no logcat capture.

Should fix https://github.com/appium/appium/issues/8743

### DIFF
--- a/lib/logcat.js
+++ b/lib/logcat.js
@@ -30,10 +30,7 @@ class Logcat {
         log.error(`Logcat terminated with code ${code}, signal ${signal}`);
         this.proc = null;
         if (!started) {
-          log.warn('Logcat not started. Log capture before exit:');
-          for (let line of this.logs) {
-            log.warn(`    ${line}`);
-          }
+          log.warn('Logcat not started. Continuing');
           resolve();
         }
       });
--- a/lib/logcat.js
+++ b/lib/logcat.js
@@ -13,13 +13,29 @@ class Logcat {
   }
 
   startCapture () {
-    return new Promise(async (resolve, reject) => {
-      log.debug("Starting logcat capture");
+    let started = false;
+    return new Promise(async (_resolve, _reject) => {
+      const resolve = function (...args) {
+        started = true;
+        _resolve(...args);
+      };
+      const reject = function (...args) {
+        started = true;
+        _reject(...args);
+      };
+
+      log.debug('Starting logcat capture');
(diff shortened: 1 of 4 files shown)


lib/logcat.js
```diff
@@ -13,13 +13,26 @@ class Logcat {
   }
 
   startCapture () {
-    return new Promise(async (resolve, reject) => {
-      log.debug("Starting logcat capture");
+    let started = false;
+    return new Promise(async (_resolve, _reject) => {
+      const resolve = function (...args) {
```
mykola-mokhnach: should it be a function or a lambda expression?

imurchie: It doesn't make use of anything that the fat arrow would provide. So there is no need.

lib/logcat.js
```diff
@@ -13,13 +13,26 @@ class Logcat {
...
+
+      log.debug('Starting logcat capture');
       this.proc = new SubProcess(this.adb.path, this.adb.defaultArgs.concat(['logcat', '-v', 'threadtime']));
-      await this.proc.start(0);
       this.proc.on('exit', (code, signal) => {
         log.error(`Logcat terminated with code ${code}, signal ${signal}`);
         this.proc = null;
+        if (!started) {
+          log.warn('Logcat not started. Continuing');
+          resolve();
```
mykola-mokhnach: I assume one can call resolve/reject multiple times without consequences. See https://stackoverflow.com/questions/20328073/is-it-safe-to-resolve-a-promise-multiple-times

imurchie: Yes.  Once it is resolved to a value, it is resolved to that value.

mykola-mokhnach: which means we don't really need the _started_ flag

jlipps: without the `started` flag, how will we know to only display the warning if this exit handler was entered during logcat startup?

mykola-mokhnach: These wrapped resolve and reject functions don't look quite elegant IMHO %)

jlipps: Yeah, unfortunately the eventemitter api doesn't return promises so we have to use reject/resolve inside of it :-(

imurchie: They could be removed, but then the setting of `started` would have to be all over the place.

PR #247, opened 2017-08-03
Update eslint and deps

imurchie:
Update linting.

### DIFF
--- a/test/unit/helper-specs.js
+++ b/test/unit/helper-specs.js
@@ -4,6 +4,7 @@ import { withMocks } from 'appium-test-support';
 import { fs } from 'appium-support';
 import path from 'path';
 import _ from 'lodash';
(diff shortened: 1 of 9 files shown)


test/unit/helper-specs.js
```diff
@@ -16,7 +17,7 @@ describe('helpers', () => {
         .returns(directories);
       mocks.fs.expects('lstat')
         .exactly(5)
-        .returns(Promise.resolve({isDirectory: () => {return true;}}));
+        .returns(B.resolve({isDirectory: () => {return true;}}));
```
mykola-mokhnach: is it mandatory to return a promise here?

imurchie: No, this was just how the folks writing this module did it. The async/await mechanics can handle a normal value.

PR #249, opened 2017-08-09
Clean up logging of avd output

imurchie:
Split multi-line avd output and log on individual lines.
At the moment we get logging like (see [this log](https://gist.github.com/vmaxim/79c608c0fc15ad9928fe7bd293f89648#file-appium_logs_2017-08-09-L83-L86)):
```
2017-08-09 05:58:57:293 - [ADB] [AVD OUTPUT] Hax is enabled
Hax ram_size 0x40000000
HAX is working and emulator runs in fast virt mode.

```
This PR will change it to:
```
2017-08-09 05:58:57:293 - [ADB] [AVD OUTPUT] Hax is enabled
2017-08-09 05:58:57:293 - [ADB] [AVD OUTPUT] Hax ram_size 0x40000000
2017-08-09 05:58:57:293 - [ADB] [AVD OUTPUT] HAX is working and emulator runs in fast virt mode.
```

### DIFF
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -438,7 +438,9 @@ systemCallMethods.launchAVD = async function (avdName, avdArgs, language, countr
   let proc = new SubProcess(emulatorBinaryPath, launchArgs);
   await proc.start(0);
   proc.on('output', (stdout, stderr) => {
-    log.info(`[AVD OUTPUT] ${stdout || stderr}`);
+    for (let line of (stdout || stderr || '').split('\n').filter(Boolean)) {
+      log.info(`[AVD OUTPUT] ${line}`);
+    }
   });
   proc.on('exit', (code, signal) => {
     if (code !== 0) {


lib/tools/system-calls.js
```diff
@@ -438,7 +438,9 @@ systemCallMethods.launchAVD = async function (avdName, avdArgs, language, countr
   let proc = new SubProcess(emulatorBinaryPath, launchArgs);
   await proc.start(0);
   proc.on('output', (stdout, stderr) => {
-    log.info(`[AVD OUTPUT] ${stdout || stderr}`);
+    for (let line of (stdout || stderr || '').split('\n').filter(Boolean)) {
```
mykola-mokhnach: `(stdout || stderr || ''`: does this mean that the situation when stdout and stderr are both defined is not possible?

imurchie: I don't know if there would ever be a case when `stdout` and `stderr` are both empty. But this adds no complexity, and no decrease in performance (since `||` short-circuits), while making sure that the call to `split` will not fail.

mykola-mokhnach: I mean in case they both are not empty then this code will print stdout only

imurchie: Oh, sorry! Yes, `stdout` and `stderr` will never both have something at the same time. They will create two different events.

imurchie: See https://github.com/appium/node-teen_process/blob/master/lib/teen_process.js#L212-L222

jlipps: another option would have been to use `lines-stderr` and `lines-stdout` events, which do this for you already. there is however no `lines-output` event which admittedly would be most appropriate for this case.

PR #251, opened 2017-08-21
Stop pretending the error does not exist

imurchie:
When we get an error, return an error.

See https://github.com/appium/appium/issues/8995


lib/tools/system-calls.js
```diff
@@ -225,7 +225,7 @@ systemCallMethods.adbExec = async function (cmd, opts = {}) {
         await this.getDevicesWithRetry();
       }
 
-      if (e.stdout) {
+      if (e.code === 0 && e.stdout) {
```
mykola-mokhnach: I assume _code_ will never be equal to zero if we are already in this block

mykola-mokhnach: I'd delete the entire condition then. Also my god feeling says it may have something to do with **linkerWarningRe** constant

PR #252, opened 2017-08-22
add forward list method

babyrjw:
Appium will shutdown chromedriver process before starting a webivew session. But It didn't remove all adb forwarded ports. We need a method to find out all forwarded ports. So I add a forwardList method into appium-adb.


lib/tools/adb-commands.js
```diff
@@ -532,6 +532,11 @@ methods.processExists = async function (processName) {
   }
 };
 
+methods.forwardList = async function () {
```
imurchie: I would prefer something like `getForwardList`. Semantically, `forwardPort` actually sets the port forwarding, so one would expect `forwardList` to set the list forwarding, which it does not.

lib/tools/adb-commands.js
```diff
@@ -532,6 +532,15 @@ methods.processExists = async function (processName) {
   }
 };
 
+/**
+ * Get TCP port forwarding with adb on the device under test.
+ * @return {string} The output of the corresponding adb command. A string contains multiple forwarding lines with line break.
+ */
+methods.getForwardList = async function () {
+  log.debug(`List forwarding ports`);
+  return await this.adbExec(['forward', '--list']);
```
imurchie: Now that I've thought about this a bit, I think this should return an array of strings (so, the output split by newlines). This makes more sense to me.

babyrjw: Yes. That's better.

imurchie:
Published in `appium-adb@2.28.0`.

PR #254, opened 2017-08-22
Improve existing docstrings and add missing ones.

mykola-mokhnach:
Used these tutorials:
https://stackoverflow.com/questions/28763257/jsdoc-return-object-structure
http://usejsdoc.org/tags-type.html

### DIFF
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -47,8 +47,8 @@ methods.initZipAlign = async function () {
 /**
  * Retrieve the API level of the device under test.
  *
- * @return (string) The API level as integer number, for example 21 for
- * Android Lollipop. The result of this method is cached, so all the further
+ * @return {string} The API level as integer number, for example 21 for
+ *                  Android Lollipop. The result of this method is cached, so all the further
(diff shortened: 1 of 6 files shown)


mykola-mokhnach:
There are still several undocumented methods there, but I really cannot fully understand how they work %)

lib/tools/adb-commands.js
```diff
@@ -47,8 +47,8 @@ methods.initZipAlign = async function () {
 /**
  * Retrieve the API level of the device under test.
  *
- * @return (string) The API level as integer number, for example 21 for
- * Android Lollipop. The result of this method is cached, so all the further
+ * @return {string} The API level as integer number, for example 21 for
```
jlipps: maybe `integer string` rather than `integer number`? (since it's a string)

jlipps: ...should we return a number instead?

mykola-mokhnach: Maybe... Maybe it even makes sense to cache this value.
But due to some historical reasons unknown to me this method is implemented this way.

PR #257, opened 2017-09-12
killProcessByPID may work well with param -9

andrewleo:
KILL PID may not close the uiautomator process on some devices like LENOVO K50-T5, add -9 may work well

### DIFF
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -1002,7 +1002,7 @@ methods.killProcessesByName = async function (name) {
  */
 methods.killProcessByPID = async function (pid) {
   log.debug(`Attempting to kill process ${pid}`);
-  return await this.shell(['kill', pid]);
+  return await this.shell(['kill', '-9', pid]);
 };
 
 /**


imurchie:
Is there an error that happens with the current technique? It would be nice to avoid using the `-9` signal generally, and only fall back to it.

andrewleo:
I found the KILL PID can not kill uiautomator process on LENOVO K50-T5, but with -9 it may work and also on other 200 devices.
So if it is recommended to avoid -9, maybe we can check whether the uiautomator process is killed, if not, use -9 may be another good way?

imurchie:
That is why I am wondering what happens on those devices. Is there some sort of error? Or does it just go on but not kill the process? 

Is there an issue this is in regard to, where there might be logs?

andrewleo:
Yeah, appium will go on when KILL PID do not take effect, then Uiautomator2Server can not work.

I have try KILL PID through adb command, and there is no error returned, while the uiautomator process can not be killed, so I wonder it maybe a bug of LENOVO K50-T5

No issues reported, cause only one phone happended. :)

andrewleo:
So maybe Appium can check whether uiautomator process is killed or not, after run KILL PID, that would be more stable and robust, especially on these phones with Bugs. :)

lib/tools/adb-commands.js
```diff
@@ -1002,7 +1002,7 @@ methods.killProcessesByName = async function (name) {
  */
 methods.killProcessByPID = async function (pid) {
   log.debug(`Attempting to kill process ${pid}`);
-  return await this.shell(['kill', pid]);
+  return await this.shell(['kill', '-9', pid]);
```
mykola-mokhnach: it might be useful to try "normal" kill first. Then make sure the process is killed by invoking kill with -0 signal and only after that send SIGKILL if the process still exists after the timeout. _waitForCondition_ helper method from asyncbox module could be also handy.

andrewleo: yeah, that will be better to try KILL PID first, and check whether uiautomator process is killed or not. 
maybe you can change the code and pull a new request? Because I am not familiar to Nodejs

mykola-mokhnach: done in https://github.com/appium/appium-adb/pull/262

andrewleo: Much thanks!

PR #258, opened 2017-09-12
Check if the package is installed before to uninstall it

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/9196

### DIFF
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -253,12 +253,12 @@ apkUtilsMethods.uninstallApk = async function (pkg) {
     log.errorAndThrow(`Unable to uninstall APK. Original error: ${e.message}`);
   }
   stdout = stdout.trim();
-  // stdout may contain warnings meaning success is not on the first line.
+  log.debug(`ADB command output: ${stdout}`);
   if (stdout.indexOf("Success") !== -1) {
     log.info(`${pkg} was successfully uninstalled`);
     return true;
   }
-  log.info(`${pkg} was not uninstalled. Command output: ${stdout}`);
+  log.info(`${pkg} was not uninstalled`);
   return false;
 };
 
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -241,21 +241,25 @@ apkUtilsMethods.waitForNotActivity = async function (pkg, act, waitMs = 20000) {
  */
 apkUtilsMethods.uninstallApk = async function (pkg) {
   log.debug(`Uninstalling ${pkg}`);
-  try {
-    await this.forceStop(pkg);
-    let stdout = await this.adbExec(['uninstall', pkg], {timeout: 20000});
+  if (this.isAppInstalled(pkg)) {
+    let stdout;
+    try {
+      await this.forceStop(pkg);
+      stdout = await this.adbExec(['uninstall', pkg], {timeout: 20000});
+    } catch (e) {
+      log.errorAndThrow(`Unable to uninstall APK. Original error: ${e.message}`);
+    }
     stdout = stdout.trim();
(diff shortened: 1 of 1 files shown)


lib/tools/apk-utils.js
```diff
@@ -241,21 +241,25 @@ apkUtilsMethods.waitForNotActivity = async function (pkg, act, waitMs = 20000) {
...
-    } else {
-      log.info("App was not uninstalled, maybe it wasn't on device?");
-      return false;
-    }
+    stdout = await this.adbExec(['uninstall', pkg], {timeout: 20000});
   } catch (e) {
     log.errorAndThrow(`Unable to uninstall APK. Original error: ${e.message}`);
   }
+  stdout = stdout.trim();
+  // stdout may contain warnings meaning success is not on the first line.
```
imurchie: Should these warnings be outputted?

lib/tools/apk-utils.js
```diff
@@ -241,21 +241,25 @@ apkUtilsMethods.waitForNotActivity = async function (pkg, act, waitMs = 20000) {
  */
 apkUtilsMethods.uninstallApk = async function (pkg) {
   log.debug(`Uninstalling ${pkg}`);
+  if (!this.isAppInstalled(pkg)) {
+    log.info(`${pkg} was not uninstalled, because it was not present on the device`);
+    return false;
```
imurchie: Is this `false` taken as an error condition anywhere?

lib/tools/apk-utils.js
mykola-mokhnach: I do output it later if there is no Success substring there.

lib/tools/apk-utils.js
mykola-mokhnach: does this matter? These changes don't affect the previous logic of this mathod

lib/tools/apk-utils.js
imurchie: But what if there are "warnings", whatever they are, but also success?

mykola-mokhnach: It's not a problem at all. Added one more log.debug call

lib/tools/apk-utils.js
imurchie: Does it not? If the app is not installed the uninstall process doesn't get attempted. That seems like a change in the logic of this method.

I'm just wanting to make sure that when `false` is returned the calling code does not consider it to have been a problem.

mykola-mokhnach: The result is only used in the removeApp API call inside general.js:

```
$ cd appium-android-driver/lib/
$ grep -R 'uninstallApk' *
android-helpers.js:      await adb.uninstallApk(pkg);
commands/general.js:  return this.adb.uninstallApk(appPackage);
commands/general.js:    await this.adb.uninstallApk(this.opts.appPackage);
driver.js:      await this.adb.uninstallApk(this.opts.appPackage);
driver.js:        await this.adb.uninstallApk(this.opts.appPackage);
$ cd appium-uiautomator2-driver/lib/
$ grep -R 'uninstallApk' *
driver.js:      await this.adb.uninstallApk(this.opts.appPackage);
driver.js:          await this.adb.uninstallApk(this.opts.appPackage);
uiautomator2.js:      await this.adb.uninstallApk(apkPackage);
uiautomator2.js:      await this.adb.uninstallApk(testApkPackage);
```

mykola-mokhnach:
```
$ cd appium-selendroid-driver/lib/
$ grep -R 'uninstallApk' *
driver.js:      await this.adb.uninstallApk(this.opts.appPackage);
selendroid.js:      await this.adb.uninstallApk(this.modServerPkg);
```

imurchie: Excellent. Thanks!

PR #260, opened 2017-09-15
Add missing await

mykola-mokhnach:
This should finally fix https://github.com/appium/appium/issues/9196


imurchie:
Fixed the tests and published in `appium-adb@2.28.2`

PR #261, opened 2017-09-18
Match receiver names to Appium Settings v 2.3.0+

mykola-mokhnach:
Update receiver names, otherwise switching WiFi/Data state won't work properly


imurchie:
Published in `appium-adb@2.28.5`.

PR #262, opened 2017-09-18
Force process killing if the initial kill command fails

mykola-mokhnach:
Replaces https://github.com/appium/appium-adb/pull/257


imurchie:
Published in `appium-adb@2.28.4`.

PR #263, opened 2017-09-18
Fix appium/appium#9228

vmaxim:


lib/helpers.js
```diff
@@ -111,10 +111,11 @@ function getJavaHome () {
 }
 
 /*
- * Checks mShowingLockscreen in dumpsys output to determine if lock screen is showing
+ * Checks mShowingLockscreen or mDreamingLockscreen in dumpsys output to determine
```
mykola-mokhnach: Could you please update the docstring to include _param_ and _return_  descriptions?

lib/helpers.js
```diff
@@ -110,11 +110,15 @@ function getJavaHome () {
...
+ /**
+  * Checks mShowingLockscreen or mDreamingLockscreen in dumpsys output to determine
+  * if lock screen is showing
+  *
+  * @param {string} dumpsys - The output of dumpsys window command.
+  * @return {boolean} True if lock screen is showing.
+  */
 function isShowingLockscreen (dumpsys) {
-  let m = /mShowingLockscreen=\w+/gi.exec(dumpsys);
+  let m = /(mShowingLockscreen=\w+|mDreamingLockscreen=\w+)/gi.exec(dumpsys);
```
imurchie: Is it never the case that `mShowingLockScreen` is present (but false) and `mDreamingLockscreen` is true?

test/unit/helper-specs.js
```diff
@@ -55,6 +55,25 @@ describe('helpers', () => {
     });
   }));
 
+  describe('isShowingLockscreen', () => {
+    it('should return mShowingLockscreen value', async () => {
+      let dumpsys = 'mShowingLockscreen=true mShowingDream=false mDreamingLockscreen=false mTopIsFullscreen=false';
+      (await isShowingLockscreen(dumpsys)).should.be.true;
+      dumpsys = 'mShowingLockscreen=false mShowingDream=false mDreamingLockscreen=true mTopIsFullscreen=false';
```
imurchie: These should be separate tests.

lib/helpers.js
vmaxim: https://android.googlesource.com/platform/frameworks/base/+/android-6.0.0_r1/services/core/java/com/android/server/policy/PhoneWindowManager.java#4507

imurchie: Does that answer my question?

My concern is that this function only checks the first match, which, if there are both `mShowingLockScreen` and `mDreamingLockscreen` entries, will only be whatever comes first. I'd rather see a check for both entries, and what their values are.

vmaxim: I'm not sure that we should skip the value of `mShowingLockScreen` if it's `false` - this can broke current behavior.
I should double check.

vmaxim: {mShowingLockScreen: false, mDreamingLockscreen: true} indicates that device starts showing dream (screensaver) from lock screen. So it seems reasonable to catch this case.

imurchie:
Published in `appium-adb@2.28.6`.

PR #264, opened 2017-09-20
Add option to full reinstall on upgrade

imurchie:
Add flag to make `installOrUpgrade` force uninstall and then install. This is needed for some cases of the settings app. In that case, we don't care about whether we save any state or data.

### DIFF
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -328,15 +328,14 @@ apkUtilsMethods.installOrUpgrade = async function (apk, pkg = null, fullReinstal
     return;
   }
 
-  if (!await this.isAppInstalled(pkg)) {
-    await this.install(apk, false, timeout);
-    return;
-  }
-
   if (fullReinstall) {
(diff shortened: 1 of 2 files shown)


lib/tools/apk-utils.js
```diff
@@ -325,32 +327,43 @@ apkUtilsMethods.installOrUpgrade = async function (apk, pkg = null, timeout = 60
     log.warn(`Cannot read the package name of ${apk}. Assuming correct app version is already installed`);
     return;
   }
+
   if (!await this.isAppInstalled(pkg)) {
     await this.install(apk, false, timeout);
     return;
   }
+
+  if (fullReinstall) {
```
jlipps: could shorten this a bit by putting the check and uninstall ahead of the `isAppInstalled` check, then relying on the install to happen in that block. NBD either way.

mykola-mokhnach: yep, uninstallApk already contains the verification of whether the destination app is installed or not

PR #265, opened 2017-10-09
Add methods to set and get animation state

vmaxim:


lib/tools/adb-commands.js
```diff
@@ -727,6 +734,32 @@ methods.setWifiAndData = async function ({wifi, data}, isEmulator = false) {
 };
 
 /**
+ * Change the state of animation on the device under test.
```
mykola-mokhnach: Can you please add a line that describes why this might be needed to change that state?

lib/tools/adb-commands.js
```diff
@@ -727,6 +734,32 @@ methods.setWifiAndData = async function ({wifi, data}, isEmulator = false) {
...
+ * Check the state of animation on the device under test.
+ *
+ * @return {boolean} True if at least one of animation scale settings
+ *                   is not equal to zero.
+ */
+methods.isAnimationOn = async function () {
+  let animator_duration_scale = await this.getSetting('global', 'animator_duration_scale');
+  let transition_animation_scale = await this.getSetting('global', 'transition_animation_scale');
+  let window_animation_scale = await this.getSetting('global', 'window_animation_scale');
+  return (parseFloat(animator_duration_scale, 10) !== 0 ||
```
mykola-mokhnach: you cannot compare float values this way. Either cast it to integer or use [Number.ESPILON](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/EPSILON) constant

lib/tools/adb-commands.js
```diff
@@ -757,7 +790,7 @@ methods.setGeoLocation = async function (location, isEmulator = false) {
     this.adbExec(['emu', 'geo', 'fix', longitude.replace('.', ','), latitude.replace('.', ',')]);
   } else {
     return await this.shell(['am', 'startservice', '-e', 'longitude', longitude,
-                             '-e', 'latitude', latitude, `${SETTINGS_HELPER_ID}/.LocationService`]);
+                             '-e', 'latitude', latitude, `${SETTINGS_HELPER_ID}/${LOCATION_SERVICE}`]);
```
mykola-mokhnach: here we might also write the whole `${SETTINGS_HELPER_ID}/${LOCATION_SERVICE}` string into a constant

lib/tools/adb-commands.js
```diff
@@ -735,28 +735,36 @@ methods.setWifiAndData = async function ({wifi, data}, isEmulator = false) {
...
+ *                   is not equal to "0.0".
  */
 methods.isAnimationOn = async function () {
   let animator_duration_scale = await this.getSetting('global', 'animator_duration_scale');
   let transition_animation_scale = await this.getSetting('global', 'transition_animation_scale');
   let window_animation_scale = await this.getSetting('global', 'window_animation_scale');
-  return (parseFloat(animator_duration_scale, 10) !== 0 ||
-          parseFloat(transition_animation_scale, 10) !== 0 ||
-          parseFloat(window_animation_scale, 10) !== 0);
+  return (animator_duration_scale !== "0.0" ||
```
mykola-mokhnach: 👍

mykola-mokhnach: can also be `!_.some([animator_duration_scale, transition_animation_scale, window_animation_scale], (x) => x === '0.0')`

PR #266, opened 2017-10-13
Only end coverage if there is actually a proc running

imurchie:


lib/tools/adb-commands.js
```diff
@@ -1071,7 +1071,7 @@ methods.broadcast = async function (intent) {
  * Kill Android instruments if they are currently running.
  */
 methods.endAndroidCoverage = async function () {
-  if (this.instrumentProc) {
+  if (this.instrumentProc && this.instrumentProc.isRunning) {
```
mykola-mokhnach: is it a property?

imurchie: It's a getter.

imurchie: https://github.com/appium/node-teen_process/blob/master/lib/teen_process.js#L131-L134

imurchie:
The only time I've seen this actually happening is when the process is never started (when we have a failure on session start and never actually get to the start coverage stage).

PR #267, opened 2017-10-16
Rely on failure to uninstall app

imurchie:
Rather than uninstall _a priori_, rely on uninstalling if there is an error in installing.

This means we don't need the `forceUninstall` parameter. This is technically a breaking change, but there is only one place where this method is used, so we can just fix that when the time comes.


imurchie:
Yes.

jlipps:
This looks OK, but didn't we add this recently to cover a specific use case? Are we sure we don't need it?

PR #268, opened 2017-10-25
Escape ampersand in URIs

sa-spag:
... so we can use URIs with multiple query parameters.


sa-spag:
It seems like something went wrong with the second Travis CI build job. Can someone restart it please?

PR #269, opened 2017-10-30
Make getApiLevel to return number instead of string

mykola-mokhnach:
I see that in many places people wrongly consider the returned value of getApiLevel is an integer number and perform string-to-number comparison, which never works properly (though throws no warnings/errors with JS). This PR unifies the stuff and makes the method to always return a number. The change should not be a problem for client libs, since `parseInt(x, 10)` will work properly if x is already an integer.

### DIFF
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -53,14 +53,18 @@ methods.initZipAlign = async function () {
 /**
  * Retrieve the API level of the device under test.
  *
- * @return {string} The API level as integer number, for example 21 for
+ * @return {number} The API level as integer number, for example 21 for
  *                  Android Lollipop. The result of this method is cached, so all the further
  * calls return the same value as the first one.
  */
 methods.getApiLevel = async function () {
-  if (!this._apiLevel) {
+  if (!_.isNumber(this._apiLevel)) {
     try {
-      this._apiLevel = await this.getDeviceProperty('ro.build.version.sdk');
+      const strOutput = await this.getDeviceProperty('ro.build.version.sdk');
+      this._apiLevel = parseInt(strOutput.trim(), 10);
+      if (isNaN(this._apiLevel)) {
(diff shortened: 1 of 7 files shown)


lib/tools/adb-commands.js
```diff
@@ -53,14 +53,18 @@ methods.initZipAlign = async function () {
...
  * calls return the same value as the first one.
  */
 methods.getApiLevel = async function () {
-  if (!this._apiLevel) {
+  if (!_.isNumber(this._apiLevel)) {
     try {
-      this._apiLevel = await this.getDeviceProperty('ro.build.version.sdk');
+      const strOutput = await this.getDeviceProperty('ro.build.version.sdk');
+      this._apiLevel = parseInt(strOutput.trim(), 10);
+      if (isNaN(this._apiLevel)) {
```
imurchie: If `this._apiLevel` is `NaN`, if this method is called again `NaN` will be returned, since `_.isNumber(NaN) === true`. Either after this check `this._apiLevel` should be set to `null`/`undefined`, or the first check should include a `isNaN` check, or, since API level should always be an integer, the check could be `_.isInteger(this._apiLevel)`.

mykola-mokhnach: _.isNumber(), which returns true for NotANumber... Something is wrong with this world

jlipps: NotANumber is definitely a number, the "not" kind!

mykola-mokhnach:
Should we also increase the major component version number?

imurchie:
It should not be necessary. Users of this function either treat it as a number or parseInt it into a number. The former will be handled better now that it's a number. The latter will just return the same number, so no problem.

You should, however, go through the Android modules and bump the `appium-adb` version and remove the parseInt parts.

imurchie:
Published in `appium-adb@4.1.0`.

PR #270, opened 2017-11-11
check if environment variable for default adb port is set

igorilin13:
Default adb port is not always 5037. If ANDROID_ADB_SERVER_PORT is set, appium will fail to find devices unless the correct adbPort is explicitly passed via capabilities.

### DIFF
--- a/lib/adb.js
+++ b/lib/adb.js
@@ -4,7 +4,7 @@ import path from 'path';
 import methods from './tools/index.js';
 import { rootDir} from './helpers';
 
-const DEFAULT_ADB_PORT = 5037;
+const DEFAULT_ADB_PORT = process.env.ANDROID_ADB_SERVER_PORT ? process.env.ANDROID_ADB_SERVER_PORT : 5037;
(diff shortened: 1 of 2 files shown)


igorilin13:
Because if the environment variable is not set, adb starts on port 5037. However, if it is set, it starts on the specified port.

Here's an example:

Let's say ANDROID_ADB_SERVER_PORT = 7500.

$ adb connect 100.101.102.103
*daemon not running. starting it now at tcp:7500 *
*daemon started successfully *
connected to 100.101.102.103

$ adb devices
List of devices attached
100.101.102.103		device

Appium executes...
$ adb -P 5037 devices
*daemon not running. starting it now at tcp:5037 *
*daemon started successfully *
List of devices attached

0 device(s) connected.

So in order to make it work, you need to set the correct port via capabilities.
I think this is an unexpected behavior. Any wrapper around adb should behave the same way adb does, that is, take the environment variable into account. This won't affect many users but it could save someone some time.

imurchie:
But that is precisely why the `adbPort` capability is there.

PR #271, opened 2017-11-19
add setting locale method via setting app

KazuCocoa:
We can set the locale via [the setting app feature](https://github.com/appium/io.appium.settings/pull/11) after the latest the app released without rebooting devices after changing the locale.

After merging this and update `helpers.ensureDeviceLocale` in `appium-android-driver`, we'll be able to set locale for Android N+ preventing https://github.com/appium/appium/issues/9619  issue.

-----

Update

- ~~I've added some `skip` to avoid travis failed which also have been happened on current master~~ => https://github.com/appium/appium-adb/pull/272
    - https://github.com/appium/appium-adb/pull/271#issuecomment-346260317
    - https://github.com/appium/appium-adb/pull/271/commits/c8d31a57944a61a41347cddffa1cea5773b95e16
    - I'll create an issue to fix it. But in this PR, I add `skip` to focus on current changes.

After merging this PR, I'll update:

- https://github.com/appium/appium-android-driver/blob/ed07b52eab38865da619877ba90eb068d758964f/lib/android-helpers.js#L91
    - remove existing duplicated logic and update test cases
    - update package.json to use the latest `appium-adb` and `io.appium.settings`

### DIFF
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -17,6 +17,8 @@ const DATA_CONNECTION_SETTING_RECEIVER = `${SETTINGS_HELPER_ID}/.receivers.DataC
 const DATA_CONNECTION_SETTING_ACTION = `${SETTINGS_HELPER_ID}.data_connection`;
 const ANIMATION_SETTING_RECEIVER = `${SETTINGS_HELPER_ID}/.receivers.AnimationSettingReceiver`;
 const ANIMATION_SETTING_ACTION = `${SETTINGS_HELPER_ID}.animation`;
+const LOCALE_SETTING_RECEIVER = `${SETTINGS_HELPER_ID}/.receivers.LocaleSettingReceiver`;
+const LOCALE_SETTING_ACTION = `${SETTINGS_HELPER_ID}.locale`;
 const LOCATION_SERVICE = `${SETTINGS_HELPER_ID}/.LocationService`;
 const MAX_SHELL_BUFFER_LENGTH = 1000;
 
@@ -768,6 +770,22 @@ methods.isAnimationOn = async function () {
   let window_animation_scale = await this.getSetting('global', 'window_animation_scale');
   return _.some([animator_duration_scale, transition_animation_scale, window_animation_scale],
                 (setting) => setting !== '0.0');
+};
+
+/**
+ * Change the locale on the device under test. Don't need to reboot the device after changing the locale.
+ * This method sets an arbitrary locale following:
+ *   https://developer.android.com/reference/java/util/Locale.html
+ *   https://developer.android.com/reference/java/util/Locale.html#Locale(java.lang.String,%20java.lang.String)
+ *
+ * @param {string} language - Language. e.g. en, ja
+ * @param {string} country - Country. e.g. US, JP
+ */
+methods.setDeviceSysLocaleViaSettingApp = async function (language, country) {
+  await this.shell(['am', 'broadcast', '-a', LOCALE_SETTING_ACTION,
+    '-n', LOCALE_SETTING_RECEIVER,
+    '--es', 'lang', language.toLowerCase(),
+    '--es', 'country', country.toUpperCase()]);
 };
 
 /**
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -479,7 +479,7 @@ apkUtilsMethods.getDeviceLocale = async function () {
  * @param {string} locale - Names of the device language and the country connected with `-`. e.g. en-US.
  */
 apkUtilsMethods.setDeviceLocale = async function (locale) {
-  let validateLocale = new RegExp(/[a-zA-Z0-9]+-[a-zA-Z0-9]+/);
+  const validateLocale = new RegExp(/[a-zA-Z]+-[a-zA-Z0-9]+/);
   if (!validateLocale.test(locale)) {
     log.warn(`setDeviceLocale requires the following format: en-US or ja-JP`);
     return;
@@ -550,7 +550,7 @@ apkUtilsMethods.setDeviceLanguageCountry = async function (language, country) {
   }
 
   if (wasSettingChanged) {
-    log.info("To apply locale setting via setting persist.sys.locale command, reboot the target device is required.");
+    log.info("Rebooting the device in order to apply new locale via 'setting persist.sys.locale' command.");
     await this.reboot();
   }
 };
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -474,6 +474,7 @@ apkUtilsMethods.getDeviceLocale = async function () {
 
 /**
  * Set the locale name of the device under test.
+ * Compatibility for appium-android-driver 1.32.4-
  *
  * @param {string} locale - The name of the new device country.
  */
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -476,9 +476,15 @@ apkUtilsMethods.getDeviceLocale = async function () {
  * Set the locale name of the device under test and the format of the locale is en-US, for example.
  * This method call setDeviceLanguageCountry, so, please use setDeviceLanguageCountry as possible.
  *
- * @param {string} locale - The name of device country connected with `-`. e.g. en-US.
+ * @param {string} locale - Names of the device language and the country connected with `-`. e.g. en-US.
  */
 apkUtilsMethods.setDeviceLocale = async function (locale) {
+  let validateLocale = new RegExp(/[a-zA-Z0-9]+-[a-zA-Z0-9]+/);
+  if (!validateLocale.test(locale)) {
+    log.warn(`setDeviceLocale requires the following format: en-US or ja-JP`);
+    return;
+  }
+
   let split_locale = locale.split("-");
   await this.setDeviceLanguageCountry(split_locale[0], split_locale[1]);
 };
@@ -493,8 +499,8 @@ apkUtilsMethods.setDeviceLanguageCountry = async function (language, country) {
   let hasLanguage = language && _.isString(language);
   let hasCountry = country && _.isString(country);
   if (!hasLanguage && !hasCountry) {
-    log.debug(`setDeviceLanguageCountry requires language or country.`);
-    log.debug(`language: '${language}' and country: '${country}'`);
+    log.warn(`setDeviceLanguageCountry requires language or country.`);
+    log.warn(`Got language: '${language}' and country: '${country}'`);
     return;
   }
   let wasSettingChanged = false;
@@ -531,8 +537,8 @@ apkUtilsMethods.setDeviceLanguageCountry = async function (language, country) {
       }
     } else { // API >= 24
       if (!hasCountry || !hasLanguage) {
-        log.debug(`setDeviceLanguageCountry requires both language and country to be set for API 24+`);
-        log.debug(`Got language: '${language}' and country: '${country}'`);
+        log.warn(`setDeviceLanguageCountry requires both language and country to be set for API 24+`);
+        log.warn(`Got language: '${language}' and country: '${country}'`);
         return;
       }
 
@@ -544,6 +550,7 @@ apkUtilsMethods.setDeviceLanguageCountry = async function (language, country) {
   }
 
   if (wasSettingChanged) {
+    log.info("To apply locale setting via setting persist.sys.locale command, reboot the target device is required.");
     await this.reboot();
   }
 };
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -478,8 +478,76 @@ apkUtilsMethods.getDeviceLocale = async function () {
  * @param {string} locale - The name of the new device country.
  */
 apkUtilsMethods.setDeviceLocale = async function (locale) {
-  // this method is only used in API >= 23
-  await this.setDeviceSysLocale(locale);
+  let split_locale = locale.split("-");
+  await this.setDeviceLanguageCountry(split_locale[0], split_locale[1]);
+};
+
+/**
+ * Set the locale name of the device under test.
+ *
+ * @param {string} language - Language. e.g. en, ja
+ * @param {string} country - Country. e.g. US, JP
+ */
+apkUtilsMethods.setDeviceLanguageCountry = async function (language, country) {
+  let haveLanguage = language && typeof language === "string";
+  let haveCountry = country && typeof country === "string";
+  if (!haveLanguage && !haveCountry) {
+    return;
+  }
+  let changed = false;
+  let apiLevel = await this.getApiLevel();
+
+  if (apiLevel < 23) {
+    let curLanguage = await this.getDeviceLanguage();
+    let curCountry = await this.getDeviceCountry();
+    if (haveLanguage && language !== curLanguage) {
+      await this.setDeviceLanguage(language);
+      changed = true;
+    }
+    if (haveCountry && country !== curCountry) {
+      await this.setDeviceCountry(country);
+      changed = true;
+    }
+  } else if (apiLevel < 24) { // API == 23
+    let curLocale = await this.getDeviceLocale();
+    let locale;
+    if (!haveCountry) {
+      locale = language.toLowerCase();
+    } else if (!haveLanguage) {
+      locale = country;
+    } else {
+      locale = `${language.toLowerCase()}-${country.toUpperCase()}`;
+    }
+
+    log.debug(`Current locale: '${curLocale}'; requested locale: '${locale}'`);
+    if (locale !== curLocale) {
+      await this.setDeviceSysLocale(locale);
+      changed = true;
+    }
+  } else { // API >= 24
+    let curLocale = await this.getDeviceLocale();
+    let localeLang;
+    let localeCountry;
+    if (!haveCountry) {
+      localeLang = language.toLowerCase();
+      localeCountry = language.toUpperCase();
+    } else if (!haveLanguage) {
+      localeLang = country.toLowerCase();
+      localeCountry = country.toUpperCase();
+    } else {
+      localeLang = language.toLowerCase();
+      localeCountry = country.toUpperCase();
+    }
+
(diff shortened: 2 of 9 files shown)


test/functional/adb-commands-e2e-specs.js
```diff
@@ -151,6 +151,17 @@ describe('adb commands', function () {
     await adb.setAnimationState(true);
     (await adb.isAnimationOn()).should.be.true;
   });
+  it('should be able to set device locale via setting app @skip-ci', async () => {
+    // Operation not allowed: java.lang.SecurityException: Package io.appium.settings has not requested permission android.permission.CHANGE_CONFIGURATION
+    // is shown if the setting apk is not updated.
+    await adb.grantPermission('io.appium.settings', 'android.permission.CHANGE_CONFIGURATION');
+
+    await adb.setDeviceSysLocaleViaSettingApp('fr', 'fr');
+    (await adb.getDeviceSysLocale().should.equal('fr-FR'));
```
mykola-mokhnach: I assume only `await adb.getDeviceSysLocale()` should be enclosed into parentheses. Same thing below.

lib/tools/adb-commands.js
```diff
@@ -770,6 +772,22 @@ methods.isAnimationOn = async function () {
...
+/**
+ * Change the locale on the device under test. Don't need to reboot the device after changing the locale.
+ * This method sets an arbitrary locale following:
+ *   https://developer.android.com/reference/java/util/Locale.html
+ *   https://developer.android.com/reference/java/util/Locale.html#Locale(java.lang.String,%20java.lang.String)
+ *
+ * @param {string} language - Language. e.g. en, ja
+ * @param {string} country - Country. e.g. US, JP
+ */
+methods.setDeviceSysLocaleViaSettingApp = async function (language, country) {
```
mykola-mokhnach: I was thinking about some approach, which would allow us to isolate country/language selection logic on appium-adb module level instead of changing it on driver level. For example we call something like `adb.setDeviceLocale` and the internal logic automatically decides which method selection to apply based on API level, so we don't have to tune that for each driver.
How do you think @imurchie ?

KazuCocoa: 👍 I'm also thinking so (the driver just call `adb.setDeviceLocale` for example) when I read codebase and added this method.

mykola-mokhnach: you can implement this approach, so then we could increase major module version and update client drivers when it is published

lib/tools/apk-utils.js
```diff
@@ -474,12 +474,81 @@ apkUtilsMethods.getDeviceLocale = async function () {
...
+    if (locale !== curLocale) {
+      await this.setDeviceSysLocale(locale);
+      changed = true;
+    }
+  } else { // API >= 24
+    let curLocale = await this.getDeviceLocale();
+    let localeLang;
+    let localeCountry;
+    if (!haveCountry) {
+      localeLang = language.toLowerCase();
```
KazuCocoa: Can cover `fr` case, for example.

lib/tools/apk-utils.js
```diff
@@ -474,12 +474,81 @@ apkUtilsMethods.getDeviceLocale = async function () {
...
+apkUtilsMethods.setDeviceLanguageCountry = async function (language, country) {
+  let haveLanguage = language && typeof language === "string";
+  let haveCountry = country && typeof country === "string";
+  if (!haveLanguage && !haveCountry) {
+    return;
+  }
+  let changed = false;
+  let apiLevel = await this.getApiLevel();
+
+  if (apiLevel < 23) {
```
KazuCocoa: `< 24` cases are almost moved from https://github.com/appium/appium-android-driver/blob/ed07b52eab38865da619877ba90eb068d758964f/lib/android-helpers.js#L91

KazuCocoa:
Same error on current master. (Travis)
https://travis-ci.org/appium/appium-adb/jobs/295270516

I'll take a look more.

lib/tools/apk-utils.js
```diff
@@ -474,12 +474,81 @@ apkUtilsMethods.getDeviceLocale = async function () {
...
+  await this.setDeviceLanguageCountry(split_locale[0], split_locale[1]);
+};
+
+/**
+ * Set the locale name of the device under test.
+ *
+ * @param {string} language - Language. e.g. en, ja
+ * @param {string} country - Country. e.g. US, JP
+ */
+apkUtilsMethods.setDeviceLanguageCountry = async function (language, country) {
```
KazuCocoa: - in `lib/tools/adb-commands.js`, focus on calling adb commands
- in `lib/tools/apk-utils.js`, handle adb commands

lib/tools/apk-utils.js
```diff
@@ -474,12 +474,81 @@ apkUtilsMethods.getDeviceLocale = async function () {
 
 /**
  * Set the locale name of the device under test.
+ * Compatibility for appium-android-driver 1.32.4-
```
mykola-mokhnach: please do not add explicit driver versions, since not only android driver depends on this module

lib/tools/apk-utils.js
```diff
@@ -474,12 +474,81 @@ apkUtilsMethods.getDeviceLocale = async function () {
...
  *
  * @param {string} locale - The name of the new device country.
```
mykola-mokhnach: please add an example value(s)

lib/tools/apk-utils.js
```diff
@@ -474,12 +474,81 @@ apkUtilsMethods.getDeviceLocale = async function () {
...
+};
+
+/**
+ * Set the locale name of the device under test.
+ *
+ * @param {string} language - Language. e.g. en, ja
+ * @param {string} country - Country. e.g. US, JP
+ */
+apkUtilsMethods.setDeviceLanguageCountry = async function (language, country) {
+  let haveLanguage = language && typeof language === "string";
```
mykola-mokhnach: use `_.isString`

lib/tools/apk-utils.js
```diff
@@ -474,12 +474,81 @@ apkUtilsMethods.getDeviceLocale = async function () {
...
+  let haveCountry = country && typeof country === "string";
+  if (!haveLanguage && !haveCountry) {
+    return;
```
mykola-mokhnach: please add some logging here

lib/tools/apk-utils.js
mykola-mokhnach: -> hasLanguage

lib/tools/apk-utils.js
```diff
@@ -474,12 +474,81 @@ apkUtilsMethods.getDeviceLocale = async function () {
...
+
+/**
+ * Set the locale name of the device under test.
+ *
+ * @param {string} language - Language. e.g. en, ja
+ * @param {string} country - Country. e.g. US, JP
+ */
+apkUtilsMethods.setDeviceLanguageCountry = async function (language, country) {
+  let haveLanguage = language && typeof language === "string";
+  let haveCountry = country && typeof country === "string";
```
mykola-mokhnach: -> hasCountry

lib/tools/apk-utils.js
```diff
@@ -474,12 +474,81 @@ apkUtilsMethods.getDeviceLocale = async function () {
...
+  if (!haveLanguage && !haveCountry) {
+    return;
+  }
+  let changed = false;
```
mykola-mokhnach: -> wasSettingChanged

lib/tools/apk-utils.js
```diff
@@ -474,12 +474,81 @@ apkUtilsMethods.getDeviceLocale = async function () {
...
+    let curCountry = await this.getDeviceCountry();
+    if (haveLanguage && language !== curLanguage) {
+      await this.setDeviceLanguage(language);
+      changed = true;
+    }
+    if (haveCountry && country !== curCountry) {
+      await this.setDeviceCountry(country);
+      changed = true;
+    }
+  } else if (apiLevel < 24) { // API == 23
```
mykola-mokhnach: why not to say `apiLevel === 23` ? ;)

lib/tools/apk-utils.js
```diff
@@ -474,12 +474,81 @@ apkUtilsMethods.getDeviceLocale = async function () {
...
+    let curLocale = await this.getDeviceLocale();
+    let locale;
+    if (!haveCountry) {
+      locale = language.toLowerCase();
+    } else if (!haveLanguage) {
+      locale = country;
```
mykola-mokhnach: country.toUppperCase?

lib/tools/apk-utils.js
```diff
@@ -474,12 +474,81 @@ apkUtilsMethods.getDeviceLocale = async function () {
...
+    log.debug(`Current locale: '${curLocale}'; requested locale: '${locale}'`);
+    if (locale !== curLocale) {
+      await this.setDeviceSysLocale(locale);
+      changed = true;
+    }
+  } else { // API >= 24
+    let curLocale = await this.getDeviceLocale();
+    let localeLang;
+    let localeCountry;
+    if (!haveCountry) {
```
mykola-mokhnach: Can we refactor the code to avoid duplicating this condition for api == 23/api > 23 cases?

test/unit/logcat-specs.js
```diff
@@ -12,7 +12,8 @@ describe('logcat', async () => {
   let adb = {path: 'dummyPath', defaultArgs: []};
   let logcat = new Logcat({adb, debug: false, debugTrace: false});
   describe('startCapture', withMocks({teen_process}, (mocks) => {
-    it('should correctly call subprocess and should resolve promise', async () => {
+    // Because of  ExpectationError: Unexpected call: resolve(...) on Travis
```
mykola-mokhnach: Shouldn't it be a separate PR?

KazuCocoa: ah, yes. I'll create another PR to skip them and rebase after this PR after merge it.

lib/tools/apk-utils.js
```diff
@@ -474,12 +474,76 @@ apkUtilsMethods.getDeviceLocale = async function () {
...
-  await this.setDeviceSysLocale(locale);
+  let split_locale = locale.split("-");
+  await this.setDeviceLanguageCountry(split_locale[0], split_locale[1]);
+};
+
+/**
+ * Set the locale name of the device under test.
+ *
+ * @param {string} language - Language. format: [a-zA-Z]{2,8}. e.g. en, ja : https://developer.android.com/reference/java/util/Locale.html
+ * @param {string} country - Country. format: [a-zA-Z]{2} | [0-9]{3}. e.g. US, JP : https://developer.android.com/reference/java/util/Locale.html
```
KazuCocoa: I've changed don't convert the language and the country to _upper case_ or _lower case_ in `setDeviceLanguageCountry` because they should accept `[a-zA-Z]` as Locale spec.

lib/tools/apk-utils.js
KazuCocoa: BTW, we can **remove** the `api == 23` case and use `setDeviceSysLocaleViaSettingApp` instead since I've used `setDeviceSysLocaleViaSettingApp` logic to change device locale for all devices in my company and the devices include api 23.

^ is just my thinking and what do you think @mykola-mokhnach ? (It's okay if you have no exact opinion, though)
But I also think it's better to keep current behaviour, using adb command, for the `api == 23/api > 23` since they work fine so far. 🤔

lib/tools/apk-utils.js
```diff
@@ -474,12 +474,76 @@ apkUtilsMethods.getDeviceLocale = async function () {
...
+
+    log.debug(`Current locale: '${curLocale}'; requested locale: '${locale}'`);
+    if (locale !== curLocale) {
+      await this.setDeviceSysLocale(locale);
+      wasSettingChanged = true;
+    }
+  } else { // API >= 24
+    let curLocale = await this.getDeviceLocale();
+    if (!hasCountry || !hasLanguage) {
+      log.debug(`setDeviceLanguageCountry requires language and country for API 24+`);
```
mykola-mokhnach: -> ... requires both language and country to be set for ...

lib/tools/apk-utils.js
```diff
@@ -474,12 +474,76 @@ apkUtilsMethods.getDeviceLocale = async function () {
...
+      log.debug(`language: '${language}' and country: '${country}'`);
```
mykola-mokhnach: ... Got language: ...

lib/tools/apk-utils.js
mykola-mokhnach: The logic is ok there, no need to remove anything. I just wanted to optimise some redundant code and the recent state of this code block makes me happy :)

KazuCocoa:
✅  and updated.
After merging https://github.com/appium/appium-adb/pull/272 and rebase to the latest master, I'll CI should green.

lib/tools/apk-utils.js
```diff
@@ -473,13 +473,79 @@ apkUtilsMethods.getDeviceLocale = async function () {
...
+      }
+
+      log.debug(`Current locale: '${curLocale}'; requested locale: '${locale}'`);
+      if (locale !== curLocale) {
+        await this.setDeviceSysLocale(locale);
+        wasSettingChanged = true;
+      }
+    } else { // API >= 24
+      if (!hasCountry || !hasLanguage) {
+        log.debug(`setDeviceLanguageCountry requires both language and country to be set for API 24+`);
```
mykola-mokhnach: I'd rather change the level to `warn` and add a mention about device locale is not going to be changed

lib/tools/apk-utils.js
```diff
@@ -473,13 +473,79 @@ apkUtilsMethods.getDeviceLocale = async function () {
...
+
+      log.debug(`Current locale: '${curLocale}'; requested locale: '${language}-${country}'`);
+      if (`${language}-${country}` !== curLocale) {
+        await this.setDeviceSysLocaleViaSettingApp(language, country);
+      }
+    }
+  }
+
+  if (wasSettingChanged) {
+    await this.reboot();
```
mykola-mokhnach: info log message here might be useful to explain why reboot is needed

lib/tools/apk-utils.js
```diff
@@ -473,13 +473,79 @@ apkUtilsMethods.getDeviceLocale = async function () {
...
+ * Set the locale name of the device under test and the format of the locale is en-US, for example.
+ * This method call setDeviceLanguageCountry, so, please use setDeviceLanguageCountry as possible.
  *
- * @param {string} locale - The name of the new device country.
+ * @param {string} locale - The name of device country connected with `-`. e.g. en-US.
  */
 apkUtilsMethods.setDeviceLocale = async function (locale) {
-  // this method is only used in API >= 23
-  await this.setDeviceSysLocale(locale);
+  let split_locale = locale.split("-");
```
mykola-mokhnach: can we add some input validation here?

mykola-mokhnach: at least that the string has a dash, so split_locale[0] won't raise an unexpected exception

lib/tools/apk-utils.js
```diff
@@ -473,13 +473,79 @@ apkUtilsMethods.getDeviceLocale = async function () {
 };
 
 /**
- * Set the locale name of the device under test.
+ * Set the locale name of the device under test and the format of the locale is en-US, for example.
+ * This method call setDeviceLanguageCountry, so, please use setDeviceLanguageCountry as possible.
  *
- * @param {string} locale - The name of the new device country.
+ * @param {string} locale - The name of device country connected with `-`. e.g. en-US.
```
mykola-mokhnach: -> ...Names of the device language and the country...

lib/tools/apk-utils.js
```diff
@@ -473,13 +473,79 @@ apkUtilsMethods.getDeviceLocale = async function () {
...
+ * Set the locale name of the device under test.
+ *
+ * @param {string} language - Language. format: [a-zA-Z]{2,8}. e.g. en, ja : https://developer.android.com/reference/java/util/Locale.html
+ * @param {string} country - Country. format: [a-zA-Z]{2} | [0-9]{3}. e.g. US, JP : https://developer.android.com/reference/java/util/Locale.html
+ */
+apkUtilsMethods.setDeviceLanguageCountry = async function (language, country) {
+  let hasLanguage = language && _.isString(language);
+  let hasCountry = country && _.isString(country);
+  if (!hasLanguage && !hasCountry) {
+    log.debug(`setDeviceLanguageCountry requires language or country.`);
```
mykola-mokhnach: ->log.warn

lib/tools/apk-utils.js
```diff
@@ -473,13 +473,79 @@ apkUtilsMethods.getDeviceLocale = async function () {
...
+    log.debug(`language: '${language}' and country: '${country}'`);
```
mykola-mokhnach: -> Got language...

lib/tools/apk-utils.js
```diff
@@ -476,9 +476,15 @@ apkUtilsMethods.getDeviceLocale = async function () {
  * Set the locale name of the device under test and the format of the locale is en-US, for example.
  * This method call setDeviceLanguageCountry, so, please use setDeviceLanguageCountry as possible.
  *
- * @param {string} locale - The name of device country connected with `-`. e.g. en-US.
+ * @param {string} locale - Names of the device language and the country connected with `-`. e.g. en-US.
  */
 apkUtilsMethods.setDeviceLocale = async function (locale) {
+  let validateLocale = new RegExp(/[a-zA-Z0-9]+-[a-zA-Z0-9]+/);
```
mykola-mokhnach: this can be module level constant. Also, are digits allowed there?

lib/tools/apk-utils.js
```diff
@@ -544,6 +550,7 @@ apkUtilsMethods.setDeviceLanguageCountry = async function (language, country) {
   }
 
   if (wasSettingChanged) {
+    log.info("To apply locale setting via setting persist.sys.locale command, reboot the target device is required.");
```
mykola-mokhnach: -> Rebooting the device in order to apply new locale via 'setting persist.sys.locale' command...

mykola-mokhnach:
@KazuCocoa We usually squash commits before merging them to master, so then the storyline looks nicer.

PR #273, opened 2017-11-25
update version

KazuCocoa:
To apply _add setting locale method via setting app #271_, this PR bump up the version number.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "appium-adb",
-  "version": "4.1.0",
+  "version": "4.2.0",
   "description": "Android Debug Bridge interface",
   "main": "./build/index.js",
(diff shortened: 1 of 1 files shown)


mykola-mokhnach:
Did we change any public interfaces there? if so then the major version number should be updated.

KazuCocoa:
I've added a new public interface, `setDeviceSysLocaleViaSettingApp`, and `setDeviceLocale` works as same as 4.1.0.

KazuCocoa:
ah, but `setDeviceLocale` has validation from now. So, it's safe to update major version, I felt.
I'll create new PR.

imurchie:
The version in `package.json` should not be manually changed. It is changed (and committed) as part of the publish process.

PR #275, opened 2017-11-26
major version up

KazuCocoa:
- update version #273
- add setting locale method via setting app #271

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "appium-adb",
-  "version": "4.1.0",
+  "version": "5.0.0",
   "description": "Android Debug Bridge interface",
   "main": "./build/index.js",
   "scripts": {


mykola-mokhnach:
I'm not sure about it, but I think module version update happens when you execute npm publish and does not require a separate commit. Correct me if I'm wrong @imurchie

imurchie:
Correct. Npm has a tool for changing the version and tagging correctly. In this case `npm version major` would do the work. The process is 
```
npm version major
git push --tags origin master
npm publish
```

KazuCocoa:
thanks!
I've tagged. https://github.com/appium/appium-adb/releases/tag/v5.0.0

Could you add me to the project if it's no problem? @imurchie 
https://www.npmjs.com/~kazucocoa

KazuCocoa:
I'll close this PR after publishing v5.0.0

imurchie:
I've added you, so you should be able to publish `appium-adb`. Be _very_ careful doing so. Build should pass, use the `npm` tools to manage versioning and tagging. Thanks!

PR #276, opened 2017-11-28
Stop skipping tests

imurchie:
Undo the skipping of the tests done in https://github.com/appium/appium-adb/pull/272 which never should have been merged.

In general, massive quarantining of tests is the **wrong way to go about things**. This means that all subsequent commits were not tested. In this case it was _one_ unit test that was the cause of all the rest failing, and relatively little work was needed to figure that out and get to a solution.

### DIFF
--- a/test/unit/helper-specs.js
+++ b/test/unit/helper-specs.js
@@ -6,6 +6,7 @@ import path from 'path';
 import _ from 'lodash';
 import B from 'bluebird';
 
+
 describe('helpers', () => {
   describe('getDirectories', withMocks({fs}, (mocks) => {
     it('should sort the directories', async () => {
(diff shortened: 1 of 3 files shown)


mykola-mokhnach:
mmb9qajq9swpi8xxy76a_400x400
... about the little work needed to get this fixed :)

test/unit/helper-specs.js
```diff
@@ -22,38 +23,58 @@ describe('helpers', () => {
...
         .exactly(2)
         .onCall(0).returns(false)
         .onCall(1).returns(true);
-      mocks.path.expects('resolve')
-        .exactly(3)
-        .onCall(0).returns('/path/to')
-        .onCall(1).returns('/path/to/apis1')
-        .onCall(2).returns('/path/to/apis2');
+      // manually mock `path`, to allow system to work
+      oldResolve = path.resolve;
```
mykola-mokhnach: shouldn't this be set in beforeEach?

imurchie: It could be, but this is the only test that needs it.

KazuCocoa:
thanks so much :thanks:

PR #278, opened 2017-11-28
Add command to get screen density from ADB.

fuzzybinary:


lib/tools/adb-commands.js
```diff
@@ -1338,6 +1338,21 @@ methods.getScreenSize = async function () {
   return null;
 };
 
+/**
+ * Get the current screen density in dpi
+ *
+ * @return {string} Device screen density as a string or _null_ if it
```
mykola-mokhnach: shouldn't it be a float number?

lib/tools/adb-commands.js
```diff
@@ -1338,6 +1338,21 @@ methods.getScreenSize = async function () {
...
+ *                  cannot be determined
+ */
+methods.getScreenDensity = async function () {
+  let stdout = await this.shell(['wm', 'density']);
+  let size = new RegExp(/Physical density: ([^\r?\n]+)*/g).exec(stdout);
+  if (size && size.length >= 2) {
+    return size[1].trim();
+  }
+  return null;
```
mykola-mokhnach: I'd rather throw an exception instead

lib/tools/adb-commands.js
fuzzybinary: I did think about that... maybe?

lib/tools/adb-commands.js
fuzzybinary: I'm matching previous functionality there from getScreenSize.  Since I intend on dropping this into the caps of the device for reading, I figure null makes sense?

fuzzybinary:
Okay, I've changed this to return a number instead of a string.

lib/tools/adb-commands.js
```diff
@@ -1338,6 +1338,22 @@ methods.getScreenSize = async function () {
...
+ * Get the current screen density in dpi
+ *
+ * @return {string} Device screen density as a number or _null_ if it
+ *                  cannot be determined
+ */
+methods.getScreenDensity = async function () {
+  let stdout = await this.shell(['wm', 'density']);
+  let density = new RegExp(/Physical density: ([^\r?\n]+)*/g).exec(stdout);
+  if (density && density.length >= 2) {
+    let densityNumber = Number(density[1].trim());
```
mykola-mokhnach: `= parseInt(..., 10)`

lib/tools/adb-commands.js
```diff
@@ -1338,6 +1338,22 @@ methods.getScreenSize = async function () {
...
+    return Number.isNaN(densityNumber) ? null : densityNumber;
```
mykola-mokhnach: just `isNaN`

lib/tools/adb-commands.js
```diff
@@ -1338,6 +1338,22 @@ methods.getScreenSize = async function () {
   return null;
 };
 
+/**
+ * Get the current screen density in dpi
+ *
+ * @return {string} Device screen density as a number or _null_ if it
```
mykola-mokhnach: please update the docstring :`{?number}`

test/unit/adb-commands-specs.js
```diff
@@ -792,6 +793,13 @@ describe('adb commands', () => {
       await adb.getScreenSize();
       mocks.adb.verify();
     });
+    it('should get device screen density', async () => {
+      mocks.adb.expects("shell")
+          .once().withExactArgs(['wm', 'density'])
+          .returns(screenDensity);
+      await adb.getScreenDensity();
```
mykola-mokhnach: Plase add the verification of the returned result. Also it would be good to have a test, which verifies `null` output case

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "appium-adb",
-  "version": "3.0.0",
+  "version": "5.0.0",
```
mykola-mokhnach: Please revert this

mykola-mokhnach: also I'm not sure why `test/fixtures/selendroid-test-app.apk` is present in the PR

fuzzybinary: Me either actually. It's in the repo, but likely it should probably be .gitignored?

imurchie:
Published in `appium-adb@5.1.0`.

PR #279, opened 2017-11-30
Add ensure current locale

KazuCocoa:
ref: https://github.com/appium/appium-android-driver/commit/7094439c7ba19a533f7e3fcd446d35f2e4fb9009#diff-97eeab6ebf462ba97a11ee0221ddaf13R102

I've implemented check current locale method in `appium-adb`.
After merging this method, I'll call `ensureCurrentLocale` from `appium-android-driver`'s `ensureDeviceLocale` method.

### DIFF
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -490,10 +490,41 @@ apkUtilsMethods.setDeviceLocale = async function (locale) {
 };
 
 /**
+ * Make sure current device locale is expected or not.
+ *
+ * @param {string} language - Language. The language field is case insensitive, but Locale always canonicalizes to lower case.
+ * @param {string} country - Country. The language field is case insensitive, but Locale always canonicalizes to lower case.
+ *
+ * @return {bool} If current locale is language and country as arguments, return true.
+ */
+apkUtilsMethods.ensureCurrentLocale = async function (language, country) {
+  if (!_.isString(language) && !_.isString(country)) {
+    log.warn('validateLocale requires language or country');
+    return;
+  }
+
+  if (await this.getApiLevel() < 23) {
+    const curLanguage = await this.getDeviceLanguage();
+    const curCountry = await this.getDeviceCountry();
+    if (language.toLowerCase() === curLanguage.toLowerCase() && country.toLowerCase() === curCountry.toLowerCase()) {
+      return true;
+    }
+  } else {
+    const curLocale = await this.getDeviceLocale();
+    if (`${language}-${country}`.toLowerCase() === curLocale.toLowerCase())  {
+      return true;
+    }
+  }
+  return false;
+};
+
+/**
  * Set the locale name of the device under test.
  *
- * @param {string} language - Language. format: [a-zA-Z]{2,8}. e.g. en, ja : https://developer.android.com/reference/java/util/Locale.html
- * @param {string} country - Country. format: [a-zA-Z]{2} | [0-9]{3}. e.g. US, JP : https://developer.android.com/reference/java/util/Locale.html
+ * @param {string} language - Language. The language field is case insensitive, but Locale always canonicalizes to lower case.
+ *                            format: [a-zA-Z]{2,8}. e.g. en, ja : https://developer.android.com/reference/java/util/Locale.html
+ * @param {string} country - Country. The country (region) field is case insensitive, but Locale always canonicalizes to upper case.
+ *                            format: [a-zA-Z]{2} | [0-9]{3}. e.g. US, JP : https://developer.android.com/reference/java/util/Locale.html
  */
 apkUtilsMethods.setDeviceLanguageCountry = async function (language, country) {
   let hasLanguage = language && _.isString(language);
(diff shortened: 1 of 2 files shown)


lib/tools/apk-utils.js
```diff
@@ -489,11 +489,42 @@ apkUtilsMethods.setDeviceLocale = async function (locale) {
...
+ * Make sure current device locale is expected or not.
+ *
+ * @param {string} language - Language. The language field is case insensitive, but Locale always canonicalizes to lower case.
+ * @param {string} country - Country. The language field is case insensitive, but Locale always canonicalizes to lower case.
+ *
+ * @return {boolean} If current locale is language and country as arguments, return true.
+ */
+apkUtilsMethods.ensureCurrentLocale = async function (language, country) {
+  if (!_.isString(language) && !_.isString(country)) {
+    log.warn('validateLocale requires language or country');
```
imurchie: `validateLocale`?

lib/tools/apk-utils.js
```diff
@@ -489,11 +489,42 @@ apkUtilsMethods.setDeviceLocale = async function (locale) {
...
+    return false;
+  }
+
+  if (await this.getApiLevel() < 23) {
+    const curLanguage = await this.getDeviceLanguage();
```
imurchie: Is it guaranteed that these methods (`getDeviceLanguage`, `getDeviceCountry`, `getDeviceLocale`) will always return a string (there is no chance of null or undefined)? If there is _any_ chance then it should be guarded against.

lib/tools/apk-utils.js
```diff
@@ -509,12 +540,12 @@ apkUtilsMethods.setDeviceLanguageCountry = async function (language, country) {
   if (apiLevel < 23) {
     let curLanguage = await this.getDeviceLanguage();
     let curCountry = await this.getDeviceCountry();
-    if (hasLanguage && language !== curLanguage) {
-      await this.setDeviceLanguage(language);
+    if (hasLanguage && language.toLowerCase() !== curLanguage.toLowerCase()) {
```
imurchie: With all these `toLowerCase` calls, if it doesn't matter in the setting (as it doesn't seem to from the implementation here), why not just do it once for everything and have the variables hold the lower case versions? Since strings are passed by value it won't affect the calling code's version of `language` and `country`.
```js
language = language.toLowerCase();
let curLanguage = (await this.getDeviceLanguage()).toLowerCase();
// ...
```

lib/tools/apk-utils.js
KazuCocoa: Yes, it returns `""` or `"en"`(some variations). So, I haven't implemented guard.

lib/tools/apk-utils.js
KazuCocoa: `curLanguage` and `curLanguage` can `.toLowerCase()`, but:

- `language` and `country` can be undefined in `apiLevel < 23` and `apiLevel === 23` case.
- `language.toLowerCase()` and `country.toUpperCase()` is called once in `else`.
- `locale.toLowerCase() !== curLocale.toLowerCase()` is called only comparition to make the comparition in case insentive following Android's `Locale`.
    - Is there any method to compare two string in case insentive in JavaScript? I haven't been able to find such method yet... 

Thus, I call `.toLowerCase()` or `.toUpperCase()` when they are needed.

lib/tools/apk-utils.js
```diff
@@ -489,11 +489,42 @@ apkUtilsMethods.setDeviceLocale = async function (locale) {
...
+apkUtilsMethods.ensureCurrentLocale = async function (language, country) {
+  if (!_.isString(language) && !_.isString(country)) {
+    log.warn('ensureCurrentLocale requires language or country');
+    return false;
+  }
+
+  if (await this.getApiLevel() < 23) {
+    const curLanguage = (await this.getDeviceLanguage()).toLowerCase();
+    const curCountry = (await this.getDeviceCountry()).toLowerCase();
+    if (language.toLowerCase() === curLanguage && country.toLowerCase() === curCountry) {
```
imurchie: All of these `toLowerCase` calls can be brought into one, too (in both of these functions)
```js
language = language.toLowerCase();
country = country.toLowerCase();
// ...
```

lib/tools/apk-utils.js
```diff
@@ -538,13 +538,13 @@ apkUtilsMethods.setDeviceLanguageCountry = async function (language, country) {
...
-    let curCountry = await this.getDeviceCountry();
-    if (hasLanguage && language.toLowerCase() !== curLanguage.toLowerCase()) {
+    let curLanguage = (await this.getDeviceLanguage()).toLowerCase();
+    let curCountry = (await this.getDeviceCountry()).toUpperCase();
+    if (hasLanguage && language.toLowerCase() !== curLanguage) {
       await this.setDeviceLanguage(language.toLowerCase());
       wasSettingChanged = true;
     }
-    if (hasCountry && country.toUpperCase() !== curCountry.toUpperCase()) {
+    if (hasCountry && country.toUpperCase() !== curCountry) {
```
KazuCocoa: 📝 I can reduce the time to call touppercase

lib/tools/apk-utils.js
imurchie: Since the empty case is handled by `hasLanguage`, then `language = (language || '').toLowerCase()` would do it.

lib/tools/apk-utils.js
KazuCocoa: ✅

lib/tools/apk-utils.js
KazuCocoa: ✅

PR #280, opened 2017-12-02
Optimize logcat output collection

mykola-mokhnach:
Currently logcat output is pushed into two arrays, which causes extra memory consumption. The fix only pushes the data into single buffer and remembers its recent index. Also, the maximum size of the buffer is limited to 10k.

See https://github.com/appium/appium/issues/9727 for more details.

### DIFF
--- a/lib/logcat.js
+++ b/lib/logcat.js
@@ -4,14 +4,17 @@ import B from 'bluebird';
 
 
 const log = logger.getLogger('Logcat');
+const MAX_BUFFER_SIZE = 10000;
+
 
 class Logcat {
   constructor (opts = {}) {
     this.adb = opts.adb;
     this.debug = opts.debug;
     this.debugTrace = opts.debugTrace;
+    this.maxBufferSize = opts.maxBufferSize || MAX_BUFFER_SIZE;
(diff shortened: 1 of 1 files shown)


lib/logcat.js
```diff
@@ -4,14 +4,17 @@ import B from 'bluebird';
...
 const log = logger.getLogger('Logcat');
+const MAX_BUFFER_SIZE = 10000;
+
 
 class Logcat {
   constructor (opts = {}) {
     this.adb = opts.adb;
     this.debug = opts.debug;
     this.debugTrace = opts.debugTrace;
+    this.maxBufferSize = opts.maxBufferSize || MAX_BUFFER_SIZE;
```
imurchie: Is 10000 a hard limit? Should people be allowed to enter a `opts.maxBufferSize` that is higher?

mykola-mokhnach: it's the default value. This default has proven to be ok with iOS. 

> Should people be allowed to enter a opts.maxBufferSize that is higher?
Why not. Currently this will always be assigned to the default value, but then we might change the the default value in client drivers if needed.

PR #281, opened 2017-12-06
Add clearDeviceLogsOnStart option to run adb logcat -c before startin…

imurchie:
…g log capture

Beginning work to be able to clear logs on startup, if user wishes.

### DIFF
--- a/lib/logcat.js
+++ b/lib/logcat.js
@@ -1,15 +1,16 @@
-import { SubProcess } from 'teen_process';
+import { SubProcess, exec } from 'teen_process';
 import { logger } from 'appium-support';
 import B from 'bluebird';
 
 
 const log = logger.getLogger('Logcat');
 const MAX_BUFFER_SIZE = 10000;
-
+const LOGCAT_PROC_STARTUP_TIMEOUT = 10000;
 
 class Logcat {
   constructor (opts = {}) {
     this.adb = opts.adb;
+    this.clearLogs = opts.clearDeviceLogsOnStart || false;
     this.debug = opts.debug;
     this.debugTrace = opts.debugTrace;
     this.maxBufferSize = opts.maxBufferSize || MAX_BUFFER_SIZE;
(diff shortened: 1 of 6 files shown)


lib/logcat.js
```diff
@@ -106,6 +114,17 @@ class Logcat {
   getAllLogs () {
     return this.logs;
   }
+
+  async clear () {
+    log.debug('Clearing logcat logs from device');
+    try {
+      const args = this.adb.defaultArgs.concat(['logcat', '-c']);
```
mykola-mokhnach: how about `args = [...this.adb.defaultArgs, 'logcat', '-c']` ?

lib/logcat.js
```diff
@@ -106,6 +114,17 @@ class Logcat {
   getAllLogs () {
     return this.logs;
   }
+
+  async clear () {
```
mykola-mokhnach: does it make a sense to have a dedicated method in appium-adb for such purpose?

lib/logcat.js
```diff
@@ -56,6 +61,8 @@ class Logcat {
         }
       });
       await this.proc.start(0);
+      // resolve after a timeout, even if no output was recorded
+      setTimeout(resolve, LOGCAT_PROC_STARTUP_TIMEOUT);
```
mykola-mokhnach: is it ok to assume the promise is resolved after the timeout? I'd rather say it should be rejected

lib/logcat.js
imurchie: This is just in the `Logcat` class, not the `appium-adb` exposed class.

lib/logcat.js
imurchie: That's not generally how we've done it in this module. Even if it is preferable (not sure about that, though) it would still be better to keep consistent.

lib/logcat.js
mykola-mokhnach: I meant whether it makes sense to create such external endpoint and expose it to the public instead of having it as a class method?

lib/logcat.js
imurchie: It's not really an error condition. If there is an error the correct event will be fired, and we'll reject. This is just the case where the device does not produce any logs after we start listening for them. This has been happening in Travis in particular.

lib/logcat.js
mykola-mokhnach: I like spread operator, because it works for any iterable object, unlike concat, which is only the case for arrays.

PR #282, opened 2017-12-10
fix error case for api level 22-

KazuCocoa:
When I fixing https://github.com/appium/appium-uiautomator2-driver/pull/115/files, I found the following use case for API 22-.

```
await androidHelpers.ensureDeviceLocale(adb, null, initialLocale);
```

In this PR, I put control flows for the issue as same as `setDeviceLanguageCountry`.

After merging this PR, I'll publish new updating patch version.

### DIFF
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -510,22 +510,21 @@ apkUtilsMethods.ensureCurrentLocale = async function (language, country) {
   const lowCountry = (country || '').toLowerCase();
 
   if (await this.getApiLevel() < 23) {
-    if (hasCountry && hasLanguage) {
-      const curLanguage = (await this.getDeviceLanguage()).toLowerCase();
-      const curCountry = (await this.getDeviceCountry()).toLowerCase();
-      if (lowLanguage === curLanguage && lowCountry === curCountry) {
+    let curLanguage, curCountry;
+    if (hasLanguage) {
+      curLanguage = (await this.getDeviceLanguage()).toLowerCase();
+      if (!hasCountry && lowLanguage === curLanguage) {
         return true;
       }
-    } else if (hasLanguage) {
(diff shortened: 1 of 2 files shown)


lib/tools/apk-utils.js
```diff
@@ -498,19 +498,34 @@ apkUtilsMethods.setDeviceLocale = async function (locale) {
...
   }
 
-  const lowLanguage = language.toLowerCase();
-  const lowCountry = country.toLowerCase();
+  const lowLanguage = (language || '').toLowerCase();
+  const lowCountry = (country || '').toLowerCase();
 
   if (await this.getApiLevel() < 23) {
-    const curLanguage = (await this.getDeviceLanguage()).toLowerCase();
-    const curCountry = (await this.getDeviceCountry()).toLowerCase();
```
mykola-mokhnach: I assume it's still ok to get these values before the if condition

mykola-mokhnach: is it an expensive operation?

KazuCocoa: It's just to reduce the number of calling `adb` commands and I also considered as you addressed. The operation cost isn't expensive, I believe...

mykola-mokhnach: how about this?
```js
let curLanguage, curCountry;
if (hasLanguage) {
    curLanguage = (await this.getDeviceLanguage()).toLowerCase();
    if (!hasCountry && lowLanguage === curLanguage) {
        return true;
    }
}
if (hasCountry) {
    curCountry = (await this.getDeviceCountry()).toLowerCase();
    if (!hasLanguage && lowCountry === curCountry) {
        return true;
    }
}
if (lowLanguage === curLanguage && lowCountry === curCountry) {
    return true;
}
```

KazuCocoa: 👍 
looks good. thanks!

PR #283, opened 2017-12-15
Improve logging for packageAndLaunchActivityFromManifest

NozomiIto:
When apk file is corrupted, `await exec(this.binaries.aapt, args);` line throws exception and appium-adb just says like `Error: packageAndLaunchActivityFromManifest failed. Original error: Command '***/aapt dump badging ***.apk' exited with code 1` without any problem details.

Of course the users can easily know more detailed message just running `***/aapt dump badging ***.apk` by themselves, but most users report this problem without doing so. (Actually my client sent me this error message without  aapt error detail)
So I want to include the std error detail to the error message to reduce the trouble shooting communication cost.

### DIFF
--- a/lib/tools/android-manifest.js
+++ b/lib/tools/android-manifest.js
@@ -101,7 +101,11 @@ manifestMethods.packageAndLaunchActivityFromManifest = async function (localApk)
     log.debug(`badging act: ${apkActivity}`);
     return {apkPackage, apkActivity};
   } catch (e) {
-    log.errorAndThrow(`packageAndLaunchActivityFromManifest failed. Original error: ${e.message}`);
+    let stdErrInfo = "";
+    if (e.stderr) {
+      stdErrInfo = `\nStd err: ${e.stderr}`
+    }
+    log.errorAndThrow(`packageAndLaunchActivityFromManifest failed. Original error: ${e.message}` + stdErrInfo);
   }
 };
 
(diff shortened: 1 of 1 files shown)


NozomiIto:
Teen_process throws error object with stderr attribute if the process call failed ( https://github.com/appium/node-teen_process/blob/master/lib/teen_process.js#L96) , and I used this stderr attribute.

lib/tools/android-manifest.js
```diff
@@ -101,7 +101,11 @@ manifestMethods.packageAndLaunchActivityFromManifest = async function (localApk)
     log.debug(`badging act: ${apkActivity}`);
     return {apkPackage, apkActivity};
   } catch (e) {
-    log.errorAndThrow(`packageAndLaunchActivityFromManifest failed. Original error: ${e.message}`);
+    let stdErrInfo = "";
+    if (e.stderr) {
+      stdErrInfo = `\nStd err: ${e.stderr}`;
+    }
+    log.errorAndThrow(`packageAndLaunchActivityFromManifest failed. Original error: ${e.message}` + stdErrInfo);
```
mykola-mokhnach: just change it it to

```
log.errorAndThrow(`packageAndLaunchActivityFromManifest failed. Original error: ${e.message}` +
                                 (e.stderr ? `; StdErr: ${e.stderr}` : ''));
```
and all the code above is not needed anymore

NozomiIto: I have fixed.

PR #284, opened 2018-01-06
Sort build-tools subfolders by modification time rather than by name

mykola-mokhnach:
Addresses issues similar to https://github.com/appium/appium/issues/9926 where build-tools folder name might not match to the common naming pattern.

### DIFF
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -2,7 +2,7 @@ import path from 'path';
 import log from '../logger.js';
 import B from 'bluebird';
 import { system, fs } from 'appium-support';
-import { getDirectories, getSdkToolsVersion } from '../helpers';
+import { getSdkToolsVersion } from '../helpers';
 import { exec, SubProcess } from 'teen_process';
 import { sleep, retry, retryInterval } from 'asyncbox';
 import _ from 'lodash';
@@ -83,20 +83,17 @@ systemCallMethods.getBinaryFromSdkRoot = _.memoize(async function (binaryName) {
                     path.resolve(this.sdkRoot, "tools", binaryName),
                     path.resolve(this.sdkRoot, "tools", "bin", binaryName)];
   // get subpaths for currently installed build tool directories
-  let buildToolDirs = await getDirectories(path.resolve(this.sdkRoot, "build-tools"));
-  const buildDirsByMTime = {};
-  for (const buildToolDir of buildToolDirs) {
-    const fullPath = path.resolve(this.sdkRoot, "build-tools", buildToolDir);
-    const {mtimeMs} = await fs.stat(fullPath);
-    buildDirsByMTime[mtimeMs] = buildToolDir;
-  }
-  // the newest version goes first
-  const sortedMtimes = _.keys(buildDirsByMTime).sort().reverse();
-  buildToolDirs = sortedMtimes.map(x => buildDirsByMTime[x]);
-  log.info(`Found the following build-tools folders (the newest comes first): ` +
-           `${buildToolDirs.map(x => path.resolve(this.sdkRoot, 'build-tools', x))}`);
-  for (let versionDir of buildToolDirs) {
-    binaryLocs.push(path.resolve(this.sdkRoot, "build-tools", versionDir, binaryName));
+
+  let buildToolDirs = await fs.glob(path.resolve(this.sdkRoot, 'build-tools', '*') + path.sep, {absolute: true});
+  if (buildToolDirs.length) {
+    const pairs = await B.map(async (dir) => [(await fs.stat(dir)).mtimeMs, dir]);
+    buildToolDirs = pairs
+      .sort((a, b) => a[0] > b[0])
+      .map((pair) => pair[1]);
+    log.info(`Found the following 'build-tools' folders (the newest comes first): ${buildToolDirs}`);
+    buildToolDirs.map(x => binaryLocs.push(path.resolve(x, binaryName)));
+  } else {
+    log.info(`Found zero 'build-tools' folders`);
   }
   for (let loc of binaryLocs) {
     if (await fs.exists(loc)) {
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -86,7 +86,7 @@ systemCallMethods.getBinaryFromSdkRoot = _.memoize(async function (binaryName) {
 
   let buildToolDirs = await fs.glob(path.resolve(this.sdkRoot, 'build-tools', '*') + path.sep, {absolute: true});
   if (buildToolDirs.length) {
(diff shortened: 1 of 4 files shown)


lib/tools/system-calls.js
```diff
@@ -85,8 +84,17 @@ systemCallMethods.getBinaryFromSdkRoot = _.memoize(async function (binaryName) {
                     path.resolve(this.sdkRoot, "tools", "bin", binaryName)];
   // get subpaths for currently installed build tool directories
   let buildToolDirs = await getDirectories(path.resolve(this.sdkRoot, "build-tools"));
+  const buildDirsByMTime = {};
+  for (const buildToolDir of buildToolDirs) {
+    const fullPath = path.resolve(this.sdkRoot, "build-tools", buildToolDir);
+    const {mtime} = await fs.stat(fullPath);
```
imurchie: `mtime` is now a string with a human readable date.

lib/tools/system-calls.js
```diff
@@ -85,8 +84,17 @@ systemCallMethods.getBinaryFromSdkRoot = _.memoize(async function (binaryName) {
...
+    buildDirsByMTime[mtime] = buildToolDir;
+  }
   // the newest version goes first
-  buildToolDirs.sort(semver.rcompare);
+  const sortedMtimes = _.keys(buildDirsByMTime).sort().reverse();
```
imurchie: This is sorting by `mtime`, which will be sorted _alphabetically_, not according to time. For an original object like
```js
{ 'Thu Jan 04 2018 12:38:38 GMT-0500 (EST)': '23.0.1',
  'Thu Jan 04 2018 12:04:22 GMT-0500 (EST)': '25.0.0',
  'Tue Dec 19 2017 14:33:17 GMT-0500 (EST)': '26.0.2',
  'Tue Dec 19 2017 14:17:43 GMT-0500 (EST)': '27.0.2',
  'Fri Dec 29 2017 13:16:46 GMT-0500 (EST)': '27.0.3' }
```
the "sorted" array will be
```js
[ 'Tue Dec 19 2017 14:33:17 GMT-0500 (EST)',
  'Tue Dec 19 2017 14:17:43 GMT-0500 (EST)',
  'Thu Jan 04 2018 12:38:38 GMT-0500 (EST)',
  'Thu Jan 04 2018 12:04:22 GMT-0500 (EST)',
  'Fri Dec 29 2017 13:16:46 GMT-0500 (EST)' ]
```

imurchie: The solution is to pull `mtimeMs` off the stat object, so the sorting is on a number.

mykola-mokhnach:
@imurchie I thought one is not allowed to user await inside ,`map`

lib/tools/system-calls.js
mykola-mokhnach: The node doc says it should be a Date object, but if you says so...

imurchie: Ah. `sort` is coercing them into strings, then.

imurchie:
Sure. It's an async function, and the result is being awaited.

lib/tools/system-calls.js
mykola-mokhnach: Changed the algorithm to sort by `mtimeMs`. Ehre is the page where I was getting the info: https://nodejs.org/api/fs.html#fs_class_fs_stats

lib/tools/system-calls.js
```diff
@@ -85,8 +84,17 @@ systemCallMethods.getBinaryFromSdkRoot = _.memoize(async function (binaryName) {
...
+  buildToolDirs = sortedMtimes.map(x => buildDirsByMTime[x]);
+  log.info(`Found the following build-tools folders (the newest comes first): ` +
```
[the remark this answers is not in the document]
mykola-mokhnach: I assume it is not very big problem. In the worst case you can just `touch` the necessary folder to fix the stuff

mykola-mokhnach:
Sorry, perhaps I'm not so experienced in await/async stuff, but I see it like

```js
const buildToolDirs = (await fs.glob(path.resolve(this.sdkRoot, 'build-tools', '*')) // here we got the result of await  glob
    .map(async (dir) => [(await fs.stat(dir)).mtimeMs, dir])) // here we get a promise
    .sort((a, b) => a[0] < b[0]) // ...instead of a list
    .reverse()
    .map((arr) => arr[1]);
```

imurchie:
Almost. The first two lines are awaited together. So `fs.glob` returns a promise, which, when resolved, is passed to `map`, which returns a promise, which is awaited. This is then sorted, reversed, and remapped.

It can all be split apart, and if you want you could use something like [asyncmap](https://github.com/jlipps/asyncbox#filtermap). The main point is that we shouldn't reinvent the globbing wheel.

mykola-mokhnach:
Ok, changed the code to use `fs.glob`

lib/tools/system-calls.js
```diff
@@ -84,11 +83,17 @@ systemCallMethods.getBinaryFromSdkRoot = _.memoize(async function (binaryName) {
...
-    binaryLocs.push(path.resolve(this.sdkRoot, "build-tools", versionDir, binaryName));
+
+  let buildToolDirs = await fs.glob(path.resolve(this.sdkRoot, 'build-tools', '*') + path.sep, {absolute: true});
+  if (buildToolDirs.length) {
+    const pairs = await B.map(buildToolDirs, async (dir) => [(await fs.stat(dir)).mtimeMs, dir]);
+    buildToolDirs = pairs
+      .sort((a, b) => a[0] > b[0])
+      .map((pair) => pair[1]);
+    log.info(`Found the following 'build-tools' folders (the newest comes first): ${buildToolDirs}`);
+    buildToolDirs.map(x => binaryLocs.push(path.resolve(x, binaryName)));
```
imurchie: I don't think `map` is the right function here, since you are just using it to iterate over the array. `_.forEach` would make more sense.

imurchie: Also, given that the `binaryLocs` is being updated, the error message when nothing is found should be changed.

mykola-mokhnach: which error message you would like to see?

imurchie: Well, the error message at https://github.com/mykola-mokhnach/appium-adb/blob/df7d0ba95b8f00fb3a5f722f78ec26094958123b/lib/tools/system-calls.js#L105-L107 will say `Could not find ${binaryName} in ${binaryLocs}, or supported build-tools under ${this.sdkRoot} do you have the Android SDK installed at this location?`, but `binaryLocs` will include all the build-tools directories, so it is redundant.

mykola-mokhnach: it might be still useful to see which binary version was used later

imurchie: I agree. This is just about the error, which is just redundant now. I like the inclusion of the information about the one used in the success case.

PR #286, opened 2018-01-16
Add retry to ensureCurrentLocale

imurchie:
There are often failures in setting the language/locale after the fact, because adb gets disconnected (see https://github.com/appium/appium/issues/9987).

So add a retry if there is a failure, after restarting adb.

### DIFF
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -498,41 +498,50 @@ apkUtilsMethods.setDeviceLocale = async function (locale) {
  * @return {boolean} If current locale is language and country as arguments, return true.
  */
 apkUtilsMethods.ensureCurrentLocale = async function (language, country) {
-  let hasLanguage = _.isString(language);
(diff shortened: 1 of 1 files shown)


lib/tools/apk-utils.js
```diff
@@ -498,41 +498,50 @@ apkUtilsMethods.setDeviceLocale = async function (locale) {
...
-  if (!hasLanguage && !hasCountry) {
-    log.warn('ensureCurrentLocale requires language or country');
-    return false;
-  }
-
-  const lowLanguage = (language || '').toLowerCase();
-  const lowCountry = (country || '').toLowerCase();
+  return await retryInterval(5, 1000, async () => {
+    try {
+      const hasLanguage = _.isString(language);
```
mykola-mokhnach: variable setting and verification stuff might be moved out of `retryInterval` body.

imurchie: Good point. Thanks!

PR #288, opened 2018-01-19
fixes regex to parse pkg permissions

ilyanesterov:
This PR is to fix an issue related to `adb pm dump` command parsing. You can find more details about it here: https://github.com/appium/appium/issues/10018. 
I also added some tests as well to cover both cases. Please let me know if it is not needed.


ilyanesterov:
Oh gosh. It seems I screwed up this licensing thing with different accounts. I'll delete the PR and will create a new one.

PR #289, opened 2018-01-19
fixes regex to parse pkg permissions

ilyanesterov:
This PR is to fix an issue related to adb pm dump command parsing. You can find more details about the issue https://github.com/appium/appium/issues/10018#issuecomment-3588045460.
I also added some tests as well to cover both cases. Please let me know if it is not needed.


imurchie:
Published in `appium-adb@5.6.1`.

PR #173, opened 2016-10-03
Grant permissions on install

imurchie:
Use `-g` to grant all the permissions from the manifest, upon install.

Addresses https://github.com/appium/appium/issues/6928 and https://github.com/appium/appium/issues/6091

PR #290, opened 2018-01-22
Extend apk installation options

mykola-mokhnach:
Granting permissions with `pm grant` takes quite a lot of time, however `adb install` utility has -g command line  argument, which allows to do it much faster and also does not require us to parse the  package info. This PR changes public method interface, so the module should become a higher major version number after it is merged.

### DIFF
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -278,21 +278,53 @@ apkUtilsMethods.installFromDevicePath = async function (apkPathOnDevice, opts = 
 };
 
 /**
+ * @typedef {Object} InstallOptions
+ * @property {number} timeout [60000] - The count of milliseconds to wait until the
+ *                                      app is installed.
+ * @property {boolean} allowTestPackages [false] - Set to true in order to allow test
+ *                                                 packages installation.
+ * @property {boolean} useSdcard [false] - Set to true to install the app on sdcard
+ *                                         instead of the device memory.
+ * @property {boolean} grantPermissions [false] - Set to true in order to grant all
+ *                                                all application permissions automatically
+ *                                                after it is installed.
+ * @property {boolean} replace [true] - Set it to false if you don't want
+ *                                      the application to be upgraded.rewritten
+ *                                      if it is already installed.
+ */
+
+/**
  * Install the package from the local file system.
  *
  * @param {string} apk - The full path to the local package.
  * @param {boolean} repalce [true] - Whether to replace the package if it
  *                                   already installed. True by default.
- * @param {number} timeout [60000] - The number of milliseconds to wait until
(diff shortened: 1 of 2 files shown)


imurchie:
The permission granting stuff is there because last time we tried with `-g` it stopped working correctly. See, for instance, the discussion here: https://github.com/appium/appium-adb/pull/173

lib/tools/apk-utils.js
```diff
@@ -328,7 +385,7 @@ apkUtilsMethods.installOrUpgrade = async function (apk, pkg = null, timeout = 60
 
   if (!await this.isAppInstalled(pkg)) {
     log.debug(`App '${apk}' not installed. Installing`);
-    await this.install(apk, false, timeout);
+    await this.install(apk, Object.assign({replace: false}, options));
```
jlipps: I think this should be `Object.assign({}, options, {replace: false})` otherwise someone could pass in `replace: true` and it would overwrite what you have.

lib/tools/apk-utils.js
```diff
@@ -350,13 +407,13 @@ apkUtilsMethods.installOrUpgrade = async function (apk, pkg = null, timeout = 60
   log.debug(`The installed '${pkg}' package is older than '${apk}' (${pkgVersionCode} < ${apkVersionCode}). ` +
             `Executing upgrade`);
   try {
-    await this.install(apk, true, timeout);
+    await this.install(apk, Object.assign({replace: true}, options));
```
jlipps: same goes for this one, the thing you want to ensure is there must come last

lib/tools/apk-utils.js
```diff
@@ -350,13 +407,13 @@ apkUtilsMethods.installOrUpgrade = async function (apk, pkg = null, timeout = 60
...
   } catch (err) {
     log.warn(`Cannot upgrade '${pkg}' because of '${err.message}'. Trying full reinstall`);
     if (!await this.uninstallApk(pkg)) {
       log.errorAndThrow(`'${pkg}' package cannot be uninstalled`);
     }
-    await this.install(apk, false, timeout);
+    await this.install(apk, Object.assign({replace: false}, options));
```
jlipps: here too

mykola-mokhnach:
@imurchie This must be an adb issue. Lets hope it will be fixed by Google at some point of time. Anyway I still find this way of options setting much more flexible and extendable than the current one.

mykola-mokhnach:
@imurchie I think I know the root of the problem in https://github.com/appium/appium-adb/pull/173

Seems like adb tool has a bug, where it does not parse merged options properly, which means that

```
adb -r -g apkPath
```

works, but

```
adb -rg apkPath
```

doesn't.

PR #292, opened 2018-01-31
Log the output of `adb install` command

mykola-mokhnach:
Sometimes this bloody tool fails to do what it is supposed to do, and the return code is still zero and the error description is printed into stdout instead of stderr.


imurchie:
Published in `appium-adb@6.0.1`.

PR #293, opened 2018-02-02
Tune upgrade verification logic

mykola-mokhnach:
I see there are packages where versionCode is not maintained properly (for example UIA2) and only versionName is updated. This PR adds version names comparison to address such things


imurchie:
Published in `appium-adb@6.1.0`.

PR #296, opened 2018-02-08
Parse main activity from the manifest file using apkanalyzer tool

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/10120

### DIFF
--- a/lib/tools/android-manifest.js
+++ b/lib/tools/android-manifest.js
@@ -1,9 +1,12 @@
 import { exec } from 'teen_process';
 import log from '../logger.js';
-import { getAndroidPlatformAndPath, unzipFile, assertZipArchive } from '../helpers.js';
+import { getAndroidPlatformAndPath, unzipFile, assertZipArchive,
+         getApkanalyzerForOs } from '../helpers.js';
 import { system, fs } from 'appium-support';
 import _ from 'lodash';
 import path from 'path';
+import xmldom from 'xmldom';
+import xpath from 'xpath';
 
 const helperJarPath = path.resolve(__dirname, '..', '..', '..', 'jars');
 let manifestMethods = {};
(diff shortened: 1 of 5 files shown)


lib/tools/android-manifest.js
```diff
@@ -59,6 +62,89 @@ manifestMethods.processFromManifest = async function (localApk) {
...
+ * Extract package and main activity name from application manifest using
+ * the custom apk tools.
+ *
+ * @param {string} localApk - The full path to application package.
+ * @return {APKInfo} The parsed application info.
+ * @throws {Error} If there was an error while getting the data from the given
+ *                 application package.
+ */
+async function extractApkInfoWithApkTools (localApk) {
+  await this.initAapt();
```
imurchie: `this` isn't defined here.

lib/tools/android-manifest.js
```diff
@@ -68,42 +154,29 @@ manifestMethods.processFromManifest = async function (localApk) {
...
+  const apkInfoGetters = [
+    async () => await extractApkInfoWithApkanalyzer(localApk),
+    async () => await extractApkInfoWithApkTools(localApk),
+  ];
+
+  let savedError;
+  for (const infoGetter of apkInfoGetters) {
+    try {
+      const {apkPackage, apkActivity} = await infoGetter();
+      log.debug(`Package name: ${apkPackage}`);
```
imurchie: I think the values ought to be in single quotes (i.e., `Package name: '${apkPackage}'`), in case the tools return something weird that doesn't get printed right.

lib/tools/android-manifest.js
mykola-mokhnach: thanks, good catch. This, actually failed the tests %)

imurchie:
Published in `appium-adb@6.2.0`.

PR #297, opened 2018-02-08
Fix getting path to apkanalyzer

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/10120#issuecomment-364227397

### DIFF
--- a/lib/tools/android-manifest.js
+++ b/lib/tools/android-manifest.js
@@ -114,18 +114,18 @@ async function extractApkInfoWithApkTools (localApk, aaptPath, jarPath, tmpRoot)
  * apkanalyzer tool.
  *
  * @param {string} localApk - The full path to application package.
+ * @param {string} apkAnalyzerPath - The full path to apkanalyzer tool.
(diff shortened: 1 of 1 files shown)


lib/tools/android-manifest.js
```diff
@@ -159,7 +159,10 @@ async function extractApkInfoWithApkanalyzer (localApk) {
  */
 manifestMethods.packageAndLaunchActivityFromManifest = async function (localApk) {
   const apkInfoGetters = [
-    async () => await extractApkInfoWithApkanalyzer(localApk),
+    async () => {
+      const apkanalyzerPath = await getApkanalyzerForOs(this);
```
imurchie: Any reason to do this here rather than in the `extractApkInfoWithApkanalyzer` function?

mykola-mokhnach: we need `this` context

imurchie: Ah! Makes sense.

imurchie:
Published in `appium-adb@6.2.1`

PR #298, opened 2018-02-12
Properly handle avdArgs if these are of type array

mykola-mokhnach:
The current implementation only works if `avdArgs` is of string type


imurchie:
Published in `appium-adb@6.2.2`.

PR #299, opened 2018-02-18
Add bugreport feature

mykola-mokhnach:
This is the part of https://github.com/appium/appium/issues/4446 implementation

### DIFF
--- a/test/functional/adb-commands-e2e-specs.js
+++ b/test/functional/adb-commands-e2e-specs.js
@@ -238,20 +238,8 @@ describe('adb commands', function () {
   });
 
   describe('bugreport', function () {
-    it('should return the report as raw string by default', async function () {
+    it('should return the report as a raw string', async function () {
       (await adb.bugreport()).should.not.be.empty;
-    });
-
-    it('should return the report as base64-encoded .zip archive if needed', async function () {
-      if (process.env.TRAVIS) {
-        // This option does not work with older adb versions (<1.0.39)
-        return this.skip();
-      }
-      const base64 = await adb.bugreport(false);
-      Buffer.from(base64, 'base64')
-        .toString('ascii')
-        .startsWith('PK')
-        .should.be.true;
     });
   });
 
--- a/test/functional/adb-commands-e2e-specs.js
+++ b/test/functional/adb-commands-e2e-specs.js
@@ -236,4 +236,19 @@ describe('adb commands', function () {
       await adb.push(localFile, remoteFile).should.be.rejectedWith(/\/foo\/bar\/remote.txt/);
     });
   });
+
+  describe('bugreport', function () {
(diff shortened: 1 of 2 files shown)


test/functional/adb-commands-e2e-specs.js
```diff
@@ -236,4 +236,23 @@ describe('adb commands', function () {
...
     });
   });
+
+  describe('bugreport', function () {
+    it('should return the report as raw string by default', async function () {
+      (await adb.bugreport()).should.not.be.empty;
+    });
+
+    it('should return the report as base64-encoded .zip archive if needed', async function () {
+      if (await adb.getApiLevel() < 23) {
```
vmaxim: Maybe it's better to move this condition into `bugreport` method, if zipped report is unavailable for API<23?

mykola-mokhnach: looks like it depends on tools version rather than API version. I'll change the condition

vmaxim: I do not think that this is related to tools.

```adb bugreport myreport.zip```

Emu API 23 output:
>Failed to get bugreportz version: 'bugreportz -v' returned '/system/bin/sh: bugreportz: not found' (code 0).
If the device does not run Android 7.0 or above, try 'adb bugreport' instead.

Real Device API 21&22:
>Bugreport is in progress and it could take minutes to complete.
Please be patient and do not cancel or disconnect your device until it completes.
Unexpected string (/system/bin/sh: bugreportz: not found) returned by bugreportz, device probably does not support -z option

File: test/functional/adb-commands-e2e-specs.js [hunk 2]
```diff
@@ -243,8 +243,8 @@ describe('adb commands', function () {
     });
 
     it('should return the report as base64-encoded .zip archive if needed', async function () {
-      if (await adb.getApiLevel() < 23) {
-        // This option does not work with older APIs
+      if (process.env.TRAVIS) {
```
vmaxim: I do not think that this is related to tools.

```adb bugreport myreport.zip```

Emu API 23 output:
>Failed to get bugreportz version: 'bugreportz -v' returned '/system/bin/sh: bugreportz: not found' (code 0).
If the device does not run Android 7.0 or above, try 'adb bugreport' instead.

Real Device API 21&22:
>Bugreport is in progress and it could take minutes to complete.
Please be patient and do not cancel or disconnect your device until it completes.
Unexpected string (/system/bin/sh: bugreportz: not found) returned by bugreportz, device probably does not support -z option

**Comment by vmaxim** [comment]
Emu API 26:
```
d:\1>adb bugreport
/data/user_de/0/com.android.shell/files/bugreports/bugreport-...31-00.zip: 1 file pulled. 63.6 MB/s (1600684 bytes in 0.024s)
d:\1>
```

PR #304, opened 2018-02-28
Improve strings extraction algorithm

mykola-mokhnach:
The current algorithm only uses the stuff provided by `aapt` without custom jars. Also now it properly handles plural forms.

### DIFF
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -6,6 +6,7 @@ import _ from 'lodash';
 import { retryInterval } from 'asyncbox';
 import { fs, util } from 'appium-support';
 import semver from 'semver';
+import os from 'os';
 
 let apkUtilsMethods = {};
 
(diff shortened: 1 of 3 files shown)


mykola-mokhnach:
> we are switching implementations, what guarantee do we have that it's not a breaking change? should this constitute a major ver bump of appium-adb?

Nothing is guaranteed, however I compared the format of json file generated by the previous algorithm and the current one and they are equal, except of the part where plurals are parsed, but this does not make big diff, since json parsing algorithm inside UIA1 simply ignore such items.

> i'm not sure what the old helper apk's logic was. is it essentially the same as what's here?

I was trying to do generalise the output format as mentioned above, so UIA driver (bootstrap) can properly parse it if needed.

> how stable is the aapt output format? which versions of aapt have been tested?

I've checked the aapt output from the most recent Android SDK (v27), which is installed on my dev machine, and also some older one (v20). In both cases the output was equal for the same apk file.

PR #305, opened 2018-03-05
[README] Fix sample code

julien-c:


jlipps:
thank you, this is more correct. technically we can't use top-level await but hopefully people will know that :-)

julien-c:
@jlipps I think top-level await is coming :)

PR #311, opened 2018-03-23
Throw errors if result of `startUri` has errors

dpgraham:
* I need this for another PR because starters was unsuccessfully starting activities but was still returning zero as the error code so I need a way to check that 

(related PR: https://github.com/appium/appium-uiautomator2-driver/pull/151)

### DIFF
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -43,7 +43,10 @@ apkUtilsMethods.startUri = async function (uri, pkg) {
   try {
     let args = ["am", "start", "-W", "-a", "android.intent.action.VIEW", "-d",
                 uri.replace(/&/g, '\\&'), pkg];
-    return await this.shell(args);
+    const res = await this.shell(args);
+    if (res.toLowerCase().includes('unable to resolve intent')) {
+      throw new Error(res);
+    }
   } catch (e) {
     log.errorAndThrow(`Error attempting to start URI. Original error: ${e}`);
   }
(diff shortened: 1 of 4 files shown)


lib/tools/apk-utils.js
```diff
@@ -43,7 +43,7 @@ apkUtilsMethods.startUri = async function (uri, pkg) {
   try {
     let args = ["am", "start", "-W", "-a", "android.intent.action.VIEW", "-d",
                 uri.replace(/&/g, '\\&'), pkg];
-    await this.shell(args);
+    return await this.shell(args);
```
mykola-mokhnach: please  update the docstring as well

mykola-mokhnach:
Would it be possible to add failures verification into this method or you'd prefer to do this in the caller class?

dpgraham:
@mykola-mokhnach I updated `startUri` to catch the failures. I think it probably doesn't hurt to put in here instead of `appium-uiautomator2-driver`.

dpgraham:
@mykola-mokhnach  Done.

test/unit/apk-utils-specs.js
```diff
@@ -465,14 +465,32 @@ describe('Apk-utils', function () {
     });
   }));
   describe('startUri', withMocks({adb}, (mocks) => {
+    before(async function () {
```
mykola-mokhnach: do we need this empty func?

dpgraham: No, it's gone now.

dpgraham:
@mykola-mokhnach When the tests pass, am I good to merge then publish?

PR #312, opened 2018-03-23
Show raw adb command instead of array of args

dpgraham:
Instead of logging ADB exec commands like this: 

```Running '/Users/danielgraham/Library/Android/sdk/platform-tools/adb' with args: ["-P",5037,"-s","emulator-5554","shell","pm","clear","io.appium.settings"]```

...log them like this: 

```Running '/Users/danielgraham/Library/Android/sdk/platform-tools/adb -P 5037 -s emulator-5554 shell pm clear io.appium.settings'```

...so that we can copy the commands and run them from command line directly for troubleshooting/testing purposes.


dpgraham:
@jlipps Done.

PR #314, opened 2018-04-08
Improve warning error message

NozomiIto:
Now I'm investigating unexpected Java9 signing problem on my PC, but warning error message just says `exited with code 1` and its information is too little to know the error cause.
To know the cause of such problem more easily, I'd like to improve the logging.

### DIFF
--- a/lib/tools/apk-signing.js
+++ b/lib/tools/apk-signing.js
@@ -70,7 +70,7 @@ apkSigningMethods.signWithDefaultCert = async function (apk) {
     await this.executeApksigner(args);
   } catch (err) {
     log.warn(`Cannot use apksigner tool for signing. Defaulting to sign.jar. ` +
-             `Original error: ${err.message}`);
+             `Original error: ${err.message}` + (err.stderr ? `; StdErr: ${err.stderr}` : ''));
     const java = getJavaForOs();
     const signPath = path.resolve(this.helperJarPath, 'sign.jar');
     log.debug("Resigning apk.");


mykola-mokhnach:
What is Java9 signing problem?

NozomiIto:
@mykola-mokhnach Finally I identified the problem is that apksigner call fails with Java9 if Appium server path contains space character. 

```
Cannot use apksigner tool for signing. Defaulting to sign.jar. Original error: Command '/Users/itonozomi/Library/Android/sdk/build-tools/27.0.2/apksigner sign --key '/Users/itonozomi/Documents/GitHub/appi um/node_modules/appium-adb/keys/testkey.pk8' --cert '/Users/itonozomi/Documents/GitHub/appi um/node_modules/appium-adb/keys/testkey.x509.pem' '/Users/itonozomi/Documents/GitHub/appi um/node_modules/appium-uiautomator2-server/apks/appium-uiautomator2-server-v1.4.0.apk'' exited with code 1; StdErr: Unexpected parameter(s) after input APK (--cert)
```

full log: https://gist.github.com/NozomiIto/4d28451b33162344718cea0eff91b511
problem detail: https://github.com/Magic-Pod/AppiumRegressionCheck/issues/9

Now that Appium stability is very important for my business, I recently created many real device regression test cases for Appium and published it on [Magic-Pod/AppiumRegressionCheck](https://github.com/Magic-Pod/AppiumRegressionCheck/blob/master/test/testAppium.js), and @KazuCocoa also helps this work.
All apps used in this test are included in this repository or pre-installed app, so you can easily reproduce various problems. 
For example, if you want to reproduce this Java9 problem, 
just rename appium server dir like to `appi um` and run like

```
APPIUM_MAIN_JS_PATH_FOR_MAGIC_POD="/Users/itonozomi/Documents/GitHub/appi um/build/lib/main.js" npm test -- --grep "Java9"
```

mykola-mokhnach:
It is quite strange, because I see all arguments are properly quoted. Do you observe the same error if the command above is executed manually?

mykola-mokhnach:
> No, this problem does not happen if I manually run the following..

Magic... I'd add some logging into apksigner script itself, so it can print arguments list when the command is executed by Appium.

KazuCocoa:
Just caught up on the issue. hmm 🤔

mykola-mokhnach:
@imurchie Do you have any ideas why this problem may happen?

imurchie:
The arguments are not actually escaped correctly. Try adding, in `lib/tools/apk-signing.js` (right before https://github.com/appium/appium-adb/blob/master/lib/tools/apk-signing.js#L40):
```js
import { quote } from 'shell-quote';

// ...
apkSigningMethods.executeApksigner = async function (args = []) {
  // ...
  args = args.map((arg) => quote([arg]));
  // ...
};
```

mykola-mokhnach:
@imurchie shouldn't teen_process do that automatically?

mykola-mokhnach:
@NozomiIto Can you please add the patch proposed by @imurchie to the PR and check if it fixes the spacing issue?

NozomiIto:
OK.

PR #315, opened 2018-04-10
Java9 uiautomator2 test did not work if Appium server path contains w…

NozomiIto:
…hite space

Related to https://github.com/appium/appium-adb/pull/314

### DIFF
--- a/lib/tools/apk-signing.js
+++ b/lib/tools/apk-signing.js
@@ -1,5 +1,6 @@
 import _ from 'lodash';
 import { exec } from 'teen_process';
+import { quote } from 'shell-quote';
 import path from 'path';
 import log from '../logger.js';
 import { tempDir, system, mkdirp, fs, zip } from 'appium-support';
@@ -37,6 +38,7 @@ apkSigningMethods.executeApksigner = async function (args = []) {
       isPatched = true;
     }
   }
+  args = args.map((arg) => quote([arg]));
   log.debug(`Starting ${isPatched ? 'patched ' : ''}'${apkSigner}' with args '${args}'`);
   try {
     return (await exec(apkSigner, args, {
@@ -70,7 +72,7 @@ apkSigningMethods.signWithDefaultCert = async function (apk) {
     await this.executeApksigner(args);
   } catch (err) {
     log.warn(`Cannot use apksigner tool for signing. Defaulting to sign.jar. ` +
-             `Original error: ${err.message}`);
+             `Original error: ${err.message}` + (err.stderr ? `; StdErr: ${err.stderr}` : ''));
     const java = getJavaForOs();
     const signPath = path.resolve(this.helperJarPath, 'sign.jar');
     log.debug("Resigning apk.");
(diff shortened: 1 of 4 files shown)


lib/tools/apk-utils.js
```diff
@@ -359,7 +360,8 @@ apkUtilsMethods.install = async function (apk, options = {}) {
   }
 
   const executeInstall = async (args) => {
-    const output = await this.adbExec(['install', ...args, apk], {timeout});
+    args = args.map((arg) => quote([arg]));
+    const output = await this.adbExec(['install', ...args, quote([apk])], {timeout});
```
mykola-mokhnach: was this piece of code also problematic with spaces?

imurchie:
Please run the tests locally and fix the errors produced.

lib/tools/apk-utils.js
NozomiIto: apk path here also must be quoted (since ui-automator server APK path also contains space char if Appium server contains space char).
But quotation for other args are redundant, so I will remove it.

NozomiIto:
> Please run the tests locally and fix the errors produced.

I found semi-colon in apksigner arguments was unexpectedly escaped by shell-quote (like `"pass\:android"`), and it made the unit test failed.

Now I carefully quoted only APK file path to avoid such problem.

I also found multiple quotation calls can cause the unexpected result. (For example, if apk path is already quoted like `"<path to apk>"`, shell-quote call for it causes redundant quote like `'"<path to apk>"'`)

I think quotation must be done carefully, and then I added API comments to `systemCallMethods.adbExec` and `apkSigningMethods.executeApksigner` about whether a method argument must be quoted beforehand to

mykola-mokhnach:
@NozomiIto I have one more idea on why it does not handle arguments with spaces properly. Does it also fail if you remove `{shell: true}` option from exec and remove the quoting helpers?

NozomiIto:
@mykola-mokhnach 
Thanks for the advice.
Wow, if I remove `{shell: true}`, teen_process `exec` does the quotation correctly..
Do you know why some exec calls use `shell: true` but other calls don't use `shell: true` in `lib/tools/apk-signing.js`?

mykola-mokhnach:
> Do you know why some exec calls use shell: true but other calls don't use shell: true in lib/tools/apk-signing.js?

I wrote that code. Perhaps I had some issues and tried this option to resolve them and then forgot to remove it. Anyway. If you see the quotation is correct now then please modify the PR and I'll be happy to merge it then. TIA

lib/tools/apk-utils.js
NozomiIto: I finally found this fix quotation is unnecessary (sorry!), and removed it.

NozomiIto:
I removed shell option and fixed the corresponding test.
Now it works well on my local environment with space Appium server path.

lib/tools/apk-signing.js
```diff
@@ -70,7 +69,7 @@ apkSigningMethods.signWithDefaultCert = async function (apk) {
     await this.executeApksigner(args);
   } catch (err) {
     log.warn(`Cannot use apksigner tool for signing. Defaulting to sign.jar. ` +
-             `Original error: ${err.message}`);
+             `Original error: ${err.message}` + (err.stderr ? `; StdErr: ${err.stderr}` : ''));
```
mykola-mokhnach: I assume this is already merged in the other PR

PR #317, opened 2018-04-20
Add a method to remove the previously assigned logcat listener

mykola-mokhnach:
This will allow to avoid unexpected memory leaks

### DIFF
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -1033,7 +1033,7 @@ methods.getLogcatLogs = function () {
  */
 methods.setLogcatListener = function (listener) {
   if (_.isEmpty(this.logcat)) {
-    throw new Error("Can't get logcat logs since logcat hasn't started");
+    throw new Error("Logcat process hasn't been started");
   }
   this.logcat.on('output', listener);
 };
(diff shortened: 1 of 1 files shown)


lib/tools/adb-commands.js
```diff
@@ -1025,19 +1025,33 @@ methods.getLogcatLogs = function () {
...
 };
 
+/**
+ * Removes the previously set callback for the logcat output event.
+ *
+ * @param {Function} listener - The listener function, which has been previously
+ *                              passed to `setLogcatListener`
+ * @throws {Error} If logcat process is not running.
+ */
+methods.removeLogcatListener = function (listener) {
```
jlipps: looks good. i often also find it useful to use the [`removeAllListeners`](https://nodejs.org/api/events.html#events_emitter_removealllisteners_eventname) method which i think is available on EventEmitters. sometimes you don't have access to the original listener and don't mind removing all of them.

mykola-mokhnach: yep, I know about it. Although here you can only assign a listener by using `setLogcatListener` method, so the client knows about the assigned callbacks.

PR #318, opened 2018-04-23
Add a possibility to set user profile index while launching applications

mykola-mokhnach:
Also improved the error messages there and added nicer documentation

### DIFF
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -53,56 +53,76 @@ apkUtilsMethods.startUri = async function (uri, pkg) {
 };
 
 /**
- * Start the particular package on the device under test.
+ * @typedef {Object} StartAppOptions
+ * @property {!string} activity - The name of the main application activity
+ * @property {!string} pkg - The name of the application package
+ * @property {?boolean} retry [true] - If this property is set to `true`
+ * and the activity name does not start with '.' then the method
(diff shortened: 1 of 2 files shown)


lib/tools/apk-utils.js
```diff
@@ -53,56 +53,76 @@ apkUtilsMethods.startUri = async function (uri, pkg) {
...
-    let apiLevel = await this.getApiLevel();
-    let cmd = buildStartCmd(startAppOptions, apiLevel);
-    let stdout = await this.shell(cmd);
-    if (stdout.indexOf("Error: Activity class") !== -1 &&
-        stdout.indexOf("does not exist") !== -1) {
-      if (startAppOptions.retry && startAppOptions.activity[0] !== ".") {
-        log.debug("We tried to start an activity that doesn't exist, " +
-                  "retrying with . prepended to activity");
+    const stdout = await this.shell(cmd);
+    if (stdout.includes("Error: Activity class") && stdout.includes("does not exist")) {
```
jlipps: would a single regex test more elegantly capture this condition?

annapurna14242:
Hi @mykola-mokhnach : Can you please suggest, how to use this inside framework to launch app from work profile?
I tried with user capability, but still app is not launching from work profile
''user" : "10"
Thanks in Advance!

PR #319, opened 2018-04-26
Return standard error from adb commands

tfcardoso:
Currently ADB commands are only returning standard output. On some cases standard error is also filled and should be returned.

For instance running the following adb command (an activity not present in device) in Android 8:
    
    adb shell am start -n 'com.android.settings/.Settings\$AccountsGroupSettingsActivity'

using:

    driver.executeScript("mobile: shell",commandAndArguments);

returns output:

    Starting: Intent { cmp=com.android.settings/.Settings\$AccountsGroupSettingsActivity }

and complete output with stdout and stderr is:

    Starting: Intent { cmp=com.android.settings/.Settings\$AccountsGroupSettingsActivity }
    Error type 3
    Error: Activity class {com.android.settings/com.android.settings.Settings\$AccountsGroupSettingsActivity} does not exist.

### DIFF
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -294,11 +294,11 @@ systemCallMethods.adbExec = async function (cmd, opts = {}) {
       }
       let args = this.executable.defaultArgs.concat(cmd);
       log.debug(`Running '${this.executable.path} ${quote(args)}'`);
-      let {stdout} = await exec(this.executable.path, args, opts);
+      let {stdout, stderr} = await exec(this.executable.path, args, opts);
       // sometimes ADB prints out weird stdout warnings that we don't want
       // to include in any of the response data, so let's strip it out
       stdout = stdout.replace(linkerWarningRe, '').trim();
-      return stdout;
+      return stderr ? stdout + '\n' + stderr : stdout;
     } catch (e) {
       let protocolFaultError = new RegExp("protocol fault \\(no status\\)", "i").test(e);
       let deviceNotFoundError = new RegExp("error: device ('.+' )?not found", "i").test(e);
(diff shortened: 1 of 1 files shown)


lib/tools/system-calls.js
```diff
@@ -294,11 +294,11 @@ systemCallMethods.adbExec = async function (cmd, opts = {}) {
       }
       let args = this.executable.defaultArgs.concat(cmd);
       log.debug(`Running '${this.executable.path} ${quote(args)}'`);
-      let {stdout} = await exec(this.executable.path, args, opts);
+      let {stdout, stderr} = await exec(this.executable.path, args, opts);
```
mykola-mokhnach: I thought that stderr is only present if error code is not equal to zero, but whatever

lib/tools/system-calls.js
```diff
@@ -294,11 +294,11 @@ systemCallMethods.adbExec = async function (cmd, opts = {}) {
...
       // sometimes ADB prints out weird stdout warnings that we don't want
       // to include in any of the response data, so let's strip it out
       stdout = stdout.replace(linkerWarningRe, '').trim();
-      return stdout;
+      return stderr ? stdout + '\n' + stderr : stdout;
```
mykola-mokhnach: this should look like

```
_.isEmpty(stderr) ? stdout : `${stdout}\n${stderr}`;
```

mykola-mokhnach: also the docstring should be properly updated

tfcardoso:
Thanks for feedback @mykola-mokhnach. I made some changes according to your comments.

mykola-mokhnach:
In general, what you can do:

- Add `timeout` and `includeStderr` arguments to the `shell` call
- Change the returned result of this method based on `includeStderr` argument (single string stdout by default or a dictionary with stderr and stdout keys)

mykola-mokhnach:
Also, the changes above should be done inside android driver component

mykola-mokhnach:
https://github.com/appium/appium-android-driver/pull/356/files

lib/tools/system-calls.js
jlipps: i would also be fine parameterizing this method to optionally return stderr

PR #326, opened 2018-05-21
Fix the error message if the command is of type string

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/10725


mykola-mokhnach:
Also I'm thinking about removing `isDeviceConnected` at all, since it is executed each time `adb shell` is invoked, which slows down test execution in general. adbExec is anyway supposed to fail if the device has been unexpectedly disconnected.

mykola-mokhnach:
WDYT @imurchie ?

imurchie:
How fast does `adb shell` fail if the device is not connected?

mykola-mokhnach:
adbExec is set to retry once by default, but it is anyway faster than executing `adb devices` for each shell command

imurchie:
Published in `appium-adb@6.10.0`.

PR #333, opened 2018-06-26
Make killEmulator method work for current emu

imurchie:


lib/tools/system-calls.js
```diff
@@ -546,13 +546,20 @@ systemCallMethods.killAllEmulators = async function () {
...
+      await this.adbExec(['emu', 'kill']);
+      log.info(`successfully killed emulator '${avdName}'`);
+    } else {
+      log.info(`no avd with name '${avdName}' running. skipping kill step.`);
+    }
   } else {
-    log.info(`no avd with name '${avdName}' running. skipping kill step.`);
+    // killing the current avd
+    log.debug(`killing avd with id '${this.curDeviceId}'`);
+    await this.adbExec(['emu', 'kill']);
```
mykola-mokhnach: should we check if the emu is running first?

lib/tools/system-calls.js
```diff
@@ -546,13 +546,20 @@ systemCallMethods.killAllEmulators = async function () {
...
-    await this.adbExec(['emu', 'kill']);
-    log.info(`successfully killed emulator '${avdName}'`);
+  if (util.hasValue(avdName)) {
+    log.debug(`killing avd '${avdName}'`);
+    let device = await this.getRunningAVD(avdName);
+    if (device) {
+      await this.adbExec(['emu', 'kill']);
+      log.info(`successfully killed emulator '${avdName}'`);
+    } else {
+      log.info(`no avd with name '${avdName}' running. skipping kill step.`);
```
mykola-mokhnach: We could probably return false or true based on the fact that emulator has been killed or not

lib/tools/system-calls.js
imurchie: Sure. It is checked in the other section, too.

lib/tools/system-calls.js
```diff
@@ -546,14 +546,24 @@ systemCallMethods.killAllEmulators = async function () {
  * @param {string} avdName - The name of the emulator to be killed.
```
mykola-mokhnach: return description is missing. Also it worth to mention about the argument is optional

imurchie: This is why I hate docstrings.

PR #334, opened 2018-07-02
Add a method for custom certificate installation

mykola-mokhnach:
Related to https://github.com/appium/appium-android-driver/pull/386

### DIFF
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -901,7 +901,7 @@ systemCallMethods.fileSize = async function (remotePath) {
  * or if there was an error while installing the certificate
  */
 systemCallMethods.installMitmCertificate = async function (cert) {
-  const openSslBinaryName = await getOpenSslForOs();
+  const openSsl = await getOpenSslForOs();
 
   if (!_.isBuffer(cert)) {
     cert = Buffer.from(cert, 'base64');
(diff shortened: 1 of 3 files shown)


lib/tools/system-calls.js
```diff
@@ -883,4 +883,75 @@ systemCallMethods.fileSize = async function (remotePath) {
...
+      await fs.writeFile(dstCert.path, dstCertContent);
+      log.debug('Mounting /system endpoint');
+      await this.shell(['mount', '-o', 'rw,remount', '/system']);
+      const dstPath = '/system/etc/security/cacerts';
+      log.debug(`Uploading the generated certificate from '${dstCert.path}' to '${dstPath}'`);
+      await this.push(dstCert.path, dstPath);
+      log.debug('Remounting the remote file system');
+      await this.adbExec(['remount']);
+    } finally {
+      await fs.rimraf(dstCert.path);
```
imurchie: Does deleting this cert destroy the file descriptor? How is that resource handled?

lib/tools/system-calls.js
```diff
@@ -883,4 +883,75 @@ systemCallMethods.fileSize = async function (remotePath) {
...
+    }
+  } catch (err) {
+    throw new Error(`Cannot inject the custom certificate. ` +
+                    `Is the certificate properly encoded into base64-string? ` +
+                    `Do you have root permissions on the device? ` +
+                    `Original error: ${err.message}`);
+  } finally {
+    await fs.rimraf(tempCert.path);
```
imurchie: Same question.

lib/tools/system-calls.js
mykola-mokhnach: I don't get the question. The idea is to free the place as soon as possible, otherwise these temp files are going to be deleted, probably, after the next OS reboot.

imurchie: The question is in regards to `dstCert`, which is created by [tempDir.open](https://github.com/appium/appium-support/blob/master/lib/tempdir.js#L32-L42). This means that the file is open. I'm wondering if, in practice, running `fs.rimraf` frees all those resources.

mykola-mokhnach: @imurchie I'm unable to find the exact description of what is going to happen in such case (perhaps, this depends on low-level OS implementation), but I assume it would be safer to call `fs.close` before doing file removal.

imurchie:
Published in `appium-adb@6.13.0`.

PR #335, opened 2018-07-09
add hide api policy settings for Android P

KazuCocoa:
Commands mentioned in https://github.com/appium/io.appium.settings/pull/15

Android P emulator returns 27 as the emulator's API level.
Thus, we can't filter if we send this command or not from API Level...

### DIFF
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -345,6 +345,28 @@ methods.toggleGPSLocationProvider = async function (enabled) {
 };
 
 /**
+ * Permit access to all non-SDK APIs, but keep warnings in the log.
+ * https://developer.android.com/preview/restrictions-non-sdk-interfaces
+ *
+ * Commands don't lead crash or something, thus simply this method calls putting all setting.
+ */
+methods.hiddenApiPolicyP = async function () {
+  await this.setSetting('global', 'hidden_api_policy_pre_p_apps', '1');
+  await this.setSetting('global', 'hidden_api_policy_p_apps', '1');
+};
+
+/**
+ * Reset accessing to non-SDK APIs to the default setting
+ * https://developer.android.com/preview/restrictions-non-sdk-interfaces
+ *
+ * Commands don't lead crash or something, thus simply this method calls deleting all setting.
+ */
+methods.defaultHiddenApiPolicyP = async function () {
+  await this.shell(['settings', 'delete', 'global', 'hidden_api_policy_pre_p_apps']);
+  await this.shell(['settings', 'delete', 'global', 'hidden_api_policy_p_apps']);
+};
+
+/**
  * Stop the particular package if it is running and clears its application data.
  *
  * @param {string} pkg - The package name to be processed.
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -350,7 +350,7 @@ methods.toggleGPSLocationProvider = async function (enabled) {
  *
  * Commands don't lead crash or something, thus simply this method calls putting all setting.
  */
-methods.hiddenApiPolicyP = async function () {
+methods.relaxHiddenApiPolicy = async function () {
   await this.setSetting('global', 'hidden_api_policy_pre_p_apps', '1');
   await this.setSetting('global', 'hidden_api_policy_p_apps', '1');
 };
@@ -361,7 +361,7 @@ methods.hiddenApiPolicyP = async function () {
  *
  * Commands don't lead crash or something, thus simply this method calls deleting all setting.
  */
-methods.defaultHiddenApiPolicyP = async function () {
+methods.setDefaultHiddenApiPolicy = async function () {
   await this.shell(['settings', 'delete', 'global', 'hidden_api_policy_pre_p_apps']);
   await this.shell(['settings', 'delete', 'global', 'hidden_api_policy_p_apps']);
 };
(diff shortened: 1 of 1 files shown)


mykola-mokhnach:
> I'm not a big fan of these function names

I second that

lib/tools/adb-commands.js
```diff
@@ -361,7 +361,7 @@ methods.hiddenApiPolicyP = async function () {
  *
  * Commands don't lead crash or something, thus simply this method calls deleting all setting.
  */
-methods.defaultHiddenApiPolicyP = async function () {
+methods.setDefaultHiddenApiPolicy = async function () {
```
mykola-mokhnach: how about `enableNonSdkApiAccess`/`disableNonSdkApiAccess`?

KazuCocoa: I also consider so.  But the default preference probably depends on companies which customise the devices, I guess. Google defines as below. If we use _enable/disable_, putting _2 or 3_/_0 or 1_ is probably proper. But if the value isn't the default value for the device, it isn't suitable for Appium to reset the value. So, I chose `relax` and `setDefault`

```
0: Disable non-SDK API usage detection. This will also disable logging, and also break the strict mode API, detectNonSdkApiUsage(). Not recommended.
1: "Just warn" - permit access to all non-SDK APIs, but keep warnings in the log. The strict mode API will keep working.
2: Disallow usage of dark grey and black listed APIs.
3: Disallow usage of blacklisted APIs, but allow usage of dark grey listed APIs.
```

Do you think enable/disable are proper?

mykola-mokhnach: Then I would rather have two methods for resettings the values to their defaults and for setting them to the particular constant from the list above. It might be also handy to mention the stuff above in the docstring

lib/tools/adb-commands.js
```diff
@@ -344,6 +344,36 @@ methods.toggleGPSLocationProvider = async function (enabled) {
...
+ * @param {number} value - The API enforcement policy.
+ *     0: Disable non-SDK API usage detection. This will also disable logging, and also break the strict mode API,
+ *        detectNonSdkApiUsage(). Not recommended.
+ *     1: "Just warn" - permit access to all non-SDK APIs, but keep warnings in the log.
+ *        The strict mode API will keep working.
+ *     2: Disallow usage of dark grey and black listed APIs.
+ *     3: Disallow usage of blacklisted APIs, but allow usage of dark grey listed APIs.
+ */
+methods.setHiddenApiPolicy = async function (value) {
+  await this.setSetting('global', 'hidden_api_policy_pre_p_apps', value.toString());
```
mykola-mokhnach: `toString` is redundant. Also the parameter type could be either string or number (`'1'` and `1` will work equally)

lib/tools/adb-commands.js
```diff
@@ -344,6 +344,36 @@ methods.toggleGPSLocationProvider = async function (enabled) {
   await this.setSetting('secure', 'location_providers_allowed', `${enabled ? "+" : "-"}gps`);
 };
 
+/**
+ * Permit access to all non-SDK APIs, but keep warnings in the log.
+ * https://developer.android.com/preview/restrictions-non-sdk-interfaces
+ *
+ * Commands don't lead crash or something, thus simply this method calls putting all setting.
```
mykola-mokhnach: I think this sentence could be safely skipped, since the link to the detailed description is already available.

lib/tools/adb-commands.js
```diff
@@ -344,6 +344,36 @@ methods.toggleGPSLocationProvider = async function (enabled) {
...
+methods.setHiddenApiPolicy = async function (value) {
+  await this.setSetting('global', 'hidden_api_policy_pre_p_apps', value.toString());
+  await this.setSetting('global', 'hidden_api_policy_p_apps', value.toString());
+};
+
+/**
+ * Reset accessing to non-SDK APIs to the default setting
+ * https://developer.android.com/preview/restrictions-non-sdk-interfaces
+ *
+ * Commands don't lead crash or something, thus simply this method calls deleting all setting.
```
mykola-mokhnach: same comment

lib/tools/adb-commands.js
```diff
@@ -344,6 +344,36 @@ methods.toggleGPSLocationProvider = async function (enabled) {
...
+ *     2: Disallow usage of dark grey and black listed APIs.
+ *     3: Disallow usage of blacklisted APIs, but allow usage of dark grey listed APIs.
+ */
+methods.setHiddenApiPolicy = async function (value) {
+  await this.setSetting('global', 'hidden_api_policy_pre_p_apps', value.toString());
+  await this.setSetting('global', 'hidden_api_policy_p_apps', value.toString());
+};
+
+/**
+ * Reset accessing to non-SDK APIs to the default setting
```
mykola-mokhnach: -> Reset access to non-SDK APIs to its default setting

lib/tools/adb-commands.js
```diff
@@ -345,11 +345,9 @@ methods.toggleGPSLocationProvider = async function (enabled) {
 };
 
 /**
- * Permit access to all non-SDK APIs, but keep warnings in the log.
+ * Set hidden api policy to manage access to non-SDK APIs.
  * https://developer.android.com/preview/restrictions-non-sdk-interfaces
  *
- * Commands don't lead crash or something, thus simply this method calls putting all setting.
- *
  * @param {number} value - The API enforcement policy.
```
mykola-mokhnach: `{number|string}`

mykola-mokhnach:
@imurchie Can you please take a look once again?

PR #340, opened 2018-07-27
Add screenrecord to adb commands

mykola-mokhnach:
I assume it would be smarter to wrap this command here up for further usage in https://github.com/appium/appium-android-driver/pull/399

### DIFF
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -1520,7 +1520,7 @@ methods.bugreport = async function (timeout = 120000) {
  * @param {?ScreenrecordOptions} options [{}]
  * @returns {SubProcess} screenrecord process, which can be then controlled by the client code
  */
-methods.screenrecord = async function (destination, options = {}) {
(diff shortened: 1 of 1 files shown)


lib/tools/adb-commands.js
```diff
@@ -1495,4 +1495,57 @@ methods.bugreport = async function (timeout = 120000) {
...
+    cmd.push('--time-limit', timeLimit);
+  }
+  if (util.hasValue(bitRate)) {
+    cmd.push('--bit-rate', bitRate);
+  }
+  if (bugReport) {
+    cmd.push('--bugreport');
+  }
+  cmd.push(destination);
+  return new SubProcess(this.executable.path, [
```
jlipps: we're intentionally just returning the process obj and not starting it?

mykola-mokhnach: yes, we need full control over it. See https://github.com/appium/appium-android-driver/pull/399/files for more details

PR #341, opened 2018-07-28
Improve logging for the screenrecord command

mykola-mokhnach:


lib/tools/adb-commands.js
```diff
@@ -1541,12 +1541,14 @@ methods.screenrecord = function (destination, options = {}) {
...
-  log.debug(`Building screenrecord process with the command line: adb shell ${cmd.join(' ')}`);
-  return new SubProcess(this.executable.path, [
+
+  const fullCmd = [
     ...this.executable.defaultArgs,
     'shell',
     ...cmd
-  ]);
+  ];
+  log.debug(`Building screenrecord process with the command line: ${fullCmd.join(' ')}`);
```
jlipps: why not use a shlex library so we see the actual command which will be run?

mykola-mokhnach: changed to shell-quote call

PR #343, opened 2018-08-07
Do not throw an exception if emulator process dies asynchronously

mykola-mokhnach:
See https://github.com/appium/appium/issues/10964#issuecomment-410871599

### DIFF
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -642,10 +642,8 @@ systemCallMethods.launchAVD = async function (avdName, avdArgs, language, countr
       log.info(`[AVD OUTPUT] ${line}`);
     }
   });
-  proc.on('exit', (code, signal) => {
-    if (code !== 0) {
-      throw new Error(`Emulator avd ${avdName} exit with code ${code}, signal ${signal}`);
-    }
+  proc.on('die', (code, signal) => {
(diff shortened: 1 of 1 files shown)


lib/tools/system-calls.js
```diff
@@ -973,6 +971,10 @@ systemCallMethods.installMitmCertificate = async function (cert) {
 systemCallMethods.isMitmCertificateInstalled = async function (cert) {
   const openSsl = await getOpenSslForOs();
 
+  if (!_.isBuffer(cert)) {
+    cert = Buffer.from(cert, 'base64');
```
KazuCocoa: 👀

lib/tools/system-calls.js
```diff
@@ -642,10 +642,8 @@ systemCallMethods.launchAVD = async function (avdName, avdArgs, language, countr
       log.info(`[AVD OUTPUT] ${line}`);
     }
   });
-  proc.on('exit', (code, signal) => {
-    if (code !== 0) {
-      throw new Error(`Emulator avd ${avdName} exit with code ${code}, signal ${signal}`);
-    }
+  proc.on('die', (code, signal) => {
+    log.warn(`Emulator avd ${avdName} exited with code ${code}${signal ? `, signal ${signal}` : ''}`);
```
imurchie: Yes, throwing in an event handler does nothing helpful :(

PR #346, opened 2018-08-27
Improve timer loop for emulator boot

mykola-mokhnach:


lib/tools/system-calls.js
```diff
@@ -727,20 +721,27 @@ systemCallMethods.checkAvdExist = async function (avdName) {
  * @throws {Error} If the emulator is not ready within the given timeout.
```
imurchie: Return value should be documented.

lib/tools/system-calls.js
```diff
@@ -727,20 +721,27 @@ systemCallMethods.checkAvdExist = async function (avdName) {
...
-  log.debug("Waiting until emulator is ready");
-  while ((Date.now() - start) < timeoutMs) {
-    try {
-      let stdout = await this.shell(["getprop", "init.svc.bootanim"]);
-      if (stdout.indexOf('stopped') > -1) {
-        return;
+  try {
+    await waitForCondition(async () => {
+      try {
+        if (!await this.shell(['getprop', 'init.svc.bootanim']).includes('stopped')) {
```
imurchie: I don't think this condition will ever work, since you are checking `includes` of the promise returned by `this.shell`. Some parentheses need to be added.

lib/tools/system-calls.js
```diff
@@ -727,20 +721,27 @@ systemCallMethods.checkAvdExist = async function (avdName) {
...
+          return false;
+        }
+        // Sometimes the package manager service might still being initialized
+        // on slow systems even after emulator booting is completed
+        const installLocation = await this.shell(['pm', 'get-install-location']);
+        return /\d+\[\w+\]/.test(installLocation);
```
imurchie: Some explanation of what this regular expression is checking would be good.

lib/tools/system-calls.js
mykola-mokhnach: the method always returns undefined

lib/tools/system-calls.js
mykola-mokhnach: yep, good catch

lib/tools/system-calls.js
mykola-mokhnach: ok will add

lib/tools/system-calls.js
imurchie: Really? Why are lines 728 and 736 returning `false`?

PR #349, opened 2018-09-05
Add a possibility to retrieve geo location

mykola-mokhnach:
The PR is based on https://github.com/appium/io.appium.settings/pull/16


mykola-mokhnach:
should work for both

mykola-mokhnach:
I've only tested on a real device though

PR #350, opened 2018-09-06
Add parsing of altitude value

mykola-mokhnach:
Selenium API also requires `altitude` value to set. The PR depends on https://github.com/appium/io.appium.settings/pull/17

### DIFF
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -856,14 +856,16 @@ methods.setDeviceSysLocaleViaSettingApp = async function (language, country) {
 
 /**
  * @typedef {Object} Location
- * @property {float|string} longitude - Valid longitude value.
- * @property {float|string} latitude - Valid latitude value.
+ * @property {number|string} longitude - Valid longitude value.
+ * @property {number|string} latitude - Valid latitude value.
+ * @property {?number|string} altitude - Valid altitude value.
  */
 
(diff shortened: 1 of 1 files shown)


lib/tools/adb-commands.js
```diff
@@ -913,13 +915,14 @@ methods.getGeoLocation = async function () {
       `services must be enabled on the device. Original error: ${err.message}`);
   }
 
-  const match = /data="(-?[\d\.]+)\s+(-?[\d\.]+)"/.exec(output);
+  const match = /data="(-?[\d\.]+)\s+(-?[\d\.]+)\s+(-?[\d\.]+)"/.exec(output);
```
jlipps: does this regex allow for the third number to be absent in the case of running an older appium settings?

mykola-mokhnach: no it does not, but I assume there will by anyway the minimum package version set in android driver, so we don't have to care much about it

jlipps: ok, as long as android driver performs an upgrade on the settings app if necessary 👍

mykola-mokhnach: yes it does
https://github.com/appium/appium-android-driver/blob/05e7a982f4c44c7bd09922caf02079d3e561681c/lib/android-helpers.js#L438

PR #354, opened 2018-09-22
Add a helper for performing editor actions

mykola-mokhnach:
The PR depends on https://github.com/appium/io.appium.settings/pull/18

### DIFF
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -1610,7 +1610,7 @@ methods.screenrecord = function (destination, options = {}) {
 };
 
 /**
- * Performs the given editor action of the focused input field.
+ * Performs the given editor action on the focused input field.
  * This method requires Appium Settings helper to be installed on the device.
  * No exception is thrown if there was a failure while performing the action.
  * You must investigate the logcat output if something did not work as expected.
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -1625,7 +1625,7 @@ methods.performEditorAction = async function (action) {
   await this.enableIME(APPIUM_IME);
   try {
     await this.setIME(APPIUM_IME);
-    await this.shell('input', 'text', `/${action}/`);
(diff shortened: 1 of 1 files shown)


lib/tools/adb-commands.js
```diff
@@ -1608,4 +1609,26 @@ methods.screenrecord = function (destination, options = {}) {
...
+ *                                 names are supported: `normal, unspecified, none,
+ *                                 go, search, send, next, done, previous`
+ */
+methods.performEditorAction = async function (action) {
+  log.debug(`Performing editor action: ${action}`);
+  const defaultIME = await this.defaultIME();
+  await this.enableIME(APPIUM_IME);
+  try {
+    await this.setIME(APPIUM_IME);
+    await this.shell('input', 'text', `/${action}/`);
```
KazuCocoa: `[]` is needed
`await this.shell(['input', 'text', `/${action}/`]);`

mykola-mokhnach: good catch. thanks

lib/tools/adb-commands.js
```diff
@@ -1608,4 +1609,26 @@ methods.screenrecord = function (destination, options = {}) {
...
+ */
+methods.performEditorAction = async function (action) {
+  log.debug(`Performing editor action: ${action}`);
+  const defaultIME = await this.defaultIME();
+  await this.enableIME(APPIUM_IME);
+  try {
+    await this.setIME(APPIUM_IME);
+    await this.shell(['input', 'text', `/${action}/`]);
+  } finally {
+    await this.setIME(defaultIME);
```
jlipps: do we want to set it back to the default? what if  the user has already set their own ime? maybe we should get the current IME instead? (could also then save some time if for whatever reason it happens to already be the appium IME)

mykola-mokhnach: actually the default ime is the currently selected one. It is just the confusing name of the function

PR #356, opened 2018-09-23
Speed up permissions detection

mykola-mokhnach:
The idea behind this is the same as for https://github.com/appium/appium-adb/pull/355
Package manager invocation is more expensive from performance perspective in comparison to dumpsys call

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -322,10 +322,7 @@ const getBuildToolsDirs = _.memoize(async function getBuildToolsDirs (sdkRoot) {
 });
 
 /**
- * Retrieves permissions list encoded in `dumpsys package`
- * command output. It is expected that the first line of
- * this output is the title of the corresponding permissions group
- * and it contains the starting indent.
+ * Retrieves the list of permission names encoded in `dumpsys package` command output.
  *
  * @param {string} dumpsysOutput - The actual command output.
  * @param {Array<string>} groupNames - The list of group names to list permissions for.
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -321,8 +321,56 @@ const getBuildToolsDirs = _.memoize(async function getBuildToolsDirs (sdkRoot) {
   return buildToolsDirs;
 });
 
+/**
+ * @typedef {Object} PermissionInfo
+ * @property {string} permission - The name of the corresponding permission
+ * @property {?boolean} granted - Whether the permission is granted or not
+ */
+
+/**
+ * Retrieves permissions list encoded in `dumpsys package`
+ * command output. It is expected that the first line of
+ * this output is the title of the corresponding permissions group
+ * and it contains the starting indent.
+ *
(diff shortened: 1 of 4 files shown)


lib/helpers.js
```diff
@@ -321,8 +321,57 @@ const getBuildToolsDirs = _.memoize(async function getBuildToolsDirs (sdkRoot) {
...
+  const lines = dumpsysOutput.split('\n');
+  if (lines.length < 2) {
+    return [];
+  }
+
+  const indentPattern = /\S|$/;
+  const titleIndent = lines[0].search(indentPattern);
+  const permissionNamePattern = /android\.permission\.\w+/;
+  const grantedStatePattern = /\bgranted=(\w+)/;
+  const result = [];
```
jlipps: i know it's valid but it seems weird to `push` to a `const` array

lib/tools/adb-commands.js
```diff
@@ -275,56 +277,81 @@ methods.revokePermission = async function (pkg, permission) {
...
  */
 methods.getGrantedPermissions = async function (pkg, cmdOutput = null) {
-  let stdout = cmdOutput || await this.shell(['pm', 'dump', pkg]);
-  let match = new RegExp(/(install permissions:|User 0)([\s\S]*?)DUMP OF SERVICE activity:/g).exec(stdout);
-  if (!match) {
-    throw new Error('Unable to get granted permissions');
+  const stdout = cmdOutput || await this.shell(['dumpsys', 'package', pkg]);
+  const grantedPermissions = [];
+  for (const pattern of [
+    new RegExp(/^(\s*install permissions:[\s\S]+)/, 'm'),
```
jlipps: might be more readable to make this array its own const

lib/tools/adb-commands.js
```diff
@@ -275,56 +277,81 @@ methods.revokePermission = async function (pkg, permission) {
...
- *                                    if present.
+ *                                    _dumpsys package_ command. It may speed up the method execution.
  * @return {array} The list of denied permissions or an empty list.
  */
 methods.getDeniedPermissions = async function (pkg, cmdOutput = null) {
-  let stdout = cmdOutput || await this.shell(['pm', 'dump', pkg]);
-  let match = new RegExp(/(install permissions:|User 0)([\s\S]*?)DUMP OF SERVICE activity:/g).exec(stdout);
-  if (!match) {
-    throw new Error('Unable to get denied permissions');
+  const stdout = cmdOutput || await this.shell(['dumpsys', 'package', pkg]);
```
jlipps: this is all the same logic as above; maybe factor out one method that just takes a bool to get granted vs denied?

mykola-mokhnach: I don't want to change module interface. And there is also not much content to extract into a common function

mykola-mokhnach:
@jlipps I did some refactoring there and now the amount of redundant code has been slightly decreased. Thanks for the hint

PR #359, opened 2018-09-26
Update CI config to include Android 7 (API 24)

mykola-mokhnach:
android 5 is missing some functional features we need to test

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,26 +1,20 @@
 language: android
-dist: precise
+sudo: required
+jdk: oraclejdk8
+dist: trusty
 addons:
   apt:
     sources:
       - ubuntu-toolchain-r-test
     packages:
-      - g++-4.9
+      - g++-6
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -1,60 +1,68 @@
 language: android
-dist: precise
+dist: trusty
```
imurchie: In my experience Android never works on Travis's "trusty" distribution.

mykola-mokhnach:
Now, finally, after several tens of commits, three days of hard work and four squashes we got our own Travis config with poker and housewives. 

Bender

mykola-mokhnach:
@jlipps Can you please check the PR again, as there were quite a lot of changes since the last review?

.travis.yml
mykola-mokhnach: it is strange, but it works with `trusty` now

PR #365, opened 2018-10-12
Make sure we try verify.jar if apksigner fails to verify

imurchie:
If `apksigner` is not there, we successfully make use of `verify.jar` to do certificate verification. But if it _is_ there, but fails, we also want to drop back to the older technology. This happens consistently, for instance, when Appium is running with Java 7 and the app was compiled with Java 8. In that case currently we _always_ assume it is unsigned and go through the whole signing process.

Also, there is no need to log stack traces into third party tools. So skip the stack trace for `apksigner` error logs.

Cc: @vrunoa

### DIFF
--- a/lib/tools/apk-signing.js
+++ b/lib/tools/apk-signing.js
@@ -70,7 +70,14 @@ apkSigningMethods.executeApksigner = async function (args = []) {
     log.warn(`Got an error during apksigner execution: ${err.message}`);
     for (const [name, stream] of [['stdout', err.stdout], ['stderr', err.stderr]]) {
       if (stream) {
-        log.warn(`apksigner ${name}: ${stream}`);
+        log.warn(`apksigner ${name}:`);
+        for (const line of stream.split('\n')) {
+          // we do not want to print out stack traces into Java code
+          if (_.isEmpty(line) || /\s+at\s/.test(line)) {
+            continue;
+          }
+          log.warn(`  ${line}`);
+        }
       }
     }
     if (system.isWindows()) {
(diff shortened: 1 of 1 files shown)


lib/tools/apk-signing.js
```diff
@@ -70,7 +70,14 @@ apkSigningMethods.executeApksigner = async function (args = []) {
     log.warn(`Got an error during apksigner execution: ${err.message}`);
     for (const [name, stream] of [['stdout', err.stdout], ['stderr', err.stderr]]) {
       if (stream) {
-        log.warn(`apksigner ${name}: ${stream}`);
+        log.warn(`apksigner ${name}:`);
```
mykola-mokhnach: I find the stacktraces useful though

lib/tools/apk-signing.js
```diff
@@ -241,27 +248,28 @@ apkSigningMethods.checkApkCert = async function (apk, pkg) {
...
-    };
+    const output = await this.executeApksigner(['verify', '--print-certs', apk]);
+    if (!_.includes(output, DEFAULT_CERT_DIGEST)) {
+      throw new Error(`'${apk}' is signed with non-default certificate`);
+    }
+    log.debug(`'${apk}' is already signed.`);
+    return true;
   } catch (e) {
-    log.warn(`Cannot use apksigner tool for signature verification. Defaulting to verify.jar. ` +
+    log.warn(`Cannot use apksigner tool for signature verification. ` +
```
mykola-mokhnach: do we want to retry even if the apk is not signed with non-default certificate?

lib/tools/apk-signing.js
```diff
@@ -70,7 +70,14 @@ apkSigningMethods.executeApksigner = async function (args = []) {
     log.warn(`Got an error during apksigner execution: ${err.message}`);
     for (const [name, stream] of [['stdout', err.stdout], ['stderr', err.stderr]]) {
       if (stream) {
-        log.warn(`apksigner ${name}: ${stream}`);
+        log.warn(`apksigner ${name}:`);
+        for (const line of stream.split('\n')) {
+          // we do not want to print out stack traces into Java code
+          if (_.isEmpty(line) || /\s+at\s/.test(line)) {
```
mykola-mokhnach: `line.trim()` ?

lib/tools/apk-signing.js
```diff
@@ -241,27 +248,28 @@ apkSigningMethods.checkApkCert = async function (apk, pkg) {
     return await this.checkCustomApkCert(apk, pkg);
```
mykola-mokhnach: should the same change be done to checkCustomApkCert ?

lib/tools/apk-signing.js
imurchie: Really? Into Android code? Like
```
Exception in thread "main" java.lang.UnsupportedClassVersionError: com/android/apksigner/ApkSignerTool : Unsupported major.minor version 52.0
	at java.lang.ClassLoader.defineClass1(Native Method)
 	at java.lang.ClassLoader.defineClass(ClassLoader.java:803)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:442)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:64)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:354)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:348)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:347)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:425)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:312)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:358)
	at sun.launcher.LauncherHelper.checkAndLoadMain(LauncherHelper.java:482
```

mykola-mokhnach: yep. that helped to investigate file access issue

lib/tools/apk-signing.js
```diff
@@ -234,34 +235,41 @@ apkSigningMethods.zipAlignApk = async function (apk) {
...
+    }
+    log.warn(`Cannot use apksigner tool for signature verification. ` +
+      `Original error: ${err.message}`);
   }
+
+  // default to verify.jar
   try {
-    await verificationFunc();
+    log.debug(`Defaulting to verify.jar`);
+    const java = getJavaForOs();
```
mykola-mokhnach: we probably don't need an intermediate variable for java path

PR #366, opened 2018-10-13
Add bundletool for .aab module

KazuCocoa:
Google start providing [appbundle](https://developer.android.com/platform/technology/app-bundle/). When we build release module with the app-bundle, `.aab` file is generated. Developers upload the `.aab` to the play console.

We can get `apks` file by https://github.com/google/bundletool on our local. We can install a proper apk included in the `apks` by the bundletool. https://github.com/KazuCocoa/AppBundleSample is an example to install a proper apk for connected device from `.aab` and `.apks`.

In this PR, I've added the 0.8 bundletool. Each command follow https://github.com/KazuCocoa/AppBundleSample which I ensured I could install a apk from aab correctly. (And could launch the installed apk)

We should take care of all of `apk` arguments. So, this is partial implementation.
One concern is a size of bundletool. It's 35MB so far...

What do you think adding `.aab` support?
(Eventyally, we'd like to make it enough to put `aab` and necessary arguments in capability by users. It means this module has a responsible for generating `apks` and install a proper apk for a connected device (which is a test device) via bundletool)

Another option is preparing a guide to get proper `apk` by users from `.aab` like https://codelabs.developers.google.com/codelabs/your-first-dynamic-app/index.html?index=..%2F..%2Fio2018#4 with unziping `.apks` and getting proper standalone apk.

### DIFF
--- a/lib/tools/bundletool.js
+++ b/lib/tools/bundletool.js
@@ -1,0 +1,114 @@
+import { exec } from 'teen_process';
+import log from '../logger.js';
+import path from 'path';
+import _ from 'lodash';
+import {getJavaForOs} from "../helpers";
+
+let bundletool = {};
+
+/**
+ * Check bundletool version
+ *
+ * @return {string} Version of bundletool. i.e., BundleTool 0.6.0
+ * @throws {Error} If get version fails
+ */
+bundletool.version = async function () {
+  const java = getJavaForOs();
+  const bundletoolPath = path.resolve(this.helperJarPath, 'bundletool.jar');
+  try {
+    return await exec(java, ['-jar', bundletoolPath, 'version']);
+  } catch (e) {
+    throw new Error(`Could not get bundle tool version. Original error ${e.message}`);
+  }
+};
+
+/**
+ * Install apks
+ *
+ * @param {string} apks - The full path to the local apks file.
+ * @param {string} deviceId - The device serial to install the apks to.
+ * @param {[string]} modules - List of modules to be installed (defaults to all of them).
+ *                           Note that the dependent modules will also be installed.
+ *                           Ignored if the device receives a standalone APK.
+ * @return {string} Command stdout
+ * @throws {Error} If installation fails. e.g. INSTALL_PARSE_FAILED_NO_CERTIFICATES
+ */
+bundletool.installApks = async function (apks, deviceId, modules = []) {
+  const java = getJavaForOs();
+  const bundletoolPath = path.resolve(this.helperJarPath, 'bundletool.jar');
+  let installArgs = ['-jar', bundletoolPath, 'install-apks', '--apks', apks, '--device-id', deviceId];
+
+  if (!_.isEmpty(modules)) {
+    installArgs.push('--modules', modules.join(','));
+  }
+
+  log.info(`Install apks with: ${installArgs}`);
+  try {
+    return await exec(java, installArgs);
+  } catch (e) {
+    throw new Error(`Failed to install apks. Original error ${e.message}`);
+  }
+};
+
+/**
+ * Check build version
+ *
+ * @param {string} bundle - The full path to the local `aab` file.
+ * @param {string} deviceId - The device serial to install the apks to.
+ * @param {string} output - List of modules to be installed (defaults to all of them).
+ *                           Note that the dependent modules will also be installed.
+ *                           Ignored if the device receives a standalone APK.
+ * @param {object} buildApksOpts - Options
+ * @property {string} ks - Path to the keystore that should be used to sign the
+ *                         generated APKs. If not set, the APKs will not be signed. If set, the
+ *                         flag 'ks-key-alias' must also be set.
+ * @property {string} ksKeyAlias - Alias of the key to use in the keystore to sign the generated APKs.
+ * @property {string} ksPass - Alias of the key to use in the keystore to sign the generated APKs.
+ *                             Password of the keystore to use to sign the generated APKs.
+ *                             If provided, must be prefixed with either 'pass:' (if the password
+ *                             is passed in clear text, e.g. 'pass:qwerty') or 'file:' (if the password
+ *                             is the first line of a file, e.g. 'file:/tmp/myPassword.txt'). If this
+ *                             flag is not set, the password will be requested on the prompt.
+ * @property {string} overwrite If set, any previous existing output will be overwritten.                           --overwrite
+ * @property {string} otherOptions - Add additional option
+ * @return {string} A path to apks
+ * @throws {Error} If building apks fails.
+ */
+bundletool.buildApks = async function (bundle, deviceId, output, buildApksOpts = {}) {
+  const java = getJavaForOs();
+  const bundletoolPath = path.resolve(this.helperJarPath, 'bundletool.jar');
+
+  let buildApksArgs = [
+    '-jar', bundletoolPath, 'build-apks',
+    '--bundle', bundle,
+    '--output', output,
+    '--connected-device', '--device-id', deviceId
+  ];
+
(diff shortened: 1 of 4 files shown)


lib/tools/bundletool.js
```diff
@@ -0,0 +1,114 @@
...
+
+/**
+ * Check bundletool version
+ *
+ * @return {string} Version of bundletool. i.e., BundleTool 0.6.0
+ * @throws {Error} If get version fails
+ */
+bundletool.version = async function () {
+  try {
+    return await exec(getJavaForOs(), [
```
mykola-mokhnach: exec returns an object `{stdout, stderr}`. This is not a string

lib/tools/bundletool.js
```diff
@@ -0,0 +1,114 @@
...
+    throw new Error(`Could not get bundle tool version. Original error ${e.message}`);
+  }
+};
+
+/**
+ * Install apks
+ *
+ * @param {string} apks - The full path to the local apks file.
+ * @param {string} deviceId - The device serial to install the apks to.
+ * @param {[string]} modules - List of modules to be installed (defaults to all of them).
```
mykola-mokhnach: `@param {?string} modules`

lib/tools/bundletool.js
```diff
@@ -0,0 +1,114 @@
...
+    '--apks', apks, '--device-id', deviceId
+  ];
+
+  if (!_.isEmpty(modules)) {
+    installArgs.push('--modules', modules.join(','));
+  }
+
+  log.info(`Install apks with: ${installArgs}`);
+  try {
+    return await exec(getJavaForOs(), installArgs);
```
mykola-mokhnach: the result is not a string

lib/tools/bundletool.js
```diff
@@ -0,0 +1,114 @@
...
+ * @param {string} apks - The full path to the local apks file.
+ * @param {string} deviceId - The device serial to install the apks to.
+ * @param {[string]} modules - List of modules to be installed (defaults to all of them).
+ *                           Note that the dependent modules will also be installed.
+ *                           Ignored if the device receives a standalone APK.
+ * @return {string} Command stdout
+ * @throws {Error} If installation fails. e.g. INSTALL_PARSE_FAILED_NO_CERTIFICATES
+ */
+bundletool.installApks = async function (apks, deviceId, modules = []) {
+  // TODO: Append install options after bundletool implement it.
```
mykola-mokhnach: implements

lib/tools/bundletool.js
```diff
@@ -0,0 +1,114 @@
...
+    throw new Error(`Failed to install apks. Original error ${e.message}`);
+  }
+};
+
+/**
+ * Check build version
+ *
+ * @param {string} bundle - The full path to the local `aab` file.
+ * @param {string} deviceId - The device serial to install the apks to.
+ * @param {string} output - List of modules to be installed (defaults to all of them).
```
mykola-mokhnach: can we have a better name for this param?

lib/tools/bundletool.js
```diff
@@ -0,0 +1,114 @@
...
+ *                           Note that the dependent modules will also be installed.
+ *                           Ignored if the device receives a standalone APK.
+ * @param {object} buildApksOpts - Options
```
mykola-mokhnach: there is `typedef` doctring syntax for defining such things. Please check the other modules for more examples

lib/tools/bundletool.js
```diff
@@ -0,0 +1,114 @@
...
+    return await exec(getJavaForOs(), installArgs);
+  } catch (e) {
+    throw new Error(`Failed to install apks. Original error ${e.message}`);
+  }
+};
+
+/**
+ * Check build version
+ *
+ * @param {string} bundle - The full path to the local `aab` file.
```
mykola-mokhnach: perhaps bundlePath?

lib/tools/bundletool.js
```diff
@@ -0,0 +1,114 @@
...
+  if (buildApksOpts.ksPass) {
+    buildApksArgs.push('--ks-pass', buildApksOpts.ksPass);
+  }
+
+  if (buildApksOpts.overwrite) {
+    buildApksArgs.push('--overwrite');
+  }
+
+  if (buildApksOpts.otherOptions) {
+    buildApksArgs.push(buildApksOpts.otherOptions);
```
mykola-mokhnach: will it work this way? shouldn't otherOptions be an array, which is them merged to buildApksArgs?

lib/tools/bundletool.js
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  log.info(`Build Apks apks with: ${buildApksArgs}`);
+  try {
+    await exec(getJavaForOs(), buildApksArgs);
+    return output;
```
mykola-mokhnach: why do we return the argument value?

lib/tools/index.js
```diff
@@ -4,14 +4,16 @@ import systemCallMethods from './system-calls.js';
...
 
 Object.assign(
     methods,
     manifestMethods,
     systemCallMethods,
     emuMethods,
     apkSigningMethods,
-    apkUtilsMethods
+    apkUtilsMethods,
+    bundletool
```
mykola-mokhnach: I assume bundletool should not have access to `this` context

lib/tools/bundletool.js
```diff
@@ -0,0 +1,114 @@
...
+
+let bundletool = {};
+
+/**
+ * Check bundletool version
+ *
+ * @return {string} Version of bundletool. i.e., BundleTool 0.6.0
+ * @throws {Error} If get version fails
+ */
+bundletool.version = async function () {
```
mykola-mokhnach: I'd rather return the version in semver-compatible format

mykola-mokhnach:
@KazuCocoa isn't bundletool already included into Android SDK?

mykola-mokhnach:
I'm not quite sure about how to make the stuff useful from external modules, which use ADB. In general, the bundletool tool itself is completely separate is not a part of ADB. 

@KazuCocoa Can you please describe an example usage scenario?

KazuCocoa:
No, I couldn't find it in 28.0.3 yet.
ah, I understood it. > PATH

----

I considered installing apk in `.aab` or `.apk` in https://github.com/appium/appium-android-driver/blob/cd485f6af09eea56ba84bb907df5b72a8dfac0e2/lib/commands/app-management.js#L158 or handle them in the `this.adb.install(localPath, options);` (this repository).

`bundletool` allow us to set below option for each command. I thought wrapping the installation, generating `.apks` from `.aab` and install a proper `.apk` to a test device, in this repository is better like `sign.jar`.

```
--adb: (Optional) Path to the adb utility. If absent, an attempt will be
        made to locate it if the ANDROID_HOME environment variable is set. Used
        only if connected-device flag is set.
```

KazuCocoa:
Thanks for all of your insight.

So far, we can't set any flags for install command like 'adb install'. we can't set -g flag, for example. Every time, the installation is equal to vanilla 'adb install', as I read the codebase and observe the behaviour.
Will continue to watch the tool, first.

mykola-mokhnach:
We could probably extract all separate apks from a bundle into tempdir and then apply the same rules to them as we normally apply to each single apk.

KazuCocoa:
Yes, I just imagined like below. (Generating apks in them directory and use it)
https://github.com/appium/appium-adb/compare/master...KazuCocoa:add-bundletool?expand=1#diff-9a49d110e1c61a8dc30a6ac9a8335185R456

KazuCocoa:
📝 
https://appium.slack.com/archives/C02956V3H/p1539558919000100

PR #367, opened 2018-10-16
Introduce apks support

mykola-mokhnach:
This PR continues what was started by @KazuCocoa  in https://github.com/appium/appium-adb/pull/366

the main idea behind the current implementation is to introduce .apks format support without making additional changes in the depending driver modules (aka it should simply work).

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -8,6 +8,7 @@ import semver from 'semver';
 
 const ZIP_MAGIC = 'PK';
 const rootDir = path.resolve(__dirname, process.env.NO_PRECOMPILE ? '..' : '../..');
+const APKS_EXTENSION = '.apks';
 
 /**
  * @typedef {Object} PlatformInfo
@@ -382,5 +383,5 @@ export {
   getIMEListFromOutput, getJavaForOs, isShowingLockscreen, isCurrentFocusOnKeyguard,
   getSurfaceOrientation, isScreenOnFully, buildStartCmd, getJavaHome,
   rootDir, getSdkToolsVersion, getApksignerForOs, getBuildToolsDirs,
-  getApkanalyzerForOs, getOpenSslForOs, extractMatchingPermissions,
+  getApkanalyzerForOs, getOpenSslForOs, extractMatchingPermissions, APKS_EXTENSION,
 };
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -1,10 +1,12 @@
 import path from 'path';
 import { system, fs, zip, util } from 'appium-support';
+import _fs from 'fs';
 import log from './logger.js';
 import { exec } from 'teen_process';
 import _ from 'lodash';
 import B from 'bluebird';
 import semver from 'semver';
+import crypto from 'crypto';
 
 const ZIP_MAGIC = 'PK';
 const rootDir = path.resolve(__dirname, process.env.NO_PRECOMPILE ? '..' : '../..');
@@ -377,10 +379,33 @@ const extractMatchingPermissions = function (dumpsysOutput, groupNames, grantedS
   return filteredResult;
 };
 
+/**
+ * Calculates hash sum for the particular file
+ *
+ * @param {string} filePath The full path to a file.
+ * @param {?string} algorithm [sha256] Any of the hash algorithm names
+ * supported by the crypto library.
+ * @returns {string} The calculated hash as hexadecimal string
+ * @throws {Error} If hash calculation fails
+ */
+async function fileHash (filePath, algorithm = 'sha256') {
+  return await new B((resolve, reject) => {
+    // Algorithm depends on availability of OpenSSL on platform
+    // Another algorithms: 'sha1', 'md5', 'sha256', 'sha512' ...
+    const shasum = crypto.createHash(algorithm);
+    const readStream = _fs.ReadStream(filePath);
+    readStream.on('error', (e) => reject(
+      new Error(`Cannot calculate ${algorithm} hash for '${filePath}'. Original error: ${e.message}`)));
+    readStream.on('data', (data) => shasum.update(data));
+    readStream.on('end', () => resolve(shasum.digest('hex')));
+  });
+}
+
 export {
   getAndroidPlatformAndPath, unzipFile, assertZipArchive,
   getIMEListFromOutput, getJavaForOs, isShowingLockscreen, isCurrentFocusOnKeyguard,
   getSurfaceOrientation, isScreenOnFully, buildStartCmd, getJavaHome,
   rootDir, getSdkToolsVersion, getApksignerForOs, getBuildToolsDirs,
   getApkanalyzerForOs, getOpenSslForOs, extractMatchingPermissions, APKS_EXTENSION,
+  fileHash,
 };
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -11,6 +11,7 @@ import crypto from 'crypto';
 const ZIP_MAGIC = 'PK';
 const rootDir = path.resolve(__dirname, process.env.NO_PRECOMPILE ? '..' : '../..');
 const APKS_EXTENSION = '.apks';
+const APK_EXTENSION = '.apk';
 const APK_INSTALL_TIMEOUT = 60000;
 const APKS_INSTALL_TIMEOUT = APK_INSTALL_TIMEOUT * 2;
 
@@ -454,5 +455,5 @@ export {
   getSurfaceOrientation, isScreenOnFully, buildStartCmd, getJavaHome,
   rootDir, getSdkToolsVersion, getApksignerForOs, getBuildToolsDirs,
   getApkanalyzerForOs, getOpenSslForOs, extractMatchingPermissions, APKS_EXTENSION,
-  fileHash, APK_INSTALL_TIMEOUT, APKS_INSTALL_TIMEOUT, buildInstallArgs,
+  fileHash, APK_INSTALL_TIMEOUT, APKS_INSTALL_TIMEOUT, buildInstallArgs, APK_EXTENSION,
 };
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -61,15 +61,16 @@ async function getAndroidPlatformAndPath () {
   return result;
 }
 
-async function unzipFile (zipPath) {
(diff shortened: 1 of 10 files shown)


lib/helpers.js
```diff
@@ -61,15 +62,16 @@ async function getAndroidPlatformAndPath () {
   return result;
 }
 
-async function unzipFile (zipPath) {
-  log.debug(`Unzipping ${zipPath}`);
+async function unzipFile (zipPath, dstRoot = null) {
+  dstRoot = dstRoot || path.dirname(zipPath);
```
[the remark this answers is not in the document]
mykola-mokhnach: Is it ok to use other arguments in default value calculation?

mykola-mokhnach:
@KazuCocoa Would you like to run some tests on this PR?

lib/helpers.js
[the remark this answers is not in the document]
imurchie: Yes, as long as they are later in the argument list. So
```js
function a (b = c, c) {
  console.log(b, c);
}
```
would fail if you called `a(undefined, 1)`, with a reference error, because `c` would not have been defined when `b` is initialized to it.

But, for every situation, the follow is fine:
```js
function a (b, c = b) {
  console.log(b, c);
}
```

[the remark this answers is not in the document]
mykola-mokhnach: ok, fixed

lib/tools/apk-utils.js
```diff
@@ -393,20 +391,24 @@ apkUtilsMethods.install = async function (apk, options = {}) {
...
-      log.debug(`Skipping granting permissions for '${apk}', since ` +
+      log.debug(`Skipping granting permissions for '${appPath}', since ` +
                 `the current API level ${apiLevel} does not support applications ` +
                 `permissions customization`);
     } else {
       additionalArgs.push('-g');
     }
   }
 
+  if (appPath.endsWith(APKS_EXTENSION)) {
```
The installation depends on device environment. Configuring device env, including language preference, is called in first step in `startAndroidSession`. The bundle-install command doesn't install unused languages resources.
https://github.com/appium/appium-android-driver/blob/2bac1672019be24c74fec7b4b5e4301558bfcb66/lib/driver.js#L234

mykola-mokhnach: in theory we could additionally install these language resources manually or change the setup flow sequence. Which way does look more convenient to you?

mykola-mokhnach: @KazuCocoa I've changed the way we install apks. Now it is much more customizable.

lib/helpers.js
```diff
@@ -376,10 +381,78 @@ const extractMatchingPermissions = function (dumpsysOutput, groupNames, grantedS
...
+/**
+ * Calculates hash sum for the particular file
+ *
+ * @param {string} filePath The full path to a file.
+ * @param {?string} algorithm [sha1] Any of the hash algorithm names
+ * supported by the crypto library.
+ * @returns {string} The calculated hash as hexadecimal string
+ * @throws {Error} If hash calculation fails
+ */
+async function fileHash (filePath, algorithm = 'sha1') {
```
jlipps: feels like this should be in appium-support?

lib/tools/apk-utils.js
```diff
@@ -368,62 +365,38 @@ const APK_INSTALL_TIMEOUT = 60000;
...
-    const apiLevel = await this.getApiLevel();
-    if (apiLevel < 23) {
-      log.debug(`Skipping granting permissions for '${apk}', since ` +
-                `the current API level ${apiLevel} does not support applications ` +
-                `permissions customization`);
-    } else {
-      additionalArgs.push('-g');
-    }
-  }
+  options = Object.assign({
```
jlipps: `Object.assign` mutates, dunno if that changes how you'd want to use it here

lib/tools/apks-utils.js
```diff
@@ -0,0 +1,210 @@
...
+
+/**
+ * Extracts the particular apks package into a temporary folder,
+ * finds and returns the full path to the file contained in this apk.
+ * The resulting temporary path, where the .apks file has been extracted,
+ * will be stored into the internal LRU cache for better performance.
+ *
+ * @param {string} apks - The full path to the .apks file
+ * @param {string|Array<String>} dstPath - The relative path to the destination file,
+ * which is going to be extract3ed, where each path component is an array item
```
jlipps: extracted

lib/tools/apks-utils.js
```diff
@@ -0,0 +1,210 @@
...
+      '--adb', this.executable.path,
+      '--apks', apks,
+      '--output-dir', tmpRoot,
+      '--device-spec', specPath,
+    ];
+    log.debug(`Extracting the apk files from '${apks}'`);
+    await this.execBundletool(args, `Cannot extract the application bundle at '${apks}'`);
+    const installArgs = buildInstallArgs(await this.getApiLevel(), options);
+    const apksToInstall = (await fs.readdir(tmpRoot))
+      .filter((name) => name.endsWith('.apk'))
```
KazuCocoa: [nits]
Is it better to define `.apk` const like APKS_EXTENSION?

lib/tools/apks-utils.js
```diff
@@ -0,0 +1,210 @@
...
+      .map((name) => path.resolve(tmpRoot, name));
+    log.debug(`Got the following apk files to install: ${JSON.stringify(apksToInstall)}`);
+    const output = await this.adbExec(['install-multiple', ...installArgs, ...apksToInstall], {
```
KazuCocoa: If the install apks has unsuitable architecture, installing apks fail. (`adb install` failed before. It happen if users use ML related libraries, for example.)

```
base-vi.apk 11K
base-x86.apk 483K
base-x86_64.apk 453K
base-xhdpi.apk 2.4M
```

~~What about add considering abis?~~
=> ah, we use `'--device-spec', specPath,`. nice 👍

lib/tools/adb-commands.js
```diff
@@ -57,6 +57,20 @@ methods.initZipAlign = async function () {
   this.binaries.zipalign = await this.getSdkBinaryPath("zipalign");
 };
 
+/**
+ * Get the full path to bundletool binary and assign it to
+ * this.binaries.bundletool property
+ */
+methods.initBundletool = async function () {
+  try {
+    const bundletoolPath = await fs.which('bundletool.jar');
```
KazuCocoa: We should export the path to 'bundletool.jar', right? 👀

lib/tools/apks-utils.js
```diff
@@ -0,0 +1,213 @@
...
+apksUtilsMethods.installApks = async function (apks, options = {}) {
+  options = Object.assign({
+    timeout: APKS_INSTALL_TIMEOUT,
+  }, options, {replace: true});
+
+  const tmpRoot = await tempDir.openDir();
+  try {
+    const specPath = await this.getDeviceSpec(path.resolve(tmpRoot, 'deviceSpec.json'));
+    const args = [
+      'extract-apks',
```
KazuCocoa: Couldn't call `--adb` here ✔️ 

```
Description:
    Extracts from an APK Set the APKs that should be installed on a given
    device.

Synopsis:
    bundletool extract-apks
        --apks=<archive.apks>
        --device-spec=<device-spec.json>
        --output-dir=<output-dir>
        [--instant]
        [--modules=<base,module1,module2>]

Flags:
    --apks: Path to the archive file generated by the 'build-apks' command.

    --device-spec: Path to the device spec file generated by the 'get-device-
        spec' command.

    --output-dir: Path to where the matched APKs will be extracted from the
        archive file.

    --instant: (Optional) When set, APKs of the instant modules will be
        extracted instead of the installable APKs.

    --modules: (Optional) List of modules to be extracted (defaults to all of
        them). Note that the dependent modules will also be extracted. Ignored
        if the device receives a standalone APK.
```

KazuCocoa:
I've implemented https://github.com/KazuCocoa/AppBundleSample/blob/master/test.rb and run it to make sure the behaviour. Then, I noticed below three points.


1. If a test app(only english resource) already has been installed, it remains. Then, 2nd test scenario failed since 2nd scenario would like to run tests against Japanese resource
    - https://github.com/KazuCocoa/AppBundleSample/blob/master/test.rb
        1. Can we make sure the installed apk's resource and try re-install if we need?
        2. Or use `standalones/standalone-xxxhdpi.apk`?
        3. (Or we'll guide full reset for `.apks` at first?)
            - If we can do 1st or 2nd way, it's better, I think. 3 is reasonable so far, I also think since buildtools also isn't 1.x yet.
2. We need to relax `APP_EXTENSION` for `configureApp` to allow `.apks` extension. We need to extend `configureApp` to allow multiple extension.
    - https://github.com/search?p=2&q=org%3Aappium+configureApp&type=Code
        - I can work them when I have a time, btw.
3. https://github.com/appium/appium-adb/pull/367/files#r226311748

lib/tools/apks-utils.js
mykola-mokhnach: thanks, will fix

KazuCocoa:
> Btw, does it help if fullReset is enabled?

Yes, but it's the final way, I believe.
If we can't do anything to identify when Appium must re-install an apk, it's helpful. But if we enable `fullReset` everytime, below case happens. (Your heavy improvement for creating session doesn't work this case.)

1. Create a session with English resource
2. Create a session with English resource as another test case <= We don't need to re-install, but re-install happens if we enable fullReset

mykola-mokhnach:
I would just assume this is the main purpose of bundling, that we only install the necessary resources. So it would be logically to assume we need to reinstall the app in order to get the new locale from the bundle (real users would also have to do the same in case of apks). Also, sometimes full reset might be even faster than soft reset, since it does not require to set permissions on the package (these are set during installation)

KazuCocoa:
> sometimes full reset might be even faster than soft reset, since it does not require to set permissions on the package (these are set during installation)

ah, it's reasonable to enable fullReset for `.apks`. I agree with the trying fullReset way.

lib/tools/apk-utils.js
jlipps: oh right, only the target is mutated and in this case it's an object literal

mykola-mokhnach:
@KazuCocoa Are you fine with merging this PR in the current state or you'd like to add something?

KazuCocoa:
Let me run test again once with fullRest cap

PR #373, opened 2018-11-02
Set  adbExecTimeout by default as DEFAULT_ADB_EXEC_TIMEOUT

KazuCocoa:
android-driver create `adb` with `adbExecTimeout`. Nothing `execTimeout` in appium repository except for this repository. So, I assume the option should be `adbExecTimeout`.

https://github.com/appium/appium-android-driver/search?q=adbExecTimeout&unscoped_q=adbExecTimeout

---

update

in a602d24 , I set `adbExecTimeout: DEFAULT_ADB_EXEC_TIMEOUT` by the default


mykola-mokhnach:
There's something wrong with the tests

lib/adb.js
```diff
@@ -3,6 +3,7 @@ import os from 'os';
 import path from 'path';
 import methods from './tools/index.js';
 import { rootDir } from './helpers';
+import DEFAULT_ADB_EXEC_TIMEOUT from './tools/system-calls';
```
mykola-mokhnach: most likely it should be imported being enclosed in curly brackets:

`import { DEFAULT_ADB_EXEC_TIMEOUT } from './tools/system-calls';`

mykola-mokhnach:
I would be happy if we could also standardise methods like `availableIMEs`, where the timeout value is hardcoded to 60000ms. This has been done in order to make Travis happy, because the emulation there is slow. With the current approach we could set the global timeout once for the adb instance used in e2e tests and avoid hardcoding it for separate method calls.

If you don't have time to do this change in scope of this PR @KazuCocoa then I would be happy to add the change in the next one.

KazuCocoa:
I see. Will do in this PR 👍 

```
$ git grep "60000"
.travis.yml:    - MOCHA_TIMEOUT=360000
lib/helpers.js:const APK_INSTALL_TIMEOUT = 60000;
lib/tools/adb-commands.js:      timeout: 60000,
lib/tools/adb-commands.js:      timeout: 60000,
lib/tools/adb-commands.js:  await this.adbExec(['pull', remotePath, localPath], {timeout: 60000});
lib/tools/adb-commands.js:    timeout: 60000,
lib/tools/apk-utils.js: * @property {number} timeout [60000] - The count of milliseconds to wait until the
lib/tools/apk-utils.js: * @property {number} timeout [60000] - The count of milliseconds to wait until the
lib/tools/system-calls.js: * @param {?number} timeout [60000] - The amount of time to wait before throwing
lib/tools/system-calls.js:systemCallMethods.killEmulator = async function (avdName = null, timeout = 60000) {
lib/tools/system-calls.js: * @param {number} avdLaunchTimeout [60000] - Emulator startup timeout in milliseconds.
lib/tools/system-calls.js:  avdLaunchTimeout = 60000, avdReadyTimeout = 60000, retryTimes = 1) {
test/functional/apk-utils-e2e-specs.js:const waitDuration = 60000;
test/functional/apk-utils-e2e-specs.js:        waitDuration: 60000,
test/functional/setup.js:let MOCHA_TIMEOUT = process.env.TRAVIS ? 240000 : 60000;
test/unit/apk-utils-specs.js:        .once().withExactArgs(['install', '-r', 'foo'], {timeout: 60000})
test/unit/apk-utils-specs.js:        .once().withExactArgs(['install', 'foo'], {timeout: 60000})
test/unit/apk-utils-specs.js:      mocks.adb.expects('install').withArgs(apkPath, {replace: true, timeout: 60000}).once().returns(true);
test/unit/apk-utils-specs.js:      mocks.adb.expects('install').withArgs(apkPath, {replace: true, timeout: 60000}).once().returns(true);
test/unit/apk-utils-specs.js:      mocks.adb.expects('install').withArgs(apkPath, {replace: true, timeout: 60000}).once().returns(true);
test/unit/apk-utils-specs.js:      mocks.adb.expects('install').withArgs(apkPath, {replace: true, timeout: 60000}).once().throws();
test/unit/apk-utils-specs.js:      mocks.adb.expects('install').withArgs(apkPath, {replace: false, timeout: 60000}).once().returns(true);
```

test/unit/adb-commands-specs.js
```diff
@@ -35,7 +35,7 @@ u0_a101   5078  3129  487404 37044 ffffffff b76ce565 S com.example.android.conta
       manufacturer = `unknown`,
       screenSize = `768x1280`;
 
-const adb = new ADB();
+const adb = new ADB({ adbExecTimeout: 60000 });
```
mykola-mokhnach: this is also needed for adb-commands-e2e-specs

KazuCocoa:
thanks!

PR #376, opened 2018-11-06
Fix reboot for Android O

vrunoa:
In Android O emulators reboot it's failing due to root issues. 

```bash
Error: Error executing adbExec. Original error: 'Command '/mnt/android/android-sdk-linux/platform-tools/adb -P 5037 shell start' exited with code 1'; Stderr: 'start: must be root'; Code: '1'
    Error: Error executing adbExec. Original error: 'Command '/mnt/android/android-sdk-linux/platform-tools/adb -P 5037 shell start' exited with code 1'; Stderr: 'start: must be root'; Code: '1'
2018/11/06 16:43:53 ui:     at ADB.execFunc$ (/tmp/chromeInstaller/node_modules/appium-adb/lib/tools/system-calls.js:317:13)
 ```

Adding root before calling shell starts fixes the issue.

### DIFF
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -787,7 +787,7 @@ systemCallMethods.reboot = async function (retries = DEFAULT_ADB_REBOOT_RETRIES)
       }
       // this device needs adb to be running as root to stop.
       // so try to restart the daemon
-      log.debug('Device requires adb to be running as root in order to reboot. Restarting daemon');
+      log.debug('Device requires adb to be running as root in order to reboot when running stop command. Restarting daemon');
       await this.root();
       await this.shell(['stop']);
     }
@@ -801,7 +801,7 @@ systemCallMethods.reboot = async function (retries = DEFAULT_ADB_REBOOT_RETRIES)
       }
       // this device needs adb to be running as root to start.
       // so try to restart the daemon
-      log.debug('Device requires adb to be running as root in order to reboot. Restarting daemon');
+      log.debug('Device requires adb to be running as root in order to reboot when running start command. Restarting daemon');
       await this.root();
       await this.shell(['start']);
     }
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -779,6 +779,7 @@ systemCallMethods.waitForDevice = async function (appDeviceReadyTimeout = 30) {
  */
 systemCallMethods.runPrivilegedShell = async function (args) {
   try {
+    this.isRoot = false;
     await this.shell(args);
   } catch (err) {
     if (!err.message.includes('must be root')) {
@@ -787,10 +788,8 @@ systemCallMethods.runPrivilegedShell = async function (args) {
     // this device needs adb to be running as root to run the specific command.
     // so try to restart the daemon
     log.debug(`Device requires adb to be running as root in order to run "adb shell ${quote(args)}". Restarting daemon`);
-    await this.root();
+    this.isRoot = await this.root();
     await this.shell(args);
-  } finally {
-    await this.unroot();
   }
 };
 
@@ -813,7 +812,12 @@ systemCallMethods.reboot = async function (retries = DEFAULT_ADB_REBOOT_RETRIES)
     // we don't want the stack trace, so no log.errorAndThrow
     let msg = 'Waiting for reboot. This takes time';
     log.debug(msg);
+    throw new Error(msg);
   });
+  // unroot if needed
+  if (this.isRoot) {
+    this.isRoot = !await this.unroot();
+  }
 };
 
 /**
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -813,7 +813,6 @@ systemCallMethods.reboot = async function (retries = DEFAULT_ADB_REBOOT_RETRIES)
     // we don't want the stack trace, so no log.errorAndThrow
     let msg = 'Waiting for reboot. This takes time';
     log.debug(msg);
-    throw new Error(msg);
   });
 };
 
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -793,8 +793,18 @@ systemCallMethods.reboot = async function (retries = DEFAULT_ADB_REBOOT_RETRIES)
     }
     await B.delay(2000); // let the emu finish stopping;
     await this.setDeviceProperty('sys.boot_completed', 0);
-    await this.root();
-    await this.shell(['start']);
+    try {
+      await this.shell(['start']);
+    } catch (err) {
+      if (err.message.indexOf('must be root') === -1) {
+        throw err;
+      }
+      // this device needs adb to be running as root to start.
+      // so try to restart the daemon
+      log.debug('Device requires adb to be running as root in order to reboot. Restarting daemon');
+      await this.root();
+      await this.shell(['start']);
+    }
     await retryInterval(retries, 1000, async () => {
       let booted = await this.getDeviceProperty('sys.boot_completed');
       if (booted === '1') {
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -776,7 +776,7 @@ systemCallMethods.privilegedShell = async function (args) {
   try {
     await this.shell(args);
   } catch (err) {
-    if (err.message.includes('must be root') === -1) {
+    if (err.message.includes('must be root')) {
       throw err;
     }
     // this device needs adb to be running as root to run the specific command.
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -804,6 +804,7 @@ systemCallMethods.runPrivilegedShell = async function (args) {
 systemCallMethods.reboot = async function (retries = DEFAULT_ADB_REBOOT_RETRIES) {
   let isRoot = false;
   isRoot = await this.runPrivilegedShell(['stop']);
+  // TODO - Figure out a condition in adb so we can replace this static delay with a waitForCondition
   await B.delay(EMU_STOP_TIMEOUT); // let the emu finish stopping;
   await this.setDeviceProperty('sys.boot_completed', 0);
   isRoot = await this.runPrivilegedShell(['start']);
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -806,7 +806,7 @@ systemCallMethods.reboot = async function (retries = DEFAULT_ADB_REBOOT_RETRIES)
   isRoot = await this.runPrivilegedShell(['stop']);
   // TODO - Figure out a condition in adb so we can replace this static delay with a waitForCondition
   await B.delay(EMU_STOP_TIMEOUT); // let the emu finish stopping;
-  await this.setDeviceProperty('sys.boot_completed', 0);
+  isRoot = await this.runPrivilegedShell(['setprop', 'sys.boot_completed', 0]);
   isRoot = await this.runPrivilegedShell(['start']);
   await retryInterval(retries, 1000, async () => {
     if ((await this.getDeviceProperty('sys.boot_completed')) === '1') {
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -801,10 +801,10 @@ systemCallMethods.runPrivilegedShell = async function (args) {
  * @throws {Error} If the device failed to reboot and number of retries is exceeded.
  */
 systemCallMethods.reboot = async function (retries = DEFAULT_ADB_REBOOT_RETRIES) {
-  await this.privilegedShell(['stop']);
+  await this.runPrivilegedShell(['stop']);
   await B.delay(EMU_STOP_TIMEOUT); // let the emu finish stopping;
   await this.setDeviceProperty('sys.boot_completed', 0);
-  await this.privilegedShell(['start']);
+  await this.runPrivilegedShell(['start']);
   await retryInterval(retries, 1000, async () => {
     let booted = await this.getDeviceProperty('sys.boot_completed');
     if (booted === '1') {
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -803,12 +803,11 @@ systemCallMethods.reboot = async function (retries = DEFAULT_ADB_REBOOT_RETRIES)
       let booted = await this.getDeviceProperty('sys.boot_completed');
       if (booted === '1') {
         return;
-      } else {
-        // we don't want the stack trace, so no log.errorAndThrow
-        let msg = 'Waiting for reboot. This takes time';
-        log.debug(msg);
-        throw new Error(msg);
       }
+      // we don't want the stack trace, so no log.errorAndThrow
+      let msg = 'Waiting for reboot. This takes time';
+      log.debug(msg);
+      throw new Error(msg);
     });
   } finally {
     await this.unroot();
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -814,7 +814,6 @@ systemCallMethods.reboot = async function (retries = DEFAULT_ADB_REBOOT_RETRIES)
     log.debug(msg);
     throw new Error(msg);
   });
-  // unroot if needed
   if (this.isRoot) {
     this.isRoot = !await this.unroot();
   }
@@ -850,6 +849,7 @@ systemCallMethods.root = async function () {
  */
 systemCallMethods.unroot = async function () {
   try {
+    log.debug("Removing root privileges. Restarting adb daemon");
     await exec(this.executable.path, ['unroot']);
     return true;
   } catch (err) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: lib/tools/system-calls.js [hunk 1]
```diff
@@ -793,6 +793,7 @@ systemCallMethods.reboot = async function (retries = DEFAULT_ADB_REBOOT_RETRIES)
     }
     await B.delay(2000); // let the emu finish stopping;
     await this.setDeviceProperty('sys.boot_completed', 0);
+    await this.root();
```
mykola-mokhnach: The design of this method is messy. Please change it so the flow looks like
- check if root and assign the result to `isRoot`
- if `!isRoot` then elevate to root
- execute all further actions in `try` block
- in `finally` block check if the user was root
- execute `unroot` if the user was not root before the method has been entered

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "appium-adb",
-  "version": "6.25.0",
+  "version": "6.25.1",
```
mykola-mokhnach: the module version should not be set manually. It is bumped as a part of publishing process

lib/tools/system-calls.js
vrunoa: @mykola-mokhnach the problem is the emulator losing root after stops so we need to run it again.

package.json
vrunoa: Wops! Fixed

lib/tools/system-calls.js
```diff
@@ -787,13 +787,24 @@ systemCallMethods.reboot = async function (retries = DEFAULT_ADB_REBOOT_RETRIES)
...
       await this.root();
       await this.shell(['stop']);
     }
     await B.delay(2000); // let the emu finish stopping;
     await this.setDeviceProperty('sys.boot_completed', 0);
-    await this.shell(['start']);
+    try {
+      await this.shell(['start']);
+    } catch (err) {
+      if (err.message.indexOf('must be root') === -1) {
```
KazuCocoa: What about using `.includes` instead of `indexOf`?

lib/tools/system-calls.js
```diff
@@ -787,13 +787,24 @@ systemCallMethods.reboot = async function (retries = DEFAULT_ADB_REBOOT_RETRIES)
...
       // so try to restart the daemon
-      log.debug('Device requires adb to be running as root in order to reboot. Restarting daemon');
+      log.debug('Device requires adb to be running as root in order to reboot when running stop command. Restarting daemon');
       await this.root();
       await this.shell(['stop']);
     }
     await B.delay(2000); // let the emu finish stopping;
     await this.setDeviceProperty('sys.boot_completed', 0);
-    await this.shell(['start']);
+    try {
```
mykola-mokhnach: I see code duplication here? Can we instead have some wrapper method for this purpose?

lib/tools/system-calls.js
```diff
@@ -787,13 +787,24 @@ systemCallMethods.reboot = async function (retries = DEFAULT_ADB_REBOOT_RETRIES)
       }
       // this device needs adb to be running as root to stop.
       // so try to restart the daemon
-      log.debug('Device requires adb to be running as root in order to reboot. Restarting daemon');
+      log.debug('Device requires adb to be running as root in order to reboot when running stop command. Restarting daemon');
       await this.root();
       await this.shell(['stop']);
     }
     await B.delay(2000); // let the emu finish stopping;
```
mykola-mokhnach: can hardcoded delay be replaced with a timer?

lib/tools/system-calls.js
```diff
@@ -787,13 +787,24 @@ systemCallMethods.reboot = async function (retries = DEFAULT_ADB_REBOOT_RETRIES)
...
+      }
+      // this device needs adb to be running as root to start.
+      // so try to restart the daemon
+      log.debug('Device requires adb to be running as root in order to reboot when running start command. Restarting daemon');
+      await this.root();
+      await this.shell(['start']);
+    }
     await retryInterval(retries, 1000, async () => {
       let booted = await this.getDeviceProperty('sys.boot_completed');
       if (booted === '1') {
```
mykola-mokhnach: `else` is not needed if there is an return before

lib/tools/system-calls.js
vrunoa: @mykola-mokhnach I added the privilegedShell method, let me know what you think

lib/tools/system-calls.js
vrunoa: added the EMU_STOP_TIMEOUT const

lib/tools/system-calls.js
```diff
@@ -771,6 +772,21 @@ systemCallMethods.waitForDevice = async function (appDeviceReadyTimeout = 30) {
   });
 };
 
+systemCallMethods.privilegedShell = async function (args) {
+  try {
+    await this.shell(args);
+  } catch (err) {
+    if (err.message.includes('must be root') === -1) {
```
mykola-mokhnach: includes returns boolean result

lib/tools/system-calls.js
```diff
@@ -771,6 +772,21 @@ systemCallMethods.waitForDevice = async function (appDeviceReadyTimeout = 30) {
   });
 };
 
+systemCallMethods.privilegedShell = async function (args) {
```
mykola-mokhnach: function names must include verbs

mykola-mokhnach: docstring is missing

lib/tools/system-calls.js
```diff
@@ -771,6 +772,21 @@ systemCallMethods.waitForDevice = async function (appDeviceReadyTimeout = 30) {
...
+  try {
+    await this.shell(args);
+  } catch (err) {
+    if (err.message.includes('must be root') === -1) {
+      throw err;
+    }
+    // this device needs adb to be running as root to run the specific command.
+    // so try to restart the daemon
+    log.debug(`Device requires adb to be running as root in order to run "adb shell ${args.join(",")}". Restarting daemon`);
```
mykola-mokhnach: you could use shell-quote lib in order to properly print args

lib/tools/system-calls.js
```diff
@@ -771,6 +772,21 @@ systemCallMethods.waitForDevice = async function (appDeviceReadyTimeout = 30) {
...
+    await this.root();
+    await this.shell(args);
+  }
```
mykola-mokhnach: how about unrooting in final block?

lib/tools/system-calls.js
```diff
@@ -779,31 +795,19 @@ systemCallMethods.waitForDevice = async function (appDeviceReadyTimeout = 30) {
...
-      }
-      // this device needs adb to be running as root to stop.
-      // so try to restart the daemon
-      log.debug('Device requires adb to be running as root in order to reboot. Restarting daemon');
-      await this.root();
-      await this.shell(['stop']);
-    }
-    await B.delay(2000); // let the emu finish stopping;
+    await this.privilegedShell(['stop']);
+    await B.delay(EMU_STOP_TIMEOUT); // let the emu finish stopping;
```
mykola-mokhnach: so, cannot we have a timer here?

lib/tools/system-calls.js
dpgraham:
```javascript
import { quote } from 'shell-quote';
...

`"adb shell ${quote(args)}". Restarting daemon`
```

lib/tools/system-calls.js
```diff
@@ -779,31 +795,19 @@ systemCallMethods.waitForDevice = async function (appDeviceReadyTimeout = 30) {
...
-        log.debug(msg);
-        throw new Error(msg);
       }
+      // we don't want the stack trace, so no log.errorAndThrow
+      let msg = 'Waiting for reboot. This takes time';
+      log.debug(msg);
+      throw new Error(msg);
     });
   } finally {
     await this.unroot();
```
mykola-mokhnach: this would probably be not needed anymore if unroot is done inside the helper method

lib/tools/system-calls.js
dpgraham: @mykola-mokhnach So you mean to use `setTimeout` instead of `delay`?

lib/tools/system-calls.js
dpgraham:
```javascript
if (err.message.includes('must be root')) {
   ...
}
```

lib/tools/system-calls.js
jlipps: `B.delay` is a timer, isn't it?

lib/tools/system-calls.js
jlipps: maybe `elevateShellPrivileges` or similar?

lib/tools/system-calls.js
vrunoa: Im confused about this too 😄

lib/tools/system-calls.js
vrunoa: `runPrivilegedShell` ?

lib/tools/system-calls.js
vrunoa: @mykola-mokhnach removed the `unroot` from this method since trying to unroot during a restart makes the code to fail since the device goes offline for a bit.

lib/tools/system-calls.js
mykola-mokhnach: I mean something that check for the expected results and waits if the result is not there or exists immediately, like the `retry` or `waitForCondition` functions, instead of the hardcoded delay

lib/tools/system-calls.js
```diff
@@ -771,42 +772,50 @@ systemCallMethods.waitForDevice = async function (appDeviceReadyTimeout = 30) {
...
 };
 
+/**
+ * Tries to run privileged shell commands without root, otherwise elevates privileges and run them again
+ *
+ * @param {args} array - Shell arguments commands
+ */
+systemCallMethods.runPrivilegedShell = async function (args) {
+  try {
+    this.isRoot = false;
```
mykola-mokhnach: I'd rather simply return the result from the function. Using instance properties for this purpose creates too many unnecessary dependencies.

lib/tools/system-calls.js
```diff
@@ -771,42 +772,50 @@ systemCallMethods.waitForDevice = async function (appDeviceReadyTimeout = 30) {
...
-      // so try to restart the daemon
-      log.debug('Device requires adb to be running as root in order to reboot. Restarting daemon');
-      await this.root();
-      await this.shell(['stop']);
+  await this.runPrivilegedShell(['stop']);
+  await B.delay(EMU_STOP_TIMEOUT); // let the emu finish stopping;
+  await this.setDeviceProperty('sys.boot_completed', 0);
+  await this.runPrivilegedShell(['start']);
+  await retryInterval(retries, 1000, async () => {
+    let booted = await this.getDeviceProperty('sys.boot_completed');
```
mykola-mokhnach: this intermediate variable seems to be redundant

lib/tools/system-calls.js
```diff
@@ -771,42 +772,50 @@ systemCallMethods.waitForDevice = async function (appDeviceReadyTimeout = 30) {
   });
 };
 
+/**
+ * Tries to run privileged shell commands without root, otherwise elevates privileges and run them again
+ *
+ * @param {args} array - Shell arguments commands
```
mykola-mokhnach: the type is `Array<string>`

lib/tools/system-calls.js
```diff
@@ -771,42 +772,50 @@ systemCallMethods.waitForDevice = async function (appDeviceReadyTimeout = 30) {
   });
 };
 
+/**
+ * Tries to run privileged shell commands without root, otherwise elevates privileges and run them again
```
mykola-mokhnach: actually, only one command with arguments

lib/tools/system-calls.js
vrunoa: @mykola-mokhnach the reboot is going to wait, right now I don't know what condition we could be using for checking stop finished, but I feel is out of the scope of this PR.

lib/tools/system-calls.js
```diff
@@ -771,41 +772,50 @@ systemCallMethods.waitForDevice = async function (appDeviceReadyTimeout = 30) {
   });
 };
 
+/**
+ * Tries to run privileged shell command without root, otherwise elevates privileges and run them again
+ *
+ * @param {args} Array<string> - Shell arguments commands
```
mykola-mokhnach: missing `return` description

lib/tools/system-calls.js
```diff
@@ -771,41 +772,50 @@ systemCallMethods.waitForDevice = async function (appDeviceReadyTimeout = 30) {
...
-      if (err.message.indexOf('must be root') === -1) {
-        throw err;
-      }
-      // this device needs adb to be running as root to stop.
-      // so try to restart the daemon
-      log.debug('Device requires adb to be running as root in order to reboot. Restarting daemon');
-      await this.root();
-      await this.shell(['stop']);
+  let isRoot = false;
+  isRoot = await this.runPrivilegedShell(['stop']);
```
mykola-mokhnach: `isRoot` can be defined right here

lib/tools/system-calls.js
```diff
@@ -771,41 +772,50 @@ systemCallMethods.waitForDevice = async function (appDeviceReadyTimeout = 30) {
...
-    } catch (err) {
-      if (err.message.indexOf('must be root') === -1) {
-        throw err;
-      }
-      // this device needs adb to be running as root to stop.
-      // so try to restart the daemon
-      log.debug('Device requires adb to be running as root in order to reboot. Restarting daemon');
-      await this.root();
-      await this.shell(['stop']);
+  let isRoot = false;
```
mykola-mokhnach: try/finally is still needed in case an exception is thrown somewhere in the middle of the method

lib/tools/system-calls.js
```diff
@@ -771,41 +772,50 @@ systemCallMethods.waitForDevice = async function (appDeviceReadyTimeout = 30) {
...
+  isRoot = await this.runPrivilegedShell(['stop']);
+  await B.delay(EMU_STOP_TIMEOUT); // let the emu finish stopping;
```
mykola-mokhnach: it might be handy to leave TODO here

lib/tools/system-calls.js
vrunoa: @mykola-mokhnach what should we set in the finally? We cannot unroot since this will raise exceptions in some cases.

mykola-mokhnach: calling unroot is safe:

```
systemCallMethods.unroot = async function () {
  try {
    await exec(this.executable.path, ['unroot']);
    return true;
  } catch (err) {
    log.warn(`Unable to unroot adb daemon: '${err.message}'. Continuing`);
    return false;
  }
};
```

lib/tools/system-calls.js
vrunoa: Are we good with this naming?

lib/tools/system-calls.js
```diff
@@ -771,41 +772,52 @@ systemCallMethods.waitForDevice = async function (appDeviceReadyTimeout = 30) {
...
+  return isRoot;
+};
+
 /**
  * Reboot the current device and wait until it is completed.
  *
  * @param {number} retries [DEFAULT_ADB_REBOOT_RETRIES] - The maximum number of reboot retries.
  * @throws {Error} If the device failed to reboot and number of retries is exceeded.
  */
 systemCallMethods.reboot = async function (retries = DEFAULT_ADB_REBOOT_RETRIES) {
```
mykola-mokhnach: any reason we cannot use `reboot` for rebooting the device? 

https://blog.csdn.net/johnnylq/article/details/6401531

lib/tools/system-calls.js
```diff
@@ -771,41 +772,52 @@ systemCallMethods.waitForDevice = async function (appDeviceReadyTimeout = 30) {
...
-      // this device needs adb to be running as root to stop.
-      // so try to restart the daemon
-      log.debug('Device requires adb to be running as root in order to reboot. Restarting daemon');
-      await this.root();
-      await this.shell(['stop']);
+  let isRoot = false;
+  isRoot = await this.runPrivilegedShell(['stop']);
+  // TODO - Figure out a condition in adb so we can replace this static delay with a waitForCondition
+  await B.delay(EMU_STOP_TIMEOUT); // let the emu finish stopping;
+  isRoot = await this.runPrivilegedShell(['setprop', 'sys.boot_completed', 0]);
```
mykola-mokhnach: i'd rather change the equal sign to `|=`

mykola-mokhnach:
@vrunoa I think we could nicely incorporate `shell` implementation from https://github.com/appium/appium-adb/pull/378 into this PR to reduce the LOC amount

lib/tools/system-calls.js
vrunoa: It fails if the device goes offline for a couple of seconds.

vrunoa:
@mykola-mokhnach @dpgraham  if possible I would like to move forward and merge this as it is. I probably wont have time to merge both PR into one this week.

mykola-mokhnach:
No problems @vrunoa. I could include the necessary changes into https://github.com/appium/appium-adb/pull/378
simply find some time then to review it

PR #378, opened 2018-11-10
Add an option to shell command for running as root

mykola-mokhnach:
I observe a need to in such shortcut for commands that can only be executed under root privileges.

### DIFF
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -332,18 +332,56 @@ systemCallMethods.adbExec = async function (cmd, opts = {}) {
 };
 
 /**
+ * @typedef {Object} ShellExecOptions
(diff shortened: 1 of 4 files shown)


lib/tools/system-calls.js
```diff
@@ -861,6 +873,17 @@ systemCallMethods.unroot = async function () {
...
 
+/**
+ * Checks whether the current user is root
+ *
+ * @return {boolean} True if the user is root
+ * @throws {Error} if there was an error while identifying
+ * the user.
+ */
+systemCallMethods.isRoot = async function () {
+  return (await this.shell(['whoami'])).trim() === 'root';
```
dpgraham: Could `await this.shell(['whoami'])` ever return 'root' in upper case?

mykola-mokhnach: I don't think so

PR #381, opened 2018-11-14
add grant multiple permissions in one time adb shell call

KazuCocoa:
This PR should help https://github.com/appium/appium-android-driver/pull/465 to reduce the cost of calling `adb shell` multiple times.
To call all of commands in a target device's shell, we must send commands in `"...."`


mykola-mokhnach:
Thanks for that, but we anyway had to do some refactoring there. Closing in favour of https://github.com/appium/appium-adb/pull/382

PR #383, opened 2018-11-14
Correct comment for restartAdb

clample:
Currently the code comment for `restartAdb` is incorrect. It says that `restartAdb` restarts the adb server _if_ `suppressKillServer` is true. The actual behavior is the opposite, though. `restartAdb` restarts the adb server _unless_ `suppressKillServer` is true. This PR simply changes the comment.


clample:
Thanks for the review @mykola-mokhnach 

Just a heads up, it looks like I don't have access to merge this. Is someone else able to do that?

PR #384, opened 2018-11-19
Workaround NodeJS bug while executing scripts on Windows

mykola-mokhnach:
Hopefully addresses https://github.com/appium/appium/issues/11704


mykola-mokhnach:
Please do not merge until the fix is confirmed

PR #385, opened 2018-11-20
Only patch apksigner path on Windows

mykola-mokhnach:
It looks like on *nix systems the workaround does not work as expected and the binary is not found even if `cwd` is set. See https://travis-ci.org/appium/appium-espresso-driver/jobs/457348839 for more details. 
So we only perform patching on Windows now

### DIFF
--- a/lib/tools/apk-signing.js
+++ b/lib/tools/apk-signing.js
@@ -50,7 +50,7 @@ apkSigningMethods.executeApksigner = async function (args = []) {
     let binaryPath = apksignerPath;
     if (system.isWindows() && util.isSubPath(binaryPath, originalFolder)) {
       // Workaround for https://github.com/nodejs/node-v0.x-archive/issues/25895
-      binaryPath = path.basename(apksignerPath);
+      binaryPath = path.basename(binaryPath);
     }
     const {stdout, stderr} = await exec(binaryPath, args, {
       cwd: originalFolder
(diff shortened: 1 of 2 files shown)


lib/tools/apk-signing.js
```diff
@@ -47,11 +47,14 @@ apkSigningMethods.executeApksigner = async function (args = []) {
   const apkSigner = await getApksignerForOs(this);
   const originalFolder = path.dirname(apkSigner);
   const getApksignerOutput = async (apksignerPath) => {
-    // Use cwd to workaround https://github.com/nodejs/node-v0.x-archive/issues/25895
-    const {stdout, stderr} = await exec(
-      util.isSubPath(apksignerPath, originalFolder) ? path.basename(apksignerPath) : apksignerPath, args, {
-        cwd: originalFolder
-      });
+    let binaryPath = apksignerPath;
+    if (system.isWindows() && util.isSubPath(binaryPath, originalFolder)) {
```
imurchie: I don't understand how the second part of this would ever be anything but true.

mykola-mokhnach: easy - when `apksignerPath !== apkSigner.`This happens when `getApksignerOutput` is called with patched apksigner script, which is stored in temporary folder

imurchie: Ah! Makes sense.

PR #387, opened 2018-11-20
Update dist in travis

imurchie:
Clean out the `sudo` and `dist` portions of the Travis config, in preparation for [upcoming changes to defaults](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration) in Travis's Linux infrastructure.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: android
-sudo: required
 jdk: oraclejdk8
-dist: trusty
 addons:
   apt:
     sources:


.travis.yml
```diff
@@ -1,7 +1,5 @@
 language: android
-sudo: required
 jdk: oraclejdk8
-dist: trusty
```
mykola-mokhnach: I don't see where `dist` is mentioned in that article

.travis.yml
```diff
@@ -1,7 +1,5 @@
 language: android
-sudo: required
```
mykola-mokhnach: the article only says about removing sudo option if it is set to false

imurchie: Yes. So there is no reason to have this. The more default the config, the better.

.travis.yml
imurchie: Yes. Just taking advantage of this time to get rid of Trusty, which is on its last legs (EOL is April 2019).

mykola-mokhnach: does Travis use the most recent available image if this is not set?

imurchie: Yes. At the moment this is still the Trusty image, but they are working on upgrading (this is part of the infrastructure consolidation). If we specify a version then they are constrained to use it.

PR #390, opened 2018-11-30
Move application install state detector into a separate method

mykola-mokhnach:
This refactoring will help to resolve some discrepancies in UIA2 driver and improves the code quality of appium-adb module

### DIFF
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -448,22 +448,22 @@ apkUtilsMethods.getApplicationInstallState = async function (appPath, pkg = null
   }
   if (_.isInteger(apkVersionCode) && _.isInteger(pkgVersionCode)) {
     if (pkgVersionCode > apkVersionCode) {
-      log.debug(`The installed '${pkg}' package does not require upgrade (${pkgVersionCode} > ${apkVersionCode})`);
+      log.debug(`The version code of the installed '${pkg}' is greater than the application version code (${pkgVersionCode} > ${apkVersionCode})`);
       return this.APP_INSTALL_STATE.NEWER_VERSION_INSTALLED;
     }
     // Version codes might not be maintained. Check version names.
     if (pkgVersionCode === apkVersionCode) {
       if (_.isString(apkVersionName) && _.isString(pkgVersionName) && semver.satisfies(pkgVersionName, `>=${apkVersionName}`)) {
-        log.debug(`The installed '${pkg}' package does not require upgrade ('${pkgVersionName}' >= '${apkVersionName}')`);
+        log.debug(`The version name of the installed '${pkg}' is greater or equal to the application version name ('${pkgVersionName}' >= '${apkVersionName}')`);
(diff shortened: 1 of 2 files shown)


lib/tools/apk-utils.js
```diff
@@ -456,34 +442,87 @@ apkUtilsMethods.installOrUpgrade = async function (appPath, pkg = null, options
...
+    case this.APP_INSTALL_STATE.NOT_INSTALLED:
+      log.debug(`Installing '${appPath}'`);
+      await this.install(appPath, Object.assign({}, options, {replace: false}));
+      return;
+    case this.APP_INSTALL_STATE.NEWER_VERSION_INSTALLED:
+    case this.APP_INSTALL_STATE.SAME_VERSION_INSTALLED:
+      log.debug(`There is no need to install/upgrade '${appPath}'`);
+      return;
+    case this.APP_INSTALL_STATE.OLDER_VERSION_INSTALLED:
+      log.debug(`Executing upgrade of '${appPath}'`);
```
KazuCocoa: Should we upgrade installed app to the new one?
`this.install(appPath, Object.assign({}, options, {replace: true}));` ? (Make sure the option)

lib/tools/apk-utils.js
```diff
@@ -456,34 +442,87 @@ apkUtilsMethods.installOrUpgrade = async function (appPath, pkg = null, options
...
+      break;
+    default:
+      log.debug(`The current install state of '${appPath}' is unknown. Installing anyway`);
```
KazuCocoa: Should we call `await this.install(appPath, Object.assign({}, options, {replace: false}));` like NOT_INSTALLED  since the debug log says `Installing anyway`?

lib/tools/apk-utils.js
```diff
@@ -456,34 +442,87 @@ apkUtilsMethods.installOrUpgrade = async function (appPath, pkg = null, options
...
+ * @property {boolean} useSdcard [false] - Set to true to install the app on sdcard
+ *                                         instead of the device memory.
+ * @property {boolean} grantPermissions [false] - Set to true in order to grant all the
+ *                                                permissions requested in the application's manifest
+ *                                                automatically after the installation is completed
+ *                                                under Android 6+.
+ */
+
+/**
+ * Install the package from the local file system of upgrade it if an older
```
KazuCocoa: s/the local file system of upgrade it/the local file system or upgrade it/ ? (_of_ => _or_)

lib/tools/apk-utils.js
mykola-mokhnach: we don't call return here, so the flow just continues to execute the operator next to the switch block (which is install call). This is also covered by unit tests

lib/tools/apk-utils.js
mykola-mokhnach: same comment as above

lib/tools/apk-utils.js
KazuCocoa: ah... gothat. I missed L527

PR #392, opened 2018-12-05
update setHttpProxy and add deleteHttpProxy

KazuCocoa:
- Update `setHttpProxy`
    - When I set the proxy in `secure` and `system` like the previous implementation, I could not delete them via `adb shell settings delete`. The setting remained. The proxy worked only `global`.
- Add `deleteHttpProxy`
    - They were generated after adding `this.setSetting('global', 'http_proxy', proxy)`

After this work, I would like to add two mobile commands to set/delete http proxy.

- note
    - After deleting the global settings, we must reboot device both emulators and real devices to apply the change
        - I tested.
        - Thus, I'm thinking to add `adb.reboot` when I implement mobile command to delete http proxy
    - I considered implementing the proxy logic in `io.appium.settings` using `WifiConfiguration` like other settings such as locale. but we probably must use reflection to achieve it. Some methods were changed/moved/deleted in recent Android API which we should have got via reflection. So, via adb command is much reasonable than it since the adb command worked Android 9 as well. (I confirmed it on Android 4 and 6 as well.)

### DIFF
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -1514,7 +1514,7 @@ methods.getScreenDensity = async function () {
 };
 
 /**
- * Setup HTTP proxy in device settings.
+ * Setup HTTP proxy in device global settings.
  *
  * @param {string} proxyHost - The host name of the proxy.
  * @param {string|number} proxyPort - The port number to be set.
@@ -1528,10 +1528,19 @@ methods.setHttpProxy = async function (proxyHost, proxyPort) {
     throw new Error(`Call to setHttpProxy method with undefined proxy_port ${proxy}`);
   }
   await this.setSetting('global', 'http_proxy', proxy);
-  await this.setSetting('secure', 'http_proxy', proxy);
-  await this.setSetting('system', 'http_proxy', proxy);
-  await this.setSetting('system', 'global_http_proxy_host', proxyHost);
-  await this.setSetting('system', 'global_http_proxy_port', proxyPort);
+  await this.setSetting('global', 'global_http_proxy_host', proxyHost);
+  await this.setSetting('global', 'global_http_proxy_port', proxyPort);
+};
+
+/**
+ * Delete HTTP proxy in device global settings.
+ * Rebooting the test device is necessary to apply the change.
+ */
+methods.deleteHttpProxy = async function () {
+  await this.shell(['settings', 'delete', 'global', 'http_proxy']);
+  await this.shell(['settings', 'delete', 'global', 'global_http_proxy_host']);
+  await this.shell(['settings', 'delete', 'global', 'global_http_proxy_port']);
+  await this.shell(['settings', 'delete', 'global', 'global_http_proxy_exclusion_list']);
 };
 
 /**
--- a/test/unit/adb-commands-specs.js
+++ b/test/unit/adb-commands-specs.js
@@ -1108,11 +1108,18 @@ describe('adb commands', withMocks({adb, logcat, teen_process, net}, function (m
       let proxyHost = "http://localhost";
       let proxyPort = 4723;
       mocks.adb.expects('setSetting').once().withExactArgs('global', 'http_proxy', `${proxyHost}:${proxyPort}`);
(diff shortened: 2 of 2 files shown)


lib/tools/adb-commands.js
```diff
@@ -1528,10 +1528,19 @@ methods.setHttpProxy = async function (proxyHost, proxyPort) {
...
+
+/**
+ * Delete HTTP proxy in device global settings.
+ * Rebooting the test device is necessary to apply the change.
+ */
+methods.deleteHttpProxy = async function () {
+  await this.shell(['settings', 'delete', 'global', 'http_proxy']);
+  await this.shell(['settings', 'delete', 'global', 'global_http_proxy_host']);
+  await this.shell(['settings', 'delete', 'global', 'global_http_proxy_port']);
+  await this.shell(['settings', 'delete', 'global', 'global_http_proxy_exclusion_list']);
```
mykola-mokhnach: where do we set exclusion list property?

lib/tools/adb-commands.js
```diff
@@ -1528,10 +1528,19 @@ methods.setHttpProxy = async function (proxyHost, proxyPort) {
...
+  await this.setSetting('global', 'global_http_proxy_host', proxyHost);
+  await this.setSetting('global', 'global_http_proxy_port', proxyPort);
+};
+
+/**
+ * Delete HTTP proxy in device global settings.
+ * Rebooting the test device is necessary to apply the change.
+ */
+methods.deleteHttpProxy = async function () {
+  await this.shell(['settings', 'delete', 'global', 'http_proxy']);
```
mykola-mokhnach: this might be more readable being implemented with a loop

lib/tools/adb-commands.js
KazuCocoa: Maybe, we can set it via adb command with `,` separation explained:
https://android.googlesource.com/platform/frameworks/base/+/android-9.0.0_r21/core/java/android/provider/Settings.java#10376

When I set `http_proxy` via adb command, `global_http_proxy_exclusion_list=` (the value was blank) was also generated. So, I deleted here to clean settings generated by `http_proxy`.
Let me leave a comment

KazuCocoa: I'll add it in `setHttpProxy`

mykola-mokhnach: an explanation comment is pretty enough

lib/tools/adb-commands.js
```diff
@@ -1527,20 +1528,33 @@ methods.setHttpProxy = async function (proxyHost, proxyPort) {
   if (_.isUndefined(proxyPort)) {
     throw new Error(`Call to setHttpProxy method with undefined proxy_port ${proxy}`);
   }
-  await this.setSetting('global', 'http_proxy', proxy);
-  await this.setSetting('global', 'global_http_proxy_host', proxyHost);
-  await this.setSetting('global', 'global_http_proxy_port', proxyPort);
+
+  // Read https://android.googlesource.com/platform/frameworks/base/+/android-9.0.0_r21/core/java/android/provider/Settings.java for each property
```
mykola-mokhnach: There is no need to repeat a comment twice if it was already mentioned in the docstring

lib/tools/adb-commands.js
```diff
@@ -1527,20 +1528,33 @@ methods.setHttpProxy = async function (proxyHost, proxyPort) {
...
 /**
  * Delete HTTP proxy in device global settings.
  * Rebooting the test device is necessary to apply the change.
  */
 methods.deleteHttpProxy = async function () {
-  await this.shell(['settings', 'delete', 'global', 'http_proxy']);
-  await this.shell(['settings', 'delete', 'global', 'global_http_proxy_host']);
-  await this.shell(['settings', 'delete', 'global', 'global_http_proxy_port']);
-  await this.shell(['settings', 'delete', 'global', 'global_http_proxy_exclusion_list']);
+  // Read https://android.googlesource.com/platform/frameworks/base/+/android-9.0.0_r21/core/java/android/provider/Settings.java for each property
```
mykola-mokhnach: same here

Issue #395, opened 2018-12-24
Getting Process Ids does'nt work for some devices


yigityardimci:
### **Problem:** 
We are trying to get pid of our application , during our test steps however for some Android devices however, we can't get any pid. We managed to find out that "ps" command doesn't work for every Android device.
**For Alcatel(TCL) 5052Y  "ps" output;**
```
$ ../ANDROID_HOME/platform-tools/adb -s YT55KVQGB64HQGMB shell
A3A:/ $ ps
USER           PID  PPID     VSZ    RSS WCHAN            ADDR S NAME
shell        21989 32761    4952   2532 sigsuspend   b2f1d778 S sh
shell        21992 21989    5568   3016 0            b4396448 R ps
A3A:/ $ getprop ro.product.model
5052Y
A3A:/ $ getprop ro.product.brand
TCL
A3A:/ $
**And "ps -ef" output:**
A3A:/ $ ps -ef
UID            PID  PPID C STIME TTY          TIME CMD
root             1     0 0 10:51:53 ?     00:00:17 init
root             2     0 0 10:51:53 ?     00:00:00 [kthreadd]
root             3     2 0 10:51:53 ?     00:00:08 [ksoftirqd/0]
root             5     2 0 10:51:53 ?     00:00:00 [kworker/0:0H]
root             7     2 0 10:51:53 ?     00:14:14 [rcu_preempt]
root             8     2 0 10:51:53 ?     00:00:10 [rcu_sched]
```

### **Our solution:**
However we find out that , using "ps -ef" isn't the sollution because this command doesn't work for every Android device either. So we found another solution which is using ;
`"for p in /proc/[0-9]*; do [[ $(<$p/cmdline) = " +  name + " ]] && echo ${p##*/}; done"` to get pids.

mykola-mokhnach:
The fix has been published in appium@beta

PR #396, opened 2018-12-27
Use pidof utility for process id detection

mykola-mokhnach:
According to the platform documentation the pidof utility should be available on all devices since API 23. Older devices might have it or might have not. In such case we fallback to the previous ps-based PID detection method.
Partially addresses https://github.com/appium/appium-adb/issues/395


mykola-mokhnach:
makes sense. I'll update this implementation as soon as `getPIDsByName` is stabilized

PR #397, opened 2018-12-27
Only retry adbExec if it is a protocol or device not found error

imurchie:
After some discussion with @mykola-mokhnach we decided to try to retry _only_ when in the special cases of a protocol or device not found error.

No one knows why exactly the `retry` is necessary. If things get flakey it can be returned and noted.

### DIFF
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -289,17 +289,17 @@ systemCallMethods.adbExecEmu = async function (cmd) {
  */
 systemCallMethods.adbExec = async function (cmd, opts = {}) {
   if (!cmd) {
-    throw new Error("You need to pass in a command to adbExec()");
+    throw new Error('You need to pass in a command to adbExec()');
   }
+
   // setting default timeout for each command to prevent infinite wait.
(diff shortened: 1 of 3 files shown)


lib/tools/system-calls.js
```diff
@@ -315,20 +315,21 @@ systemCallMethods.adbExec = async function (cmd, opts = {}) {
         log.info(`Error sending command, reconnecting device and retrying: ${cmd}`);
         await sleep(1000);
         await this.getDevicesWithRetry();
+
+        // try again
+        return await execFunc();
```
mykola-mokhnach: can this create an endless recursion?

imurchie: Yes, indeed. :|

lib/tools/system-calls.js
```diff
@@ -316,8 +318,10 @@ systemCallMethods.adbExec = async function (cmd, opts = {}) {
         await sleep(1000);
         await this.getDevicesWithRetry();
 
-        // try again
-        return await execFunc();
+        // try again one time
+        if (callCount <= 1) {
```
mykola-mokhnach: I'd rather have a constant for it

PR #422, opened 2019-03-16
Change 'windows' to 'displays'

dpgraham:
As of Android SDK 29 (Android Q), this apk (com.google.android.permissioncontroller) opens prior to the AUT opening to prompt the user which permissions to allow. 

https://user-images.githubusercontent.com/852574/54451599-6faa6e80-4710-11e9-8917-95126af79e79.png

^ This breaks tests and makes the AUT not start.

The way around this (what this PR does) is automatically grants all permissions for SDK level 29+ on install. Automatically granting this permission makes it bypass this APK.

This would be a breaking change though, so I'd like other opinions.

The choices are
1. Always grant permissions automatically, thus deprecating `autoGrantPermissions` capability. Add a new capability, something like `revokePermissions`, which takes an array of permissions and revokes them after app is started
1. Use UiAutomator to select the "Continue" button when it encounters the `permissioncontroller` apk. Although by the looks of it, this would do the same thing, because it would allow all permissions by default.
1. Use UiAutomator to unselect all permissions on the list. This should maintain backwards compatibility, but seems kind of hacky.

(thank you to @sridharreddysuram123 for showing me the fix for this)

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -390,7 +390,7 @@ function buildInstallArgs (apiLevel, options = {}) {
   if (options.useSdcard) {
     result.push('-s');
   }
-  if (options.grantPermissions) {
+  if (options.grantPermissions || apiLevel >= 28) {
     if (apiLevel < 23) {
       log.debug(`Skipping permissions grant option, since ` +
                 `the current API level ${apiLevel} does not support applications ` +
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -390,7 +390,7 @@ function buildInstallArgs (apiLevel, options = {}) {
   if (options.useSdcard) {
     result.push('-s');
   }
-  if (options.grantPermissions || apiLevel >= 28) {
+  if (options.grantPermissions) {
     if (apiLevel < 23) {
       log.debug(`Skipping permissions grant option, since ` +
                 `the current API level ${apiLevel} does not support applications ` +
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -87,6 +87,7 @@ methods.getApiLevel = async function getApiLevel () {
 
       // Temp workaround. Android Q beta emulators report SDK 28 when they should be 29
       if (apiLevel === 28 && (await this.getDeviceProperty('ro.build.version.release')).toLowerCase() === 'q') {
+        log.debug('Release version is Q but found API Level 28. Setting API Level to 29');
         apiLevel = 29;
       }
       this._apiLevel = apiLevel;
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -86,8 +86,10 @@ methods.getApiLevel = async function getApiLevel () {
       let apiLevel = parseInt(strOutput.trim(), 10);
 
       // Temp workaround. Android Q beta emulators report SDK 28 when they should be 29
-      if ((await this.getDeviceProperty('ro.build.version.release')).toLowerCase() === 'q') {
-        apiLevel = 29;
+      if (apiLevel === 28) {
+        if ((await this.getDeviceProperty('ro.build.version.release')).toLowerCase() === 'q') {
+          apiLevel = 29;
+        }
       }
       this._apiLevel = apiLevel;
       log.debug(`Device API level: ${this._apiLevel}`);
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -83,7 +83,13 @@ methods.getApiLevel = async function getApiLevel () {
   if (!_.isInteger(this._apiLevel)) {
     try {
       const strOutput = await this.getDeviceProperty('ro.build.version.sdk');
-      this._apiLevel = parseInt(strOutput.trim(), 10);
+      let apiLevel = parseInt(strOutput.trim(), 10);
+
+      // Temp workaround. Android Q beta emulators report SDK 28 when they should be 29
+      if ((await this.getDeviceProperty('ro.build.version.release')).toLowerCase() === 'q') {
+        apiLevel = 29;
+      }
+      this._apiLevel = apiLevel;
       log.debug(`Device API level: ${this._apiLevel}`);
       if (isNaN(this._apiLevel)) {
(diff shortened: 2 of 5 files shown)


KazuCocoa:
Sounds reasonable for me, basically.

https://github.com/appium/appium-android-driver/pull/514 will come in next release. Users can revoke permissions if they would like to handle permissions. We can address the command in `autoGrantPermissions` doc and Android Q adaptation release note.

I think we do not need to make `autoGrantPermissions` deprecate.
Can we address 

> Use UiAutomator to unselect all permissions on the list. This should maintain backwards compatibility, but seems kind of hacky.

as a documentation? e.g. _Set xxxxx as waitForActivity and call `find_element(xxxx)`. Then, you can enable arbitrary permissions in the controller_.

If a user would like to simulate user actions from the `permissioncontroller` when they upgrade their app under test from old one to the latest one, he/she can follow it with UIA.

mykola-mokhnach:
I tend to agree with @KazuCocoa 's point. Lets simply describe the new behaviour for Android Q in the doc (e. g. what to do it permissions are not granted automatically, which activity to expect and so on) 

Although, I'm not quite sure this dialog can be properly handled by Espresso. If it really couldn't then I would add another Espresso cap to handle it and select all/none/some permissions there.

mykola-mokhnach:
unit tests should be updated

lib/tools/apk-utils.js
```diff
@@ -164,7 +164,9 @@ apkUtilsMethods.startApp = async function startApp (startAppOptions = {}) {
  */
 apkUtilsMethods.getFocusedPackageAndActivity = async function getFocusedPackageAndActivity () {
   log.debug('Getting focused package and activity');
-  const cmd = ['dumpsys', 'window', 'windows'];
+  const apiLevel = await this.getApiLevel();
+  const dumpsysArg = apiLevel >= 29 ? 'displays' : 'windows';
```
mykola-mokhnach: there should be a comment about why it is necessary. Also, we need to find other places in the code where this call is used and probable isolate them into a separate method call

lib/helpers.js
```diff
@@ -390,7 +390,7 @@ function buildInstallArgs (apiLevel, options = {}) {
   if (options.useSdcard) {
     result.push('-s');
   }
-  if (options.grantPermissions) {
+  if (options.grantPermissions || apiLevel >= 28) {
```
mykola-mokhnach: 👎

lib/tools/adb-commands.js
```diff
@@ -83,7 +83,13 @@ methods.getApiLevel = async function getApiLevel () {
   if (!_.isInteger(this._apiLevel)) {
     try {
       const strOutput = await this.getDeviceProperty('ro.build.version.sdk');
-      this._apiLevel = parseInt(strOutput.trim(), 10);
+      let apiLevel = parseInt(strOutput.trim(), 10);
+
+      // Temp workaround. Android Q beta emulators report SDK 28 when they should be 29
+      if ((await this.getDeviceProperty('ro.build.version.release')).toLowerCase() === 'q') {
```
mykola-mokhnach: check if apiLevel === 28 first to avoid an extra shell call

dpgraham:
Tests get stuck when the `permissioncontroller` activity comes up because it times out waiting for the AUT main activity.

To fix this we need to either 1. use uia2 to dismiss it; 2. have it not show up in the first place by setting the `-g` flag on `adb install`; 3. make the `permissioncontroller` interactable.

As far as I can tell, any solution will break tests, so we need to pick the functionality that will minimize the impact.

I say we  do this:
* Deprecate `autoGrantPermissions` and grant permissions the default behaviour (for SDK >= 29)
* Add a new capability that blacklists the permissions should not be set (something like `deniedPermissions="BLUETOOTH, CHANGE_NETWORK_STATE"`). 
* If that cap is set, disable those permissions in `permissioncontroller`, using UIA2

mykola-mokhnach:
> 2. have it not show up in the first place by setting the -g flag on adb install

For that we'll have to change the fastReset (the default) behaviour, so instead `pm clear` (which clears permissions as well) we could perform uninstall with `-k` option (keep data) and then `install -g`.

mykola-mokhnach:
The problem with `permissioncontroller` is that it's might be customised by different OS vendors. Lets split the task into two. I'll take care about correct autoGrantPermissions behaviour, so we don't see permissions dialog if it is set. The second task might be different for UIA2/Espresso.

For UIA2 we could do nothing (since the permissions dialog is anyway) and for Espresso we could have a new capability

dpgraham:
Fixed it @mykola-mokhnach .... got rid of the API Level check for granting permissions.

lib/tools/adb-commands.js
```diff
@@ -83,7 +83,15 @@ methods.getApiLevel = async function getApiLevel () {
   if (!_.isInteger(this._apiLevel)) {
     try {
       const strOutput = await this.getDeviceProperty('ro.build.version.sdk');
-      this._apiLevel = parseInt(strOutput.trim(), 10);
+      let apiLevel = parseInt(strOutput.trim(), 10);
+
+      // Temp workaround. Android Q beta emulators report SDK 28 when they should be 29
+      if (apiLevel === 28) {
+        if ((await this.getDeviceProperty('ro.build.version.release')).toLowerCase() === 'q') {
```
mykola-mokhnach: no need to have a nested if here. Use &&

dpgraham: Done.

lib/tools/adb-commands.js
```diff
@@ -83,7 +83,13 @@ methods.getApiLevel = async function getApiLevel () {
   if (!_.isInteger(this._apiLevel)) {
     try {
       const strOutput = await this.getDeviceProperty('ro.build.version.sdk');
-      this._apiLevel = parseInt(strOutput.trim(), 10);
+      let apiLevel = parseInt(strOutput.trim(), 10);
+
+      // Temp workaround. Android Q beta emulators report SDK 28 when they should be 29
+      if (apiLevel === 28 && (await this.getDeviceProperty('ro.build.version.release')).toLowerCase() === 'q') {
+        apiLevel = 29;
```
mykola-mokhnach: some log message here might be useful

PR #423, opened 2019-03-17
Introduce apk caching

mykola-mokhnach:
Normal call of `adb install` internally uploads the apk to the remote fs, executes `pm install` on it and then deleted the apk (see https://android.googlesource.com/platform/tools/base/+/master/ddmlib/src/main/java/com/android/ddmlib/Device.java#879). With Appium, though, it is usually necessary to reinstall a package multiple times and uploading it to the device for each install is time-consuming, especially for bigger APKs. This  PR automatically puts an APK into the internal cache, which can contain up to 10 items in order to save time for upload. In case the size of the remote cache reaches the size limit the algorithm will clean out the outdated items.

### DIFF
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -23,10 +23,7 @@ apkUtilsMethods.APP_INSTALL_STATE = {
   OLDER_VERSION_INSTALLED: 'olderVersionInstalled',
 };
 const REMOTE_CACHE_ROOT = '/data/local/tmp/appium_cache';
-const REMOTE_CACHE_ENTRIES_LIMIT = parseInt(process.env.APPIUM_ANDROID_REMOTE_APK_CACHE_LIMIT || 10, 10);
-const REMOTE_CACHE = new LRU({
-  max: REMOTE_CACHE_ENTRIES_LIMIT,
-});
+
 
 /**
  * Check whether the particular package is present on the device under test.
@@ -386,14 +383,22 @@ apkUtilsMethods.cacheApk = async function (apkPath, options = {}) {
     log.info(`The application at '${apkPath}' is already cached to '${remotePath}'`);
   } else {
     log.info(`Caching the application at '${apkPath}' to '${remotePath}'`);
+    const started = process.hrtime();
     await this.push(apkPath, remotePath, {timeout: options.timeout});
+    const [seconds, nanos] = process.hrtime(started);
+    log.info(`The upload of '${path.basename(apkPath)}' took ${(seconds + nanos / 1000000000.0).toFixed(2)}s`);
   }
-  REMOTE_CACHE.set(appHash, remotePath);
+  if (!this.remoteAppsCache) {
+    this.remoteAppsCache = new LRU({
+      max: this.remoteAppsCacheLimit,
+    });
+  }
+  this.remoteAppsCache.set(appHash, remotePath);
   // If the remote cache exceeds REMOTE_CACHE_ENTRIES_LIMIT, remove least recently used entries
   const entriesToCleanup = remoteCachedFiles
     .map((x) => path.posix.join(REMOTE_CACHE_ROOT, x))
-    .filter((x) => !REMOTE_CACHE.has(path.posix.parse(x).name))
-    .slice(REMOTE_CACHE_ENTRIES_LIMIT);
+    .filter((x) => !this.remoteAppsCache.has(path.posix.parse(x).name))
+    .slice(this.remoteAppsCacheLimit);
   if (!_.isEmpty(entriesToCleanup)) {
     try {
       await this.shell(['rm', '-f', ...entriesToCleanup]);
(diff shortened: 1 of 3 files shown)


lib/tools/apk-utils.js
```diff
@@ -349,6 +355,54 @@ apkUtilsMethods.installFromDevicePath = async function (apkPathOnDevice, opts =
...
+
+/**
+ * Caches the given APK at a remote location to speed up further APK deployments.
+ *
+ * @param {string} apkPath - Full path to the apk on the local FS
+ * @param {?CachingOptions} options - Caching options
+ * @returns {string} - Full path to the cached apk on the remote file system
+ */
+apkUtilsMethods.cacheApk = async function (apkPath, options = {}) {
+  const appHash = await fs.hash(apkPath);
```

lib/tools/apk-utils.js
```diff
@@ -349,6 +355,54 @@ apkUtilsMethods.installFromDevicePath = async function (apkPathOnDevice, opts =
...
+  const remotePath = path.posix.join(REMOTE_CACHE_ROOT, `${appHash}.apk`);
+  const remoteCachedFiles = [];
```
dpgraham: A comment explaining this block of code would be helpful

```js
// Get current contents of remote cache
```

lib/tools/apk-utils.js
```diff
@@ -349,6 +355,54 @@ apkUtilsMethods.installFromDevicePath = async function (apkPathOnDevice, opts =
...
+  try {
+    remoteCachedFiles.push(...(await this.ls(REMOTE_CACHE_ROOT, ['-t', '-1'])));
+  } catch (e) {
+    log.debug(`Got an error '${e.message}' while getting the list of files in the cache. ` +
+      `Assuming the cache does not exist yet`);
+    await this.shell(['mkdir', '-p', REMOTE_CACHE_ROOT]);
+  }
+  log.debug(`The count of applications in the cache: ${remoteCachedFiles.length}`);
```
dpgraham:
```js
// Pushing apk to remote cache
```

lib/tools/apk-utils.js
```diff
@@ -349,6 +355,54 @@ apkUtilsMethods.installFromDevicePath = async function (apkPathOnDevice, opts =
...
+  if (_.isEmpty(remoteCachedFiles.filter((x) => path.posix.parse(x).name === appHash))) {
+    log.info(`Caching the application at '${apkPath}' to '${remotePath}'`);
+    await this.push(apkPath, remotePath, {timeout: options.timeout});
+  } else {
+    log.info(`The application at '${apkPath}' is already cached to '${remotePath}'`);
+  }
+  REMOTE_CACHE.set(appHash, remotePath);
+  const entriesToCleanup = remoteCachedFiles
```
dpgraham:
```js
// If remote cache exceeds REMOTE_CACHE_SIZE, remove least recently used entries
```

lib/tools/apk-utils.js
```diff
@@ -21,6 +22,11 @@ apkUtilsMethods.APP_INSTALL_STATE = {
   SAME_VERSION_INSTALLED: 'sameVersionInstalled',
   OLDER_VERSION_INSTALLED: 'olderVersionInstalled',
 };
+const REMOTE_CACHE_ROOT = '/data/local/tmp/appium_cache';
+const REMOTE_CACHE_SIZE = 10;
```
dpgraham: I'd rather call it `REMOTE_CACHE_ENTRIES_LIMIT`

lib/tools/apk-utils.js
```diff
@@ -349,6 +355,54 @@ apkUtilsMethods.installFromDevicePath = async function (apkPathOnDevice, opts =
...
+ */
+
+/**
+ * Caches the given APK at a remote location to speed up further APK deployments.
+ *
+ * @param {string} apkPath - Full path to the apk on the local FS
+ * @param {?CachingOptions} options - Caching options
+ * @returns {string} - Full path to the cached apk on the remote file system
+ */
+apkUtilsMethods.cacheApk = async function (apkPath, options = {}) {
```
dpgraham: Could we have one e2e test for this. Maybe something that calls `cacheApk` method `REMOTE_CACHE_SIZE + 1` times and verifies that the cache `/data/local/tmp/appium_cache` only contains `REMOTE_CACHE_SIZE` entries. I'm pretty confident this implementation will work, but I would like something to ensure that if we make adjustments to this algorithm, that it will always clean up for itself and that it won't ever grow indefinitely.

dpgraham:
Some comments, but I like this approach a lot.

Using this approach, would it make sense to change the `fastReset` behaviour of Android Driver to always uninstall the AUT and then re-install the apk that's sitting in cache?

lib/tools/apk-utils.js
dpgraham: Also, maybe a test that validates that the exact same file never gets added to the cache more than once. But that when it gets added back, it moves up in "recently used" priority.

mykola-mokhnach: added unit test

PR #427, opened 2019-03-20
Throw proper exception if the application is not writeable for zipalign

mykola-mokhnach:
Related to https://github.com/appium/appium/issues/12355


lib/tools/apk-signing.js
```diff
@@ -230,6 +231,13 @@ apkSigningMethods.zipAlignApk = async function zipAlignApk (apk) {
   } catch (e) {
     log.debug(`'${apk}' is not zip-aligned. Aligning`);
   }
+  try {
+    await fs.access(apk, _fs.W_OK);
+  } catch (e) {
+    throw new Error(`The file at '${apk}' is not writeable. ` +
+      `Please grant write permissions to this file or to its parent folder '${path.dirname(apk)}' ` +
+      `for the Appium process, so it could zip-align the file`);
```
imurchie: `so it can zip-align`.

PR #428, opened 2019-03-20
Properly handle ls attributes presence for different API levels

mykola-mokhnach:
Android API 28 does not align each file in ls output into a separate row (`-1` option is needed for that). Although the older APIs don't support such option at all and return an error if it is provided. This PR tries to properly handle both cases

### DIFF
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -373,15 +373,15 @@ apkUtilsMethods.cacheApk = async function (apkPath, options = {}) {
   try {
     const errorMarker = '_ERROR_';
     let lsOutput = null;
-    if (this._areExtendedLsOptionsSupported === true || !util.hasValue(this._areExtendedLsOptionsSupported)) {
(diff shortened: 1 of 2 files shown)


lib/tools/apk-utils.js
```diff
@@ -371,9 +371,31 @@ apkUtilsMethods.cacheApk = async function (apkPath, options = {}) {
...
   try {
-    remoteCachedFiles.push(...(await this.ls(REMOTE_CACHE_ROOT, ['-t', '-1'])));
+    const errorMarker = '_ERROR_';
+    let lsOutput = null;
+    if (this._areExtendedLsOptionsSupported === true || !util.hasValue(this._areExtendedLsOptionsSupported)) {
+      lsOutput = await this.shell([`ls -t -1 ${REMOTE_CACHE_ROOT} || echo ${errorMarker}`]);
+    }
+    if (!_.isString(lsOutput) || (lsOutput.includes(errorMarker) && !lsOutput.includes(REMOTE_CACHE_ROOT))) {
+      if (!util.hasValue(this._areExtendedLsOptionsSupported)) {
+        log.debug('The current Android API does not support extended ls options. ' +
```

PR #429, opened 2019-03-21
Only apply caching if remoteAppsCacheLimit is greater than zero

mykola-mokhnach:
Sometimes it might be needed to disable caching (for example in cloud, where devices are reset automatically after each test)

### DIFF
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -472,16 +472,21 @@ apkUtilsMethods.install = async function install (appPath, options = {}) {
     timeoutCapName: 'androidInstallTimeout',
   }, options);
 
-  const cachedRemotePath = await this.cacheApk(appPath, {
-    timeout: options.timeout,
-  });
   const installArgs = buildInstallArgs(await this.getApiLevel(), options);
+  const installOpts = {
+    timeout: options.timeout,
+    timeoutCapName: options.timeoutCapName,
+  };
+  let installMethod = async () => await this.adbExec(['install', ...installArgs, appPath], installOpts);
+  if (this.remoteAppsCacheLimit > 0) {
+    const cachedRemotePath = await this.cacheApk(appPath, {
+      timeout: options.timeout,
+    });
+    installMethod = async () => await this.shell(['pm', 'install', ...installArgs, cachedRemotePath], installOpts);
+  }
   try {
     const started = process.hrtime();
-    const output = await this.shell(['pm', 'install', ...installArgs, cachedRemotePath], {
-      timeout: options.timeout,
-      timeoutCapName: options.timeoutCapName,
-    });
+    const output = await installMethod();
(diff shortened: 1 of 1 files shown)


lib/tools/apk-utils.js
```diff
@@ -472,16 +472,21 @@ apkUtilsMethods.install = async function install (appPath, options = {}) {
...
   const installArgs = buildInstallArgs(await this.getApiLevel(), options);
-  try {
-    const started = process.hrtime();
-    const output = await this.shell(['pm', 'install', ...installArgs, cachedRemotePath], {
+  const installOpts = {
+    timeout: options.timeout,
+    timeoutCapName: options.timeoutCapName,
+  };
+  let installMethod = async () => await this.adbExec(['install', ...installArgs, appPath], installOpts);
+  if (this.remoteAppsCacheLimit > 0) {
```
dpgraham: Maybe just add a comment here to clarify that `remoteAppCacheLimit <= 0` means to not cache apk on devices.

mykola-mokhnach: I try to write code the way it does not need explanation comments, but if you think it is still necessary...

dpgraham: I'm not a stickler for having comments _everywhere_ but in this case, I just want an explanation of why the caching behaviour is different for `this.remoteAppsCacheLimit > 0` vs. `this.remoteAppsCacheLimit <= 0`.

dpgraham:
I've been thinking about this some more and I think we'll need to inform cloud providers of this feature (particularly real device providers, emulator providers typically delete the entire VM after session completion). 

They'll need to make sure they set this to zero for _every_ session (or have a clean-up script that wipes out the cache). We don't want apk's lying around for the next user to be able to access.

@KazuCocoa is this a possible issue for Headspin?

KazuCocoa:
In our environment, users can use their own devices. In the case, probably usually no. (But in some aspect, we need to consider we should erase tmp files in `/data/local/tmp/appium_cache`

I agree with you to share this change with other providers to ensure they can erase unexpected files in their devices.

---

a note for the _hash_

The cached file name is hashed.
If an apk sign with another cert, the hash should change.
e.g.
An apk has xxxxxxxxxxxxxxxx with cert A.
Sign with appium cert, then the hash will be 0a2f432fa924f765d9xxxxx.
This cache logic can identify they are different (so it does not use cached apk.)
https://github.com/appium/appium-adb/pull/423/files#diff-9a49d110e1c61a8dc30a6ac9a8335185R369

mykola-mokhnach:
@dpgraham I assume it is a good practice to make sure temp files were cleaned up after each session in any case, since these might be also use by the apps or the system itself to store sensitive data.

mykola-mokhnach:
@KazuCocoa Signing happens before an apk gets uploaded. Also, Appium replaces the original apk with the resigned one by default unless noSign option is set.

PR #433, opened 2019-03-28
Make emulator name comparison case insensitive

mykola-mokhnach:
Resolves https://discuss.appium.io/t/android-emulator-is-being-started-again/25681


KazuCocoa:
I checked created AVD id was unique even if I created in the same avd which is case insensitive name. (so conflict won't happen)

PR #439, opened 2019-04-03
Create directory before push to avoid permission issue

dpgraham:
`pushFile` isn't working on some later versions of Android SDK. 

Was getting error `remote secure_mkdirs failed: Operation not permitted`. Workaround is to create the directory first using `adb shell mkdir` before pushing the file. That directory will have the correct permissions.

(another issue that had this problem: https://github.com/bazelbuild/examples/issues/77)

### DIFF
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -966,7 +966,7 @@ methods.rimraf = async function rimraf (path) {
  *                        options.
  */
 methods.push = async function push (localPath, remotePath, opts) {
-  await this.mkdir(path.dirname(remotePath));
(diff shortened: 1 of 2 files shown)


mykola-mokhnach:
Please update the unit tests

lib/tools/adb-commands.js
```diff
@@ -966,6 +966,7 @@ methods.rimraf = async function rimraf (path) {
  *                        options.
  */
 methods.push = async function push (localPath, remotePath, opts) {
+  await this.mkdir(path.dirname(remotePath));
```
mykola-mokhnach: path.posix.dirname

PR #442, opened 2019-04-08
Add a possibility to enforce SDK Tools version

mykola-mokhnach:
Related to https://github.com/appium/appium/issues/12442

### DIFF
--- a/lib/adb.js
+++ b/lib/adb.js
@@ -28,6 +28,7 @@ const DEFAULT_OPTS = {
   adbPort: DEFAULT_ADB_PORT,
   adbExecTimeout: DEFAULT_ADB_EXEC_TIMEOUT,
   remoteAppsCacheLimit: 10,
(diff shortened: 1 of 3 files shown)


dpgraham:
Also, looks like it's CI is failing.

They're all of the form `Error: Cannot extract resources from '/Users/vsts/agent/2.149.2/work/1/s/test/fixtures/ContactManager.apk'. Original error: The "file" argument must be of type string. Received type undefined`

mykola-mokhnach:
Should be better now

test/functional/adb-e2e-specs.js
```diff
@@ -23,6 +26,13 @@ describe('ADB', function () {
     await adb.initAapt();
     adb.binaries.aapt.should.contain('aapt');
   });
+  it('should initialize aapt using the enforced build tools path', async function () {
```
dpgraham: 👍

PR #443, opened 2019-04-08
Fix: Restart ADB server when devices go offline

dpgraham:
This addresses two problems: `adb root/unroot` 1. works intermittently, 2. sometimes cases the device to go offline.

The fix is:
1. retries `root/unroot` instead of calling it just once
2. after a root/unroot occurs, check if there's no connected devices and if there aren't any, restart adb (uses a new method called `restartAdbIfDevicesOffline ()`

This PR also:
1. Uses mocha multireporters so that the logs show tests being called
2. Fixes an IME test that wasn't working

### DIFF
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -478,7 +478,7 @@ methods.setIME = async function setIME (imeId) {
 /**
  * Get the default input method on the device under test.
  *
- * @return {string} The name of the default input method.
+ * @return {string} The name of the default input method or null if none found.
  */
 methods.defaultIME = async function defaultIME () {
   try {
@@ -1461,16 +1461,22 @@ methods.getDeviceProperty = async function getDeviceProperty (property) {
 };
 
 /**
+ * @typedef setPropOpts
+ * @property {boolean} privileged - Do we run setProp as a privileged command?
+ */
+
+/**
  * Set the particular property of the device under test.
  *
  * @param {string} property - The name of the property. This name should
  *                            be known to _adb shell setprop_ tool.
  * @param {string} val - The new property value.
- * @param {boolean} privileged - Do we need to set root access for call to 'setProp'?
+ * @param {setPropOpts} opts
  *
  * @throws {error} If _setprop_ utility fails to change property value.
  */
-methods.setDeviceProperty = async function setDeviceProperty (prop, val, privileged = true) {
+methods.setDeviceProperty = async function setDeviceProperty (prop, val, opts = {}) {
+  const {privileged} = opts;
   log.debug(`Setting device property '${prop}' to '${val}'`);
   await this.shell(['setprop', prop, val], {
     privileged,
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -1476,7 +1476,7 @@ methods.getDeviceProperty = async function getDeviceProperty (property) {
  * @throws {error} If _setprop_ utility fails to change property value.
  */
 methods.setDeviceProperty = async function setDeviceProperty (prop, val, opts = {}) {
-  const {privileged} = opts;
+  const {privileged = true} = opts;
   log.debug(`Setting device property '${prop}' to '${val}'`);
   await this.shell(['setprop', prop, val], {
     privileged,
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -1303,10 +1303,8 @@ methods.killProcessByPID = async function killProcessByPID (pid) {
         throw e;
       }
       log.info(`Cannot kill PID ${pid} due to insufficient permissions. Retrying as root`);
-      try {
-        let {isSuccessful} = await this.root();
-        becameRoot = isSuccessful;
-      } catch (ign) {}
+      let {isSuccessful} = await this.root();
+      becameRoot = isSuccessful;
       await this.shell(['kill', '-0', pid]);
     }
     const timeoutMs = 1000;
@@ -1462,7 +1460,7 @@ methods.getDeviceProperty = async function getDeviceProperty (property) {
 
 /**
  * @typedef {object} setPropOpts
- * @property {boolean} privileged - Do we run setProp as a privileged command?
+ * @property {boolean} privileged - Do we run setProp as a privileged command? Default true.
  */
 
 /**
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -1301,7 +1301,8 @@ methods.killProcessByPID = async function killProcessByPID (pid) {
       }
       log.info(`Cannot kill PID ${pid} due to insufficient permissions. Retrying as root`);
       try {
-        becameRoot = await this.root();
+        let {isSuccessful} = await this.root();
+        becameRoot = isSuccessful;
       } catch (ign) {}
       await this.shell(['kill', '-0', pid]);
     }
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -478,7 +478,7 @@ methods.setIME = async function setIME (imeId) {
 /**
  * Get the default input method on the device under test.
  *
- * @return {string} The name of the default input method or null if none found.
+ * @return {?string} The name of the default input method
  */
 methods.defaultIME = async function defaultIME () {
   try {
@@ -1461,7 +1461,7 @@ methods.getDeviceProperty = async function getDeviceProperty (property) {
 };
 
 /**
- * @typedef setPropOpts
+ * @typedef {object} setPropOpts
  * @property {boolean} privileged - Do we run setProp as a privileged command?
  */
 
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -1466,13 +1466,14 @@ methods.getDeviceProperty = async function getDeviceProperty (property) {
  * @param {string} property - The name of the property. This name should
  *                            be known to _adb shell setprop_ tool.
  * @param {string} val - The new property value.
+ * @param {boolean} privileged - Do we need to set root access for call to 'setProp'?
  *
  * @throws {error} If _setprop_ utility fails to change property value.
  */
-methods.setDeviceProperty = async function setDeviceProperty (prop, val) {
+methods.setDeviceProperty = async function setDeviceProperty (prop, val, privileged = true) {
   log.debug(`Setting device property '${prop}' to '${val}'`);
   await this.shell(['setprop', prop, val], {
-    privileged: true,
+    privileged,
   });
 };
 
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -1109,7 +1109,7 @@ methods.restart = async function restart () {
     await this.waitForDevice(60);
     await this.startLogcat();
   } catch (e) {
-    throw new Error(`Restart failed. Orginial error: ${e.message}`);
+    throw new Error(`Restart failed. Original error: ${e.message}`);
   }
 };
 
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -483,6 +483,9 @@ methods.setIME = async function setIME (imeId) {
 methods.defaultIME = async function defaultIME () {
   try {
     let engine = await this.getSetting('secure', 'default_input_method');
+    if (engine === 'null') {
+      return null;
+    }
     return engine.trim();
   } catch (e) {
     throw new Error(`Error getting default IME. Original error: ${e.message}`);
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -857,7 +857,7 @@ systemCallMethods.reboot = async function reboot (retries = DEFAULT_ADB_REBOOT_R
  * @return {boolean} True if the device is rooted.
  */
 systemCallMethods.root = async function root (unroot = false) {
-  let cmd = unroot ? 'unroot' : 'root';
+  const cmd = unroot ? 'unroot' : 'root';
   let wasAlreadyRooted = false;
   try {
     let {stdout} = await exec(this.executable.path, [cmd]);
@@ -879,8 +879,7 @@ systemCallMethods.root = async function root (unroot = false) {
 
     // If 'closed' was in the stderr (e.g.: `adb: unable to connect for root: closed\n"`) then the
     // device probably went offline, so restart ADB
-    console.log('!!!!!!!!!', err, JSON.stringify(err));
-    if (stderr.toLowerCase().includes('closed')) {
+    if (stderr.toLowerCase().includes('closed') || stderr.toLowerCase().includes('device offline')) {
       log.warn(`${cmd} caused device to go offline. Restarting adb.`);
       await this.restartAdb();
     }
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -832,14 +832,19 @@ systemCallMethods.waitForDevice = async function waitForDevice (appDeviceReadyTi
  */
 systemCallMethods.reboot = async function reboot (retries = DEFAULT_ADB_REBOOT_RETRIES) {
   try {
-    await this.shell(['stop'], {
-      privileged: true,
-    });
+    // Get root access so we can run the following shell commands that require root access
+    const {wasAlreadyRooted} = await this.root();
+
+    // Stop and re-start the device
+    await this.shell(['stop']);
     await B.delay(2000); // let the emu finish stopping;
-    await this.setDeviceProperty('sys.boot_completed', 0);
-    await this.shell(['start'], {
-      privileged: true,
-    });
+    await this.setDeviceProperty('sys.boot_completed', 0, false);
+    await this.shell(['start']);
+
+    // Return root state to what it was before
+    if (!wasAlreadyRooted) {
+      await this.unroot();
+    }
   } catch (e) {
     const {message} = e;
 
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -375,6 +375,7 @@ systemCallMethods.shell = async function shell (cmd, opts = {}) {
     keepPrivileged,
   } = opts;
   let shouldRestoreUser = false;
+  let isRoot = await this.isRoot();
   if (privileged) {
     try {
       shouldRestoreUser = !(await this.adbExec(['root'], opts)).includes('already running as root');
@@ -383,9 +384,9 @@ systemCallMethods.shell = async function shell (cmd, opts = {}) {
         // Do not show the warning for real devices, where root is locked
         log.warn(`Cannot run adbd as root. Original error: ${err.message}`);
       }
+      // adb 'root' can cause device to go offline, restart if so
+      await this.restartAdbIfDevicesOffline();
     }
-    // adb 'root' can cause device to go offline, restart if so
-    await this.restartAdbIfDevicesOffline();
   }
   let didCommandFail = false;
   try {
@@ -401,9 +402,10 @@ systemCallMethods.shell = async function shell (cmd, opts = {}) {
         await this.adbExec(['unroot'], opts);
       } catch (err) {
         log.warn(`Cannot run adbd as non-root. Original error: ${err.message}`);
+
+        // adb 'unroot' can cause device to go offline, restart if so
+        await this.restartAdbIfDevicesOffline();
       }
-      // adb 'unroot' can cause device to go offline, restart if so
-      await this.restartAdbIfDevicesOffline();
     }
   }
 };
@@ -877,23 +879,21 @@ systemCallMethods.restartAdbIfDevicesOffline = async function restartAdbIfDevice
  *                   if the switch failed.
  */
 systemCallMethods.root = async function root () {
-  let isRootSuccessful = true;
   try {
-    let {stdout} = await retryInterval(10, 1000, async () => await exec(this.executable.path, ['root']));
+    let {stdout} = await exec(this.executable.path, ['root']);
 
     // on real devices in some situations we get an error in the stdout
     if (stdout && stdout.indexOf('adbd cannot run as root') !== -1) {
       throw new Error(stdout.trim());
     }
+    return true;
   } catch (err) {
     log.warn(`Unable to root adb daemon: '${err.message}'. Continuing.`);
-    isRootSuccessful = false;
+    // adb 'root' can cause device to go offline, restart if so
+    await this.restartAdbIfDevicesOffline();
+
+    return false;
   }
-
-  // adb 'root' can cause device to go offline, restart if so
-  await this.restartAdbIfDevicesOffline();
-
-  return isRootSuccessful;
 };
 
 /**
@@ -903,17 +903,15 @@ systemCallMethods.root = async function root () {
  *                   if the switch failed.
  */
 systemCallMethods.unroot = async function unroot () {
-  let isUnrootSuccessful = true;
   try {
-    await retryInterval(10, 1000, async () => await exec(this.executable.path, ['unroot']));
+    await exec(this.executable.path, ['unroot']);
+    return true;
   } catch (err) {
     log.warn(`Unable to unroot adb daemon: '${err.message}'. Continuing`);
-    isUnrootSuccessful = false;
+    // adb 'unroot' can cause device to go offline, restart if so
+    await this.restartAdbIfDevicesOffline();
+    return false;
   }
-  // adb 'unroot' can cause device to go offline, restart if so
-  await this.restartAdbIfDevicesOffline();
-
-  return isUnrootSuccessful;
 };
 
 /**
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -399,7 +399,7 @@ systemCallMethods.shell = async function shell (cmd, opts = {}) {
   } finally {
     if (privileged && shouldRestoreUser && (!keepPrivileged || didCommandFail)) {
       try {
-        await this.adbExec(['unroot'], opts);
+        await this.unroot();
       } catch (err) {
         log.warn(`Cannot run adbd as non-root. Original error: ${err.message}`);
 
@@ -875,8 +875,7 @@ systemCallMethods.restartAdbIfDevicesOffline = async function restartAdbIfDevice
 /**
  * Switch adb server to root mode.
  * @param {boolean} unroot - Should we unroot instead of root? (default false)
- * @return {boolean} True of the switch was successful or false
- *                   if the switch failed.
+ * @return {boolean} True if the device is rooted.
  */
 systemCallMethods.root = async function root (unroot = false) {
   let cmd = unroot ? 'unroot' : 'root';
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -381,12 +381,12 @@ systemCallMethods.shell = async function shell (cmd, opts = {}) {
     log.info(`'adb shell ${cmd}' requires root access. Attempting to gain root access now.`);
     const {wasAlreadyRooted} = await this.root();
     shouldRestoreUser = !wasAlreadyRooted;
+    log.info(wasAlreadyRooted ? 'Device already had root access' : 'Root access successfully gained');
   }
   let didCommandFail = false;
   try {
     try {
       const res = await this.adbExec(_.isArray(cmd) ? ['shell', ...cmd] : ['shell', cmd], opts);
-      console.log('$$$$$$$$$res', res, JSON.stringify(res));
       return res;
     } catch (err) {
       didCommandFail = true;
@@ -897,7 +897,7 @@ systemCallMethods.changeUserPrivileges = async function changeUserPrivileges (un
     // If 'closed' was in the stderr (e.g.: `adb: unable to connect for root: closed\n"`) then the
     // device probably went offline, so restart ADB
     if (stderr.toLowerCase().includes('closed') || stderr.toLowerCase().includes('device offline')) {
-      log.warn(`${cmd} caused device to go offline. Restarting adb.`);
+      log.warn(`Attempt to 'adb ${cmd}' caused device to go offline. Restarting adb.`);
       await this.restartAdb();
     }
 
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -379,15 +379,18 @@ systemCallMethods.shell = async function shell (cmd, opts = {}) {
   let shouldRestoreUser = false;
   if (privileged) {
     log.info(`'adb shell ${cmd}' requires root access. Attempting to gain root access now.`);
-    const {wasAlreadyRooted} = await this.root();
+    const {wasAlreadyRooted, isSuccessful} = await this.root();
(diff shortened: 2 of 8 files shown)


dpgraham:
@imurchie @mykola-mokhnach this is the cause of the ADB CI flakiness:

```
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell 'ls -t -1 /data/local/tmp/appium_cache 2>&1 || echo _ERROR_''
dbug ADB Got an error 'ls: /data/local/tmp/appium_cache: No such file or directory' while getting the list of files in the cache. Assuming the cache does not exist yet
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell mkdir -p /data/local/tmp/appium_cache'
dbug ADB The count of applications in the cache: 0
info ADB Caching the application at '/Users/vsts/agent/2.149.2/work/1/s/test/fixtures/ContactManager.apk' to '/data/local/tmp/appium_cache/3b32eb4d0ff70756a89a6801cf5b5d1fe2740b4b.apk'
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell mkdir -p /data/local/tmp/appium_cache'
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 push /Users/vsts/agent/2.149.2/work/1/s/test/fixtures/ContactManager.apk /data/local/tmp/appium_cache/3b32eb4d0ff70756a89a6801cf5b5d1fe2740b4b.apk'
info ADB The upload of 'ContactManager.apk' (25.32 KB) took 0.228s
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell pm install -r /data/local/tmp/appium_cache/3b32eb4d0ff70756a89a6801cf5b5d1fe2740b4b.apk'
info ADB The installation of 'ContactManager.apk' took 4.772s
dbug ADB Install command stdout: Success
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell am start -W -n com.example.android.contactmanager/ContactManager -S'
dbug ADB We tried to start an activity that doesn't exist, retrying with '.ContactManager' activity name
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell am start -W -n com.example.android.contactmanager/.ContactManager -S'
dbug ADB Attempting to kill all com.example.android.contactmanager processes
dbug ADB Getting IDs of all 'com.example.android.contactmanager' processes
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell pgrep -f com\\.example\\.android\\.contactmanager'
dbug ADB Attempting to kill process 2910
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell kill -0 2910'
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell whoami'
info ADB Cannot kill PID 2910 due to insufficient permissions. Retrying as root
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell kill -0 2910'
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell kill 2910'
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell kill 2910'
WARN ADB Unable to unroot adb daemon: 'Command '/Users/vsts/Library/Android/sdk/platform-tools/adb unroot' exited with code 1'. Continuing
dbug ADB Getting IDs of all 'com.example.android.contactmanager' processes
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell pgrep -f com\\.example\\.android\\.contactmanager'
    ✓ killProcessesByName should kill process (17881ms)
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell 'ls -t -1 /data/local/tmp/appium_cache 2>&1 || echo _ERROR_''
dbug ADB Got an error 'Error executing adbExec. Original error: 'Command '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell 'ls -t -1 /data/local/tmp/appium_cache 2>&1 || echo _ERROR_'' exited with code 1'; Stderr: 'error: device offline'; Code: '1'' while getting the list of files in the cache. Assuming the cache does not exist yet
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell mkdir -p /data/local/tmp/appium_cache'
    2) killProcessByPID should kill process
```

The device goes offline when we retry as root and then fail to unroot.

dpgraham:
I'm going to try restarting ADB whenever `unroot` fails.

dpgraham:
This is what happens when it passes

```
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell 'ls -t -1 /data/local/tmp/appium_cache 2>&1 || echo _ERROR_''
dbug ADB The count of applications in the cache: 1
info ADB The application at '/Users/vsts/agent/2.149.2/work/1/s/test/fixtures/ContactManager.apk' is already cached to '/data/local/tmp/appium_cache/3b32eb4d0ff70756a89a6801cf5b5d1fe2740b4b.apk'
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell pm install -r /data/local/tmp/appium_cache/3b32eb4d0ff70756a89a6801cf5b5d1fe2740b4b.apk'
info ADB The installation of 'ContactManager.apk' took 2.035s
dbug ADB Install command stdout: Success
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell am start -W -n com.example.android.contactmanager/ContactManager -S'
dbug ADB We tried to start an activity that doesn't exist, retrying with '.ContactManager' activity name
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell am start -W -n com.example.android.contactmanager/.ContactManager -S'
dbug ADB Getting IDs of all 'com.example.android.contactmanager' processes
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell pgrep -f com\\.example\\.android\\.contactmanager'
dbug ADB Attempting to kill process 3341
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell kill -0 3341'
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell whoami'
info ADB Cannot kill PID 3341 due to insufficient permissions. Retrying as root
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell kill -0 3341'
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell kill 3341'
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell kill 3341'
dbug ADB Getting IDs of all 'com.example.android.contactmanager' processes
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell pgrep -f com\\.example\\.android\\.contactmanager'
    ✓ killProcessByPID should kill process (11285ms)
    - should get device language and country
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell am broadcast -a io.appium.settings.locale -n io.appium.settings/.receivers.LocaleSettingReceiver --es lang en --es country US'
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell getprop persist.sys.locale'
dbug ADB Current device property 'persist.sys.locale': 
dbug ADB Running '/Users/vsts/Library/Android/sdk/platform-tools/adb -P 5037 shell getprop ro.product.locale'
dbug ADB Current device property 'ro.product.locale': en-US
    ✓ should get device locale (20257ms)
```

Doesn't fail to unroot

lib/tools/system-calls.js
```diff
@@ -855,26 +861,41 @@ systemCallMethods.reboot = async function reboot (retries = DEFAULT_ADB_REBOOT_R
   });
 };
 
+/**
+ * Restart adb but only if no devices are connected
+ */
+systemCallMethods.restartAdbIfDevicesOffline = async function restartAdbIfDevicesOffline () {
+  const connectedDevices = await this.getConnectedDevices();
+  if (connectedDevices.length === 0) {
```
mykola-mokhnach: this will not work for parallel tests

dpgraham: What do you recommend?

lib/tools/system-calls.js
```diff
@@ -855,26 +861,41 @@ systemCallMethods.reboot = async function reboot (retries = DEFAULT_ADB_REBOOT_R
   });
 };
 
+/**
+ * Restart adb but only if no devices are connected
+ */
+systemCallMethods.restartAdbIfDevicesOffline = async function restartAdbIfDevicesOffline () {
```
mykola-mokhnach: I don't like this method in in general. Can we instead analyze the output of root/unroot command and perform the necessary actions if it contains some particular keywords/return exit codes?

lib/tools/system-calls.js
```diff
@@ -396,10 +400,12 @@ systemCallMethods.shell = async function shell (cmd, opts = {}) {
   } finally {
     if (privileged && shouldRestoreUser && (!keepPrivileged || didCommandFail)) {
       try {
-        await this.adbExec(['unroot'], opts);
+        retryInterval(10, 1000, async () => await this.adbExec(['unroot'], opts));
```
mykola-mokhnach: "blind" retry will only slow down a test. There should be a better way to distinguish the state after the command has been executed

lib/tools/system-calls.js
mykola-mokhnach: we should rely on the actual command output instead

mykola-mokhnach:
We know that adb issue only happens if the device is already rooted and we try to unroot it, so it makes sense to either analyse the unroot output/exit code and restart adb based on that or check if the current user is root before executing unroot and perform adb restart if unroot fails in such case. Also, all this logic should be encapsulated in root/unroot methods and should not be exposed to the external code

lib/tools/system-calls.js
```diff
@@ -855,25 +845,43 @@ systemCallMethods.reboot = async function reboot (retries = DEFAULT_ADB_REBOOT_R
...
 /**
  * Switch adb server to root mode.
- *
- * @return {boolean} True of the switch was successful or false
- *                   if the switch failed.
+ * @param {boolean} unroot - Should we unroot instead of root? (default false)
+ * @return {boolean} True if the device is rooted.
  */
-systemCallMethods.root = async function root () {
+systemCallMethods.root = async function root (unroot = false) {
```
mykola-mokhnach: I would rather extract the common logic into `changeUserPrivileges` method

lib/tools/system-calls.js
```diff
@@ -884,13 +892,7 @@ systemCallMethods.root = async function root () {
...
  */
 systemCallMethods.unroot = async function unroot () {
-  try {
-    await exec(this.executable.path, ['unroot']);
-    return true;
-  } catch (err) {
-    log.warn(`Unable to unroot adb daemon: '${err.message}'. Continuing`);
-    return false;
-  }
+  await this.root(true);
```
mykola-mokhnach: how about the returned value?

lib/tools/system-calls.js
```diff
@@ -855,25 +845,43 @@ systemCallMethods.reboot = async function reboot (retries = DEFAULT_ADB_REBOOT_R
...
-    return false;
+    const {stderr = ''} = err;
+    log.warn(`Unable to ${cmd} adb daemon. Original error: '${stderr}'. Continuing.`);
+    // adb 'root' can cause device to go offline, restart if so
+    if (stderr.toLowerCase().includes('closed')) {
+      log.warn(`${cmd} caused device to go offline. Restarting adb.`);
+      await this.restartAdb();
+    }
+
+    return {isRooted: unroot};
```
mykola-mokhnach: I assume both `isRooted` and `wasRooted ` should be always returned

lib/tools/system-calls.js
```diff
@@ -855,25 +845,43 @@ systemCallMethods.reboot = async function reboot (retries = DEFAULT_ADB_REBOOT_R
...
+  let wasAlreadyRooted = false;
   try {
-    let {stdout} = await exec(this.executable.path, ['root']);
+    let {stdout} = await exec(this.executable.path, [cmd]);
 
     // on real devices in some situations we get an error in the stdout
-    if (stdout && stdout.indexOf('adbd cannot run as root') !== -1) {
-      throw new Error(stdout.trim());
+    if (stdout) {
+      if (stdout.includes('adbd cannot run as root')) {
```
mykola-mokhnach: if NOT includes

lib/tools/system-calls.js
```diff
@@ -855,25 +845,43 @@ systemCallMethods.reboot = async function reboot (retries = DEFAULT_ADB_REBOOT_R
...
  * Switch adb server to root mode.
- *
- * @return {boolean} True of the switch was successful or false
- *                   if the switch failed.
+ * @param {boolean} unroot - Should we unroot instead of root? (default false)
+ * @return {boolean} True if the device is rooted.
  */
-systemCallMethods.root = async function root () {
+systemCallMethods.root = async function root (unroot = false) {
+  let cmd = unroot ? 'unroot' : 'root';
```
mykola-mokhnach: const

lib/tools/system-calls.js
mykola-mokhnach: is adb still in rooted state after restart?

test/functional/adb-commands-e2e-specs.js
```diff
@@ -48,7 +48,7 @@ describe('adb commands', function () {
   });
   it('defaultIME should get default IME', async function () {
     const defaultIME = await adb.defaultIME();
-    if (defaultIME) {
+    if (defaultIME && defaultIME !== 'null') {
```
mykola-mokhnach: we should probably mention that `defaultIME` can return `null` strings in the docstring

lib/tools/system-calls.js
dpgraham: This is what it was before:

```js
if (stdout && stdout.indexOf('adbd cannot run as root') !== -1) {
  throw new Error(stdout.trim());
}
```

`stdout.indexOf('adbd cannot run as root') !== -1` is equivalent to `stdout.includes('adbd cannot run as root')`

I think we want to throw the error if we get that message

lib/tools/system-calls.js
dpgraham: Should we add the `prefer-const` ESLint rule? I'm definitely in favor of _always_ using `const` unless you need to re-assign a variable.

lib/tools/system-calls.js
dpgraham: I changed it to: `isSuccessful` to indicate that the shell to `root/unroot` was successful or not and `wasAlreadyRooted` tells you if it was rooted already.

lib/tools/system-calls.js
dpgraham: Yep.... I really wish we used Typescript.

lib/tools/system-calls.js
dpgraham: nvm I see what you mean now....

I fixed it.

lib/tools/system-calls.js
```diff
@@ -835,14 +831,30 @@ systemCallMethods.waitForDevice = async function waitForDevice (appDeviceReadyTi
...
+    await this.shell(['stop']);
+    await B.delay(2000); // let the emu finish stopping;
+    await this.setDeviceProperty('sys.boot_completed', 0, false);
+    await this.shell(['start']);
+
+    // Return root state to what it was before
+    if (!wasAlreadyRooted) {
+      await this.unroot();
+    }
+  } catch (e) {
```
dpgraham: ^ @mykola-mokhnach I changed reboot logic to not be 'privileged' because it was causing a root+unroot on the 3 different shell calls. Opted for just rooting before the calls and then unrooting after.

lib/tools/system-calls.js
mykola-mokhnach: I think it would be server to remove this exception at all and only log a warning

lib/tools/adb-commands.js
```diff
@@ -483,6 +483,9 @@ methods.setIME = async function setIME (imeId) {
 methods.defaultIME = async function defaultIME () {
   try {
     let engine = await this.getSetting('secure', 'default_input_method');
+    if (engine === 'null') {
```
mykola-mokhnach: please update the `return` description

lib/tools/adb-commands.js
```diff
@@ -1301,7 +1304,8 @@ methods.killProcessByPID = async function killProcessByPID (pid) {
       }
       log.info(`Cannot kill PID ${pid} due to insufficient permissions. Retrying as root`);
       try {
-        becameRoot = await this.root();
+        let {isSuccessful} = await this.root();
```
mykola-mokhnach: const

lib/tools/adb-commands.js
```diff
@@ -1462,13 +1466,14 @@ methods.getDeviceProperty = async function getDeviceProperty (property) {
  * @param {string} property - The name of the property. This name should
  *                            be known to _adb shell setprop_ tool.
  * @param {string} val - The new property value.
+ * @param {boolean} privileged - Do we need to set root access for call to 'setProp'?
  *
  * @throws {error} If _setprop_ utility fails to change property value.
  */
-methods.setDeviceProperty = async function setDeviceProperty (prop, val) {
+methods.setDeviceProperty = async function setDeviceProperty (prop, val, privileged = true) {
```
mykola-mokhnach: it makes sense to add a generic `opts` argument and put `privileged` there. It might be we'd need to add more parameters there

lib/tools/adb-commands.js
dpgraham: That doesn't work, it can't override the previous definition of `let becameRoot = false`.

lib/tools/system-calls.js
```diff
@@ -835,14 +836,30 @@ systemCallMethods.waitForDevice = async function waitForDevice (appDeviceReadyTi
...
+
+    // Stop and re-start the device
+    await this.shell(['stop']);
+    await B.delay(2000); // let the emu finish stopping;
+    await this.setDeviceProperty('sys.boot_completed', 0, false);
+    await this.shell(['start']);
+
+    // Return root state to what it was before
+    if (!wasAlreadyRooted) {
+      await this.unroot();
```
mykola-mokhnach: this should be in finally

lib/tools/system-calls.js
```diff
@@ -855,42 +872,65 @@ systemCallMethods.reboot = async function reboot (retries = DEFAULT_ADB_REBOOT_R
...
+ * @property {boolean} isSuccessful True if the call to root/unroot was successful
+ * @property {boolean} wasAlreadyRooted True if the device was already rooted
+ */
+
 /**
  * Switch adb server to root mode.
- *
- * @return {boolean} True of the switch was successful or false
- *                   if the switch failed.
+ * @param {boolean} unroot - Should we unroot instead of root? (default false)
```
mykola-mokhnach: how about isElevated arg name?

lib/tools/system-calls.js
```diff
@@ -855,42 +872,65 @@ systemCallMethods.reboot = async function reboot (retries = DEFAULT_ADB_REBOOT_R
...
+
+    // Check the output of the stdErr to see if there's any clues that show that the device went offline
+    // and if it did go offline, restart ADB
+    const s = stderr.toLowerCase();
+    if (s.includes('closed') || s.includes('device offline')) {
+      log.warn(`Attempt to 'adb ${cmd}' caused device to go offline. Restarting adb.`);
+      await this.restartAdb();
+    }
+
+    return {isSuccessful: false, wasAlreadyRooted};
```
mykola-mokhnach: `wasAlreadyRooted ` may return wrong result here. How about checking isRoot at the beginning of this method. also, we could skip the rest of the logic (and save time), if the user already had the necessary permissions

lib/tools/adb-commands.js
```diff
@@ -478,11 +478,14 @@ methods.setIME = async function setIME (imeId) {
 /**
  * Get the default input method on the device under test.
  *
- * @return {string} The name of the default input method.
+ * @return {string} The name of the default input method or null if none found.
```
mykola-mokhnach: `{?string}`

lib/tools/adb-commands.js
```diff
@@ -1456,19 +1460,26 @@ methods.getDeviceProperty = async function getDeviceProperty (property) {
   return val;
 };
 
+/**
+ * @typedef setPropOpts
```
mykola-mokhnach: `@typedef {Object}`

lib/tools/system-calls.js
```diff
@@ -835,14 +836,31 @@ systemCallMethods.waitForDevice = async function waitForDevice (appDeviceReadyTi
...
-  });
-  await B.delay(2000); // let the emu finish stopping;
-  await this.setDeviceProperty('sys.boot_completed', 0);
-  await this.shell(['start'], {
-    privileged: true,
-  });
+  let wasAlreadyRooted;
+  try {
+    // Get root access so we can run the next shell commands which require root access
+    wasAlreadyRooted = (await this.root()).wasAlreadyRooted;
```
mykola-mokhnach: you can move it out of `try` block, since root call is now safe

lib/tools/system-calls.js
```diff
@@ -855,42 +873,65 @@ systemCallMethods.reboot = async function reboot (retries = DEFAULT_ADB_REBOOT_R
...
-    let {stdout} = await exec(this.executable.path, ['root']);
+    let {stdout} = await exec(this.executable.path, [cmd]);
 
     // on real devices in some situations we get an error in the stdout
-    if (stdout && stdout.indexOf('adbd cannot run as root') !== -1) {
-      throw new Error(stdout.trim());
+    if (stdout) {
+      if (stdout.includes('adbd cannot run as root')) {
+        log.warn('Cannot run device as root');
+      }
```
mykola-mokhnach: we can return here, since we know the result

lib/tools/system-calls.js
```diff
@@ -922,6 +921,10 @@ systemCallMethods.changeUserPrivileges = async function changeUserPrivileges (is
  * @return {rootResult}
  */
 systemCallMethods.root = async function root () {
+  // If it's already rooted, our job is done. No need to root it again.
+  if (await this.isRoot()) {
```
mykola-mokhnach: I assume this call should be inside `changeUserPrivilege` method. So, if isRoot is true and isElevated true or !isRoot and !isElevated we should do a short circuit. Also, this will fix the initial value of wasAlreadyRooted

lib/tools/adb-commands.js
```diff
@@ -1456,19 +1460,26 @@ methods.getDeviceProperty = async function getDeviceProperty (property) {
...
  * @param {string} property - The name of the property. This name should
  *                            be known to _adb shell setprop_ tool.
  * @param {string} val - The new property value.
+ * @param {setPropOpts} opts
  *
  * @throws {error} If _setprop_ utility fails to change property value.
  */
-methods.setDeviceProperty = async function setDeviceProperty (prop, val) {
+methods.setDeviceProperty = async function setDeviceProperty (prop, val, opts = {}) {
+  const {privileged} = opts;
```
mykola-mokhnach: privileged = true

lib/tools/system-calls.js
dpgraham: I could do that... I"ll have to pick it up again tomorrow, because when I tried that way a broke a bunch of unit tests.

dpgraham: Never mind... I fixed it now. It was easy.

dpgraham:
Changes pushed and CI passing (Azure at least). Any more comments will be addressed tomorrow.

lib/tools/adb-commands.js
```diff
@@ -1456,19 +1460,26 @@ methods.getDeviceProperty = async function getDeviceProperty (property) {
   return val;
 };
 
+/**
+ * @typedef {object} setPropOpts
+ * @property {boolean} privileged - Do we run setProp as a privileged command?
```
mykola-mokhnach: true by default

lib/tools/system-calls.js
```diff
@@ -394,12 +396,11 @@ systemCallMethods.shell = async function shell (cmd, opts = {}) {
...
     }
   } finally {
+    // Return the 'root' state to what it was before 'shell' was called
     if (privileged && shouldRestoreUser && (!keepPrivileged || didCommandFail)) {
-      try {
-        await this.adbExec(['unroot'], opts);
-      } catch (err) {
-        log.warn(`Cannot run adbd as non-root. Original error: ${err.message}`);
-      }
+      log.info(`Call to '${cmd}' completed. Returning device to original unrooted state.`);
```
mykola-mokhnach: this log line seems redundant to me

lib/tools/system-calls.js
```diff
@@ -374,15 +374,17 @@ systemCallMethods.shell = async function shell (cmd, opts = {}) {
...
   let shouldRestoreUser = false;
   if (privileged) {
-    try {
-      shouldRestoreUser = !(await this.adbExec(['root'], opts)).includes('already running as root');
-    } catch (err) {
-      if (!err.message.includes('adbd cannot run as root')) {
-        // Do not show the warning for real devices, where root is locked
-        log.warn(`Cannot run adbd as root. Original error: ${err.message}`);
-      }
+    log.info(`'adb shell ${cmd}' requires root access. Attempting to gain root access now.`);
```
mykola-mokhnach: log.debug

lib/tools/system-calls.js
```diff
@@ -855,42 +871,73 @@ systemCallMethods.reboot = async function reboot (retries = DEFAULT_ADB_REBOOT_R
...
+systemCallMethods.changeUserPrivileges = async function changeUserPrivileges (isElevated) {
+  const cmd = isElevated ? 'root' : 'unroot';
+
+  // If it's already rooted, our job is done. No need to root it again.
+  const isRoot = await this.isRoot();
+  if ((isRoot && isElevated) || (!isRoot && !isElevated)) {
+    return {isSuccessful: true, wasAlreadyRooted: isRoot};
+  }
+
+  let wasAlreadyRooted = false;
```
mykola-mokhnach: = isRoot

lib/tools/system-calls.js
```diff
@@ -855,42 +871,73 @@ systemCallMethods.reboot = async function reboot (retries = DEFAULT_ADB_REBOOT_R
...
   try {
-    let {stdout} = await exec(this.executable.path, ['root']);
+    let {stdout} = await exec(this.executable.path, [cmd]);
 
     // on real devices in some situations we get an error in the stdout
-    if (stdout && stdout.indexOf('adbd cannot run as root') !== -1) {
-      throw new Error(stdout.trim());
+    if (stdout) {
+      if (stdout.includes('adbd cannot run as root')) {
+        log.warn('Cannot run device as root');
```
mykola-mokhnach: I think we can skip this log at all otherwise it will appear too often on real devices

lib/tools/system-calls.js
```diff
@@ -855,42 +871,73 @@ systemCallMethods.reboot = async function reboot (retries = DEFAULT_ADB_REBOOT_R
...
+ * @return {rootResult}
  */
-systemCallMethods.root = async function root () {
+systemCallMethods.changeUserPrivileges = async function changeUserPrivileges (isElevated) {
+  const cmd = isElevated ? 'root' : 'unroot';
+
+  // If it's already rooted, our job is done. No need to root it again.
+  const isRoot = await this.isRoot();
+  if ((isRoot && isElevated) || (!isRoot && !isElevated)) {
+    return {isSuccessful: true, wasAlreadyRooted: isRoot};
```
mykola-mokhnach: 👍

lib/tools/system-calls.js
```diff
@@ -855,42 +871,73 @@ systemCallMethods.reboot = async function reboot (retries = DEFAULT_ADB_REBOOT_R
...
   } catch (err) {
-    log.warn(`Unable to root adb daemon: '${err.message}'. Continuing`);
-    return false;
+    const {stderr = '', message} = err;
+    log.warn(`Unable to ${cmd} adb daemon. Original error: '${message}'. Stderr: '${stderr}'. Continuing.`);
+
+    // Check the output of the stdErr to see if there's any clues that show that the device went offline
+    // and if it did go offline, restart ADB
+    const s = stderr.toLowerCase();
+    if (s.includes('closed') || s.includes('device offline')) {
```
mykola-mokhnach: ['closed', 'device offline'].some((x) => stderr.toLowerCase().includes(x))

lib/tools/system-calls.js
```diff
@@ -855,42 +871,73 @@ systemCallMethods.reboot = async function reboot (retries = DEFAULT_ADB_REBOOT_R
...
 };
 
+/**
+ * @typedef {Object} rootResult
+ * @property {boolean} isSuccessful True if the call to root/unroot was successful
+ * @property {boolean} wasAlreadyRooted True if the device was already rooted
+ */
+
 /**
  * Switch adb server to root mode.
```
mykola-mokhnach: Please update the description

lib/tools/adb-commands.js
```diff
@@ -1301,7 +1304,8 @@ methods.killProcessByPID = async function killProcessByPID (pid) {
       }
       log.info(`Cannot kill PID ${pid} due to insufficient permissions. Retrying as root`);
       try {
```
mykola-mokhnach: try block is now redundant, since root call is not going to throw an exception

lib/tools/system-calls.js
```diff
@@ -394,12 +396,11 @@ systemCallMethods.shell = async function shell (cmd, opts = {}) {
...
+    // Return the 'root' state to what it was before 'shell' was called
     if (privileged && shouldRestoreUser && (!keepPrivileged || didCommandFail)) {
-      try {
-        await this.adbExec(['unroot'], opts);
-      } catch (err) {
-        log.warn(`Cannot run adbd as non-root. Original error: ${err.message}`);
-      }
+      log.info(`Call to '${cmd}' completed. Returning device to original unrooted state.`);
+      const {isSuccessful} = await this.unroot();
+      log.info(isSuccessful ? 'Returned device to unrooted state' : 'Could not return device to unrooted state');
```
mykola-mokhnach: log.debug

lib/tools/system-calls.js
```diff
@@ -835,14 +836,29 @@ systemCallMethods.waitForDevice = async function waitForDevice (appDeviceReadyTi
...
-  await this.shell(['start'], {
-    privileged: true,
-  });
+  // Get root access so we can run the next shell commands which require root access
+  const { wasAlreadyRooted } = await this.root();
+  try {
+    // Stop and re-start the device
+    await this.shell(['stop']);
+    await B.delay(2000); // let the emu finish stopping;
+    await this.setDeviceProperty('sys.boot_completed', 0, {privileged: false});
```
mykola-mokhnach: a comment about why privileged is set to false would be useful here

PR #447, opened 2019-04-22
[wip] Fixes to support Q

dpgraham:
The biggest problem is the permission requesting APK that comes up before an app opens. Right now I set `grantPermissions` to true in the tests to prevent this.

For the drivers (Espresso, UiAutomator2, etc...) we'll either need to grant permissions by default and deprecate `autoGrantPermissions`; or we'll have to find some alternative way of getting through the permissions screen.

* Temporary workaround to return platform version as 10 if platformVersion == 'Q'
* Unrelated to Q.... memoize getApiLevel
* Install io.appium.settings before tests to make test work
* Add 10 to the API_LEVEL_MAP

### DIFF
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -78,28 +78,26 @@ methods.initBundletool = async function initBundletool () {
  *                  Android Lollipop. The result of this method is cached, so all the further
  * calls return the same value as the first one.
  */
-methods.getApiLevel = async function getApiLevel () {
-  if (!_.isInteger(this._apiLevel)) {
-    try {
-      const strOutput = await this.getDeviceProperty('ro.build.version.sdk');
-      let apiLevel = parseInt(strOutput.trim(), 10);
+methods.getApiLevel = _.memoize(async function getApiLevel () {
+  try {
+    const strOutput = await this.getDeviceProperty('ro.build.version.sdk');
+    let apiLevel = parseInt(strOutput.trim(), 10);
 
-      // Temp workaround. Android Q beta emulators report SDK 28 when they should be 29
-      if (apiLevel === 28 && (await this.getDeviceProperty('ro.build.version.release')).toLowerCase() === 'q') {
-        log.debug('Release version is Q but found API Level 28. Setting API Level to 29');
-        apiLevel = 29;
-      }
-      this._apiLevel = apiLevel;
-      log.debug(`Device API level: ${this._apiLevel}`);
-      if (isNaN(this._apiLevel)) {
-        throw new Error(`The actual output '${strOutput}' cannot be converted to an integer`);
-      }
-    } catch (e) {
-      throw new Error(`Error getting device API level. Original error: ${e.message}`);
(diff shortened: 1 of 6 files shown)


lib/tools/adb-commands.js
```diff
@@ -78,28 +78,26 @@ methods.initBundletool = async function initBundletool () {
...
-        apiLevel = 29;
-      }
-      this._apiLevel = apiLevel;
-      log.debug(`Device API level: ${this._apiLevel}`);
-      if (isNaN(this._apiLevel)) {
-        throw new Error(`The actual output '${strOutput}' cannot be converted to an integer`);
-      }
-    } catch (e) {
-      throw new Error(`Error getting device API level. Original error: ${e.message}`);
+methods.getApiLevel = _.memoize(async function getApiLevel () {
```
mykola-mokhnach: I don't think this is a good idea. We must do memoization based on ADB instance rather than statically based. Why you don't like the previous approach

test/functional/apk-utils-e2e-specs.js
```diff
@@ -69,7 +69,7 @@ describe('apk utils', function () {
   });
   describe('startApp', function () {
     it('should be able to start', async function () {
-      await adb.install(contactManagerPath);
+      await adb.install(contactManagerPath, {grantPermissions: true}, {grantPermissions: true});
```
mykola-mokhnach: why twice?

test/functional/adb-commands-e2e-specs.js
```diff
@@ -32,6 +32,10 @@ describe('adb commands', function () {
   const androidInstallTimeout = 90000;
   before(async function () {
     adb = await ADB.createADB({ adbExecTimeout: 60000 });
+    await adb.install(path.resolve('node_modules', 'io.appium.settings', 'apks', 'settings_apk-debug.apk'));
```
mykola-mokhnach: path to appium settings can be a constant

test/functional/setup.js
```diff
@@ -12,6 +12,7 @@ const API_LEVEL_MAP = {
   8.0: '26',
   8.1: '27',
   9: '28',
+  10: '29',
```
mykola-mokhnach: is there a way to have this list autofilled?

lib/tools/adb-commands.js
dpgraham: The previous approach was fine I can change it back. I just liked using memoize because it better expresses what the code is doing.

test/functional/apk-utils-e2e-specs.js
dpgraham: Redundant copy and replace. I'm going to clean up this code some more today I just wanted to see test results from putting in a PR

test/functional/setup.js
dpgraham: Maybe. I could see if there's a JSON endpoint that defines this.

imurchie: You could use https://www.npmjs.com/package/android-versions, but it'd need a PR for Q.

PR #450, opened 2019-06-07
Use ANDROID_SDK_ROOT along with ANDROID_HOME

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/12735

### DIFF
--- a/lib/adb.js
+++ b/lib/adb.js
@@ -33,8 +33,8 @@ const DEFAULT_OPTS = {
 
 class ADB {
   constructor (opts = {}) {
-    if (typeof opts.sdkRoot === 'undefined') {
-      opts.sdkRoot = process.env.ANDROID_HOME || '';
+    if (_.isUndefined(opts.sdkRoot)) {
+      opts.sdkRoot = process.env.ANDROID_HOME || process.env.ANDROID_SDK_ROOT || '';
     }
 
     Object.assign(this, opts);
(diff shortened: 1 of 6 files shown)


lib/adb.js
```diff
@@ -33,8 +33,8 @@ const DEFAULT_OPTS = {
 
 class ADB {
   constructor (opts = {}) {
-    if (typeof opts.sdkRoot === 'undefined') {
-      opts.sdkRoot = process.env.ANDROID_HOME || '';
+    if (_.isUndefined(opts.sdkRoot)) {
+      opts.sdkRoot = process.env.ANDROID_HOME || process.env.ANDROID_SDK_ROOT || '';
```
KazuCocoa: Is it better to respect `ANDROID_SDK_ROOT` than `ANDROID_HOME` to follow Google?

mykola-mokhnach: > Sets the path to the SDK installation directory. Once set, the value does not typically change, and can be shared by multiple users on the same machine. ANDROID_HOME, which also points to the SDK installation directory, is deprecated. If you continue to use it, the following rules apply:
> If ANDROID_HOME is defined and contains a valid SDK installation, its value is used instead of the value in ANDROID_SDK_ROOT.
> If ANDROID_HOME is not defined, the value in ANDROID_SDK_ROOT is used.
> If ANDROID_HOME is defined but does not exist or does not contain a valid SDK installation, the value in ANDROID_SDK_ROOT is used instead.

KazuCocoa: Ah, understood

PR #455, opened 2019-06-15
Warn install failed due to INSTALL_FAILED_TEST_ONLY

vrunoa:
The idea is too add more visibility to the testers when adb fails to this by explicitly saying 
`Set allowTestPackages capability to true in order to allow test packages installation.` in the logs. 

Work in progress, I want to add another unit test to check this.

### DIFF
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -525,7 +525,9 @@ apkUtilsMethods.install = async function install (appPath, options = {}) {
     log.debug(`Install command stdout: ${truncatedOutput}`);
     if (/\[INSTALL[A-Z_]+FAILED[A-Z_]+\]/.test(output)) {
       if (this.isTestPackageOnlyError(output)) {
-        log.warn('Set `allowTestPackages` capability to true in order to allow test packages installation.');
+        const msg = `Set 'allowTestPackages' capability to true in order to allow test packages installation.`;
+        log.warn(msg);
+        throw new Error(`${output}\n${msg}`);
       }
       throw new Error(output);
     }
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -524,8 +524,8 @@ apkUtilsMethods.install = async function install (appPath, options = {}) {
       output : `${output.substr(0, 150)}...${output.substr(output.length - 150)}`;
     log.debug(`Install command stdout: ${truncatedOutput}`);
     if (/\[INSTALL[A-Z_]+FAILED[A-Z_]+\]/.test(output)) {
-      if (/\[INSTALL_FAILED_TEST_ONLY\]/.test(output)) {
-        log.error('Set `allowTestPackages` capability to true in order to allow test packages installation.');
+      if (this.isTestPackageOnlyError(output)) {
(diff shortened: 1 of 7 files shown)


lib/tools/apk-utils.js
```diff
@@ -524,6 +524,9 @@ apkUtilsMethods.install = async function install (appPath, options = {}) {
       output : `${output.substr(0, 150)}...${output.substr(output.length - 150)}`;
     log.debug(`Install command stdout: ${truncatedOutput}`);
     if (/\[INSTALL[A-Z_]+FAILED[A-Z_]+\]/.test(output)) {
+      if (/\[INSTALL_FAILED_TEST_ONLY\]/.test(output)) {
+        log.error('Set `allowTestPackages` capability to true in order to allow test packages installation.');
```
mykola-mokhnach: this is not an error. Also, it makes since to include this hint into the error description above as not all users are checking server logs

mykola-mokhnach:
I assume the `enableIME and disableIME should enable  and disable IME` test does not work for API 28+. thus it makes sense to skip it for newer api versions

lib/tools/apk-utils.js
KazuCocoa: sounds good to add it in error message

vrunoa: @mykola-mokhnach I'll change it to a `warning`.

lib/tools/apk-utils.js
```diff
@@ -524,6 +524,9 @@ apkUtilsMethods.install = async function install (appPath, options = {}) {
       output : `${output.substr(0, 150)}...${output.substr(output.length - 150)}`;
     log.debug(`Install command stdout: ${truncatedOutput}`);
     if (/\[INSTALL[A-Z_]+FAILED[A-Z_]+\]/.test(output)) {
+      if (this.isTestPackageOnlyError(output)) {
+        log.warn('Set `allowTestPackages` capability to true in order to allow test packages installation.');
+      }
       throw new Error(output);
```
mykola-mokhnach: please add the hint to the actual error description here

vrunoa: @mykola-mokhnach you mean a link to the capability? or our documentation?

mykola-mokhnach: I mean 

```
if (this.isTestPackageOnlyError(output)) {
        const msg = `Set 'allowTestPackages' capability to true in order to allow test packages installation.`;
        log.warn(msg);
        throw new Error(`${output}\n${msg}`);
      }
```

test/functional/setup.js
```diff
@@ -21,6 +21,9 @@ let apiLevel = process.env.API_LEVEL ||
                API_LEVEL_MAP[parseFloat(platformVersion)];
 apiLevel = parseInt(apiLevel, 10);
 
+// eslint-disable-next-line no-console
+console.log('apiLevel vruno', apiLevel);
```
mykola-mokhnach: this should be removed

vrunoa: 🤦‍♂

mykola-mokhnach:
@vrunoa I think the problem with 5.1 emulator might be the default storage space size. See https://stackoverflow.com/questions/5153515/how-do-i-solve-the-install-failed-dexopt-error for more details.

PR #464, opened 2019-07-16
modify apk analyzer query to find launch activity

donmezburak:
Apkanalyzer parse can give wrong launch activity name when more than MAIN and LAUNCHER activities exist and they have 'enabled' attribute.

case:
`<activity-alias>
     ... 
    android:enabled='true'
    ...><MAIN> <LAUNCHER>
  />   
<activity-alias>
     ... 
    android:enabled='false'
    ...><MAIN> <LAUNCHER>
  />`
 
To avoid this case, parser must check enabled tag of activity. Firstly, i checked whether 'enabled' attribute exists or not. If not exist, it is ok to take as launchable activity. If exist, i checked the whether 'enabled' is set to true or not.


mykola-mokhnach:
Thanks, the change makes a lot of sense.

PR #465, opened 2019-08-01
Prefer apkanalyzer for package information retrieval

mykola-mokhnach:
Google has completely deprecated aapt tool since SDK 28 and removed it from build tools, so we need to start shifting to aapt2/apkanalyzer usage.


KazuCocoa:
👍

PR #466, opened 2019-08-01
Update manifest helpers to use aapt2

mykola-mokhnach:
The only method which is left (still uses the deprecated aapt) is the resources extractor

### DIFF
--- a/lib/tools/android-manifest.js
+++ b/lib/tools/android-manifest.js
@@ -3,7 +3,7 @@ import log from '../logger.js';
 import {
   getAndroidPlatformAndPath, unzipFile,
   getApkanalyzerForOs, APKS_EXTENSION, parseManifest } from '../helpers.js';
-import { fs, zip } from 'appium-support';
+import { fs, zip, tempDir } from 'appium-support';
 import _ from 'lodash';
 import path from 'path';
 import { quote } from 'shell-quote';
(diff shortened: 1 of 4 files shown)


lib/tools/android-manifest.js
```diff
@@ -276,47 +319,35 @@ manifestMethods.insertManifest = async function insertManifest (manifest, srcApk
...
-  }
-};
-
-/**
- * Prints out the manifest extracted from the apk
- *
- * @param {string} appPath - The full path to application package.
- * @param {?string} logLevel - The level at which to log. E.g., 'debug'
- */
-manifestMethods.printManifestFromApk = async function printManifestFromApk (appPath, logLevel = 'debug') {
```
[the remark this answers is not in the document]
mykola-mokhnach: yes, there is no usage of this method across appium org.

lib/tools/android-manifest.js
```diff
@@ -218,23 +218,45 @@ manifestMethods.compileManifest = async function compileManifest (manifest, mani
...
+    const args = [
+      'link',
+      '-o', resultPath,
+      '--manifest', manifest,
+      '--rename-manifest-package', manifestPackage,
+      '--rename-instrumentation-target-package', targetPackage,
+      '-I', androidJarPath,
+      '-v',
+    ];
+    log.debug(`Compiling the manifest using ${this.binaries.aapt2} ${quote(args)}`);
```
imurchie: For clarity of the command, I think there should be single quotes around the whole command (like https://github.com/appium/appium-adb/blob/c36f4eb1c6cf7672c28005b2edbdc36d7e6f4e37/lib/tools/system-calls.js#L316).

mykola-mokhnach: changed

PR #469, opened 2019-09-03
chore: Refactor the implementation of getConnectedDevices helper

mykola-mokhnach:


lib/tools/system-calls.js
```diff
@@ -153,34 +153,38 @@ systemCallMethods.getBinaryFromPath = async function getBinaryFromPath (binaryNa
...
   }
+  const listHeader = 'List of devices';
+  // expecting adb devices to return output as
+  // List of devices attached
+  // emulator-5554	device
+  const startingIndex = stdout.indexOf(listHeader);
+  if (startingIndex < 0) {
+    throw new Error(`Unexpected output while trying to get devices. Output was: ${stdout}`);
+  }
+  // slicing ouput we care about
```
imurchie: output

lib/tools/system-calls.js
```diff
@@ -153,34 +153,38 @@ systemCallMethods.getBinaryFromPath = async function getBinaryFromPath (binaryNa
...
+  stdout = stdout.slice(startingIndex);
+  const excludedLines = [listHeader, 'adb server', '* daemon', 'offline'];
+  const devices = stdout.split('\n')
+    .map((line) => line.trim())
```
imurchie: this line can also just be `.map(_.trim)`

lib/tools/system-calls.js
mykola-mokhnach: fixed

lib/tools/system-calls.js
mykola-mokhnach: changed

PR #473, opened 2019-11-15
feat: Add a possibility to retrieve the clipboard content on Android 10

mykola-mokhnach:
The part of https://github.com/appium/appium/issues/13466 implementation

### DIFF
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -25,9 +25,23 @@ const LOCALE_SETTING_ACTION = `${SETTINGS_HELPER_ID}.locale`;
 const LOCATION_SERVICE = `${SETTINGS_HELPER_ID}/.LocationService`;
 const LOCATION_RECEIVER = `${SETTINGS_HELPER_ID}/.receivers.LocationInfoReceiver`;
 const LOCATION_RETRIEVAL_ACTION = `${SETTINGS_HELPER_ID}.location`;
+const CLIPBOARD_RECEIVER = `${SETTINGS_HELPER_ID}/.receivers.ClipboardReceiver`;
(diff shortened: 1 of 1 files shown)


lib/tools/adb-commands.js
```diff
@@ -1765,4 +1786,42 @@ methods.getTimeZone = async function getTimeZone () {
...
+    throw new Error(`Cannot retrieve the current clipboard content from the device. ` +
+      `Make sure the Appium Settings application is up to date. ` +
+      `Original error: ${err.message}`);
+  }
+
+  const match = /data="([^"]*)"/.exec(output);
+  if (!match) {
+    throw new Error(`Cannot parse the actual cliboard content from the command output: ${output}`);
+  }
+  return Buffer.from(_.trim(match[1]), 'base64').toString('utf8');
```
KazuCocoa: I noticed here should be pass without decoding since the endpoint expects Base 64 encoded string.


```
# Current
[17] pry(#<AppiumLibCoreTest::Android::DeviceTest>)> c = @driver.get_clipboard
=> "\x85\xAAi\xCA\xD7\xAC\xB6)\xE0"

# Here returns `_.trim(match[1]`
[18] pry(#<AppiumLibCoreTest::Android::DeviceTest>)> @driver = @@core.start_driver
=> #<Appium::Core::Base::Driver:0x..fee76647dbb527b8e browser="unknown">
[19] pry(#<AppiumLibCoreTest::Android::DeviceTest>)> c = @driver.get_clipboard
=> "happy testing"
```

mykola-mokhnach: true, will fix

thanks for checking it

PR #475, opened 2019-11-26
Allow starting an app with just intent

vikki:
Hi there,

I raised [an issue](https://github.com/appium/appium/issues/13628) around starting apps with just intents and I think currently it's not possible to launch an android app with just an intent in the desired capabilities; this PR should allow you to do that. I know it's possible to do it with a `mobile:deeplink` command at the moment, but if I want to run this against multiple configs where some can be launched directly with a pkg/activity, then I have to special case the code, and it would be great to keep that encapsulated in the desired capabilities.

I've just used what was already there, but changed the validation to allow you to call `startApp` with `intentAction` and `optionalIntentArgs` even when there isn't an activity provided. 

This works for me (though you do have to explicitly set the category for it to work), and allows me to launch my expo app with just the intent (previously `adb` would use the package and the intent ends up getting ignored as far as I can tell). 

I did want to flag that I'm not sure the error messaging when the app doesn't start correctly is right;  I'm happy to fix that if you could provide a little bit of guidance on how to change it, assuming the rest seems OK.

### DIFF
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -78,8 +78,11 @@ apkUtilsMethods.startUri = async function startUri (uri, pkg) {
 
 /**
  * @typedef {Object} StartAppOptions
- * @property {!string} activity - The name of the main application activity
  * @property {!string} pkg - The name of the application package
+ * @property {?string} activity - The name of the main application activity.
+ * This or action is required in order to be able to launch an app.
+ * @property {?string} action - The name of the intent action that will launch the required app.
+ * This or activity is required in order to be able to launch an app.
  * @property {?boolean} retry [true] - If this property is set to `true`
  * and the activity name does not start with '.' then the method
  * will try to add the missing dot and start the activity once more
@@ -107,11 +110,8 @@ apkUtilsMethods.startUri = async function startUri (uri, pkg) {
  * @throws {Error} If there is an error while executing the activity
  */
 apkUtilsMethods.startApp = async function startApp (startAppOptions = {}) {
-  const hasAppAndPkg = startAppOptions.activity && startAppOptions.pkg;
-  const hasIntent = startAppOptions.action;
-
-  if (!hasAppAndPkg && !hasIntent) {
-    throw new Error('activity and pkg, or intent, are required to start an application');
+  if (!startAppOptions.pkg || !(startAppOptions.activity || startAppOptions.action)) {
+    throw new Error('pkg, and activity or intent action, are required to start an application');
   }
 
   startAppOptions = _.clone(startAppOptions);
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -131,6 +131,7 @@ apkUtilsMethods.startApp = async function startApp (startAppOptions = {}) {
 
   const apiLevel = await this.getApiLevel();
   const cmd = buildStartCmd(startAppOptions, apiLevel);
+  const intentName = `${startAppOptions.action}${startAppOptions.optionalIntentArguments ? ' ' + startAppOptions.optionalIntentArguments : ''}`;
   try {
     const shellOpts = {};
     if (_.isInteger(startAppOptions.waitDuration) && startAppOptions.waitDuration > 20000) {
@@ -148,10 +149,6 @@ apkUtilsMethods.startApp = async function startApp (startAppOptions = {}) {
       throw new Error(`Activity name '${startAppOptions.activity}' used to start the app doesn't ` +
                       `exist or cannot be launched! Make sure it exists and is a launchable activity`);
     } else if (stdout.includes('Error: Intent does not match any activities') || stdout.includes('Error: Activity not started, unable to resolve Intent')) {
-      let intentName = startAppOptions.action;
-      if (startAppOptions.optionalIntentArguments) {
-        intentName += ` ${startAppOptions.optionalIntentArguments}`;
-      }
       throw new Error(`Activity for intent '${intentName}' used to start the app doesn't ` +
                       `exist or cannot be launched! Make sure it exists and is a launchable activity`);
     } else if (stdout.includes('java.lang.SecurityException')) {
@@ -164,7 +161,7 @@ apkUtilsMethods.startApp = async function startApp (startAppOptions = {}) {
     }
     return stdout;
   } catch (e) {
-    const appDescriptor = startAppOptions.pkg || `${startAppOptions.action} ${startAppOptions.optionalIntentArguments}`;
+    const appDescriptor = startAppOptions.pkg || intentName;
     throw new Error(`Cannot start the '${appDescriptor}' application. ` +
       `Visit ${ACTIVITIES_TROUBLESHOOTING_LINK} for troubleshooting. ` +
       `Original error: ${e.message}`);
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -108,7 +108,7 @@ apkUtilsMethods.startUri = async function startUri (uri, pkg) {
  */
 apkUtilsMethods.startApp = async function startApp (startAppOptions = {}) {
   const hasAppAndPkg = startAppOptions.activity && startAppOptions.pkg;
-  const hasIntent = startAppOptions.optionalIntentArguments && startAppOptions.action;
+  const hasIntent = startAppOptions.action;
 
   if (!hasAppAndPkg && !hasIntent) {
     throw new Error('activity and pkg, or intent, are required to start an application');
@@ -146,6 +146,9 @@ apkUtilsMethods.startApp = async function startApp (startAppOptions = {}) {
         return await this.startApp(startAppOptions);
       }
       throw new Error(`Activity name '${startAppOptions.activity}' used to start the app doesn't ` +
+                      `exist or cannot be launched! Make sure it exists and is a launchable activity`);
+    } else if (stdout.includes('Error: Intent does not match any activities')) {
+      throw new Error(`Activity for intent '${startAppOptions.action} ${startAppOptions.optionalIntentArguments}' used to start the app doesn't ` +
                       `exist or cannot be launched! Make sure it exists and is a launchable activity`);
     } else if (stdout.includes('java.lang.SecurityException')) {
       // if the app is disabled on a real device it will throw a security exception
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -157,7 +157,8 @@ apkUtilsMethods.startApp = async function startApp (startAppOptions = {}) {
     }
     return stdout;
   } catch (e) {
-    throw new Error(`Cannot start the '${startAppOptions.pkg}' application. ` +
+    const appDescriptor = startAppOptions.pkg || `${startAppOptions.action} ${startAppOptions.optionalIntentArguments}`;
+    throw new Error(`Cannot start the '${appDescriptor}' application. ` +
       `Visit ${ACTIVITIES_TROUBLESHOOTING_LINK} for troubleshooting. ` +
       `Original error: ${e.message}`);
   }
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -147,8 +147,12 @@ apkUtilsMethods.startApp = async function startApp (startAppOptions = {}) {
       }
       throw new Error(`Activity name '${startAppOptions.activity}' used to start the app doesn't ` +
                       `exist or cannot be launched! Make sure it exists and is a launchable activity`);
-    } else if (stdout.includes('Error: Intent does not match any activities')) {
-      throw new Error(`Activity for intent '${startAppOptions.action} ${startAppOptions.optionalIntentArguments}' used to start the app doesn't ` +
+    } else if (stdout.includes('Error: Intent does not match any activities') || stdout.includes('Error: Activity not started, unable to resolve Intent')) {
+      let intentName = startAppOptions.action;
+      if (startAppOptions.optionalIntentArguments) {
+        intentName += ` ${startAppOptions.optionalIntentArguments}`;
+      }
+      throw new Error(`Activity for intent '${intentName}' used to start the app doesn't ` +
                       `exist or cannot be launched! Make sure it exists and is a launchable activity`);
     } else if (stdout.includes('java.lang.SecurityException')) {
       // if the app is disabled on a real device it will throw a security exception
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -107,12 +107,17 @@ apkUtilsMethods.startUri = async function startUri (uri, pkg) {
  * @throws {Error} If there is an error while executing the activity
  */
 apkUtilsMethods.startApp = async function startApp (startAppOptions = {}) {
-  if (!startAppOptions.activity || !startAppOptions.pkg) {
-    throw new Error('activity and pkg are required to start an application');
+  const hasAppAndPkg = startAppOptions.activity && startAppOptions.pkg;
+  const hasIntent = startAppOptions.optionalIntentArguments && startAppOptions.action;
+
+  if (!hasAppAndPkg && !hasIntent) {
+    throw new Error('activity and pkg, or intent, are required to start an application');
   }
 
   startAppOptions = _.clone(startAppOptions);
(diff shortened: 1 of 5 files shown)


test/functional/apk-utils-e2e-specs.js
```diff
@@ -78,6 +78,21 @@ describe('apk utils', function () {
       await assertPackageAndActivity();
 
     });
+
+    it('should be able to start with an intent and no activity', async function () {
+      await adb.install(contactManagerPath);
+      await adb.startApp({
+        pkg: 'com.example.android.contactmanager',
+        category: 'android.intent.category.DEFAULT',
```
jlipps: if category is required, should we also validate it as part of this case?

vikki: Hmm, I'm not sure.I don't think it's required technically (I know I said that in the description though!), but it was required for my use case.

It will run etc. without an explicit category but I think if you don't provide one, it uses the default one `android.intent.category.LAUNCHER` - I *think* that's set in `appium` main. 

It might be clearer/less coupled in the test to leave out the category? I just tried it locally and it still works so I'm tempted to take it out.

jlipps: Yeah, I'd consider it a better test if it didn't have other stuff that potentially alters whether or not the tested functionality works.

vikki: Totally agree; I wouldn't have left it in if I'd realised beforehand. Removed now, thanks!

lib/tools/apk-utils.js
```diff
@@ -107,12 +107,17 @@ apkUtilsMethods.startUri = async function startUri (uri, pkg) {
  * @throws {Error} If there is an error while executing the activity
  */
 apkUtilsMethods.startApp = async function startApp (startAppOptions = {}) {
-  if (!startAppOptions.activity || !startAppOptions.pkg) {
-    throw new Error('activity and pkg are required to start an application');
+  const hasAppAndPkg = startAppOptions.activity && startAppOptions.pkg;
+  const hasIntent = startAppOptions.optionalIntentArguments && startAppOptions.action;
+
+  if (!hasAppAndPkg && !hasIntent) {
+    throw new Error('activity and pkg, or intent, are required to start an application');
```
mykola-mokhnach: just intent or intent and action?

mykola-mokhnach:
apk utils
       startApp
         should be able to start with an intent and no activity:
     Error: Cannot start the 'undefined' application. Visit https://github.com/appium/appium/blob/master/docs/en/writing-running-appium/android/activity-startup.md for troubleshooting. Original error: Error executing adbExec. Original error: 'Command '/Users/runner/Library/Android/sdk/platform-tools/adb -P 5037 shell am start -W -S -a 
The tests seem to fail:

```
android.intent.action.VIEW -d content\://com.android.contacts/contacts' exited with code 1'; Stderr: 'Error: Intent does not match any activities: Intent { act=android.intent.action.VIEW dat=content://com.android.contacts/contacts }'; Code: '1'
      at ADB.startApp (lib/tools/apk-utils.js:160:11)
```

mykola-mokhnach:
@vikki Do you need help fixing the failing test?

btw, I've recently fixed the Travis pipeline. Please rebase with master to get the fix

vikki:
@mykola-moknach I think I'm good; sorry for the delay, I've been off sick for a bit but will hopefully have a fix later on today :)
Thanks for the heads up around rebasing also

mykola-mokhnach:
np, get well soon!

vikki:
Thank you :) I refactored the code to handle errors when apps are started with an intent instead of an activity, and added an extra test to cover that. Just rebased and slightly messed up the branch but now have a meeting, will try to sort later. Also if I squished the branch that would sort the CLA issue, would that be ok?

mykola-mokhnach:
We anyway squash on merge to master, so I don't really see a need to do that

vikki:
Oh I just meant because I committed originally from my new work laptop without git author properly configured, so the CLA bot has an author-not-signed https://github.com/appium/appium-adb/pull/475#issuecomment-558753194

mykola-mokhnach:
Just make sure the email address for the local GitHub account config is the same as it was set while signing the CLA. The next commit should with the fixed email should make the bot happy. If not then just recreate the PR

vikki:
Sorry about the failing tests! I think the main issue is to do with finding an intent that is supported on the CI setup, could you help me with that please? I tried to pick generic ones that haven't been deprecated (e.g. DIAL and WEB_SEARCH) and they pass locally and via local cmd line but doesn't seem to work once pushed. I'm happy to look into it further but not sure how to go about understanding what's different about the remote builds.

mykola-mokhnach:
Usually the stuff on Travis is pretty slow and that might be the reason. If the tests passes locally and fails on Travis then make sure you set appropriate timeouts for that.

mykola-mokhnach:
Also [here](https://dev.azure.com/AppiumCI/Appium%20CI/_build/results?buildId=6608) the log says ```Error: Cannot start the 'android.intent.action.WEB_SEARCH undefined' application. Visit https://github.com/appium/appium/blob/master/docs/en/writing-running-appium/android/activity-startup.md for troubleshooting. Original error: Error executing adbExec. Original error: 'Command '/Users/runner/Library/Android/sdk/platform-tools/adb -P 5037 shell am start -W -S -a android.intent.action.WEB_SEARCH' exited with code 1'; Stderr: 'Error: Intent does not match any activities: Intent { act=android.intent.action.WEB_SEARCH }'; Code: '1'
      at ADB.startApp (lib/tools/apk-utils.js:164:11)```
I'm pretty suspicious about `undefined` trace there

vikki:
> Also [here](https://dev.azure.com/AppiumCI/Appium%20CI/_build/results?buildId=6608) the log says `Error: Cannot start the 'android.intent.action.WEB_SEARCH undefined' application. Visit https://github.com/appium/appium/blob/master/docs/en/writing-running-appium/android/activity-startup.md for troubleshooting. Original error: Error executing adbExec. Original error: 'Command '/Users/runner/Library/Android/sdk/platform-tools/adb -P 5037 shell am start -W -S -a android.intent.action.WEB_SEARCH' exited with code 1'; Stderr: 'Error: Intent does not match any activities: Intent { act=android.intent.action.WEB_SEARCH }'; Code: '1' at ADB.startApp (lib/tools/apk-utils.js:164:11)`
> I'm pretty suspicious about `undefined` trace there

I know what the undefined thing is (it's just because I made the optionalIntentArguments, err, optional, so I need to coalesce undefined to empty string in the error message.  I can fix that pretty quickly but more keen to understand the other issue.

The tests where it *should* start the activity via the intent all run locally for me, so now trying to work out what causes them to fail in CI - if it's the latency or something to with the build setup (when the intent is unsupported it tends to fail quite quickly when I run them locally, so not certain it's latency but want to confirm).

mykola-mokhnach:
Maybe it also has to do with the stop (-S) option enabled by default. See https://android.googlesource.com/platform/frameworks/base/+/21bdaf1/cmds/am/src/com/android/commands/am/Am.java#837

vikki:
It was the stop argument, thank you! I've not included that in validation as it *can* be useful sometimes (I think i've been using it locally when I run tests on the same app over and over so I don't think you'd *always* want set `stopApp` to `false`).

lib/tools/apk-utils.js
```diff
@@ -142,6 +147,13 @@ apkUtilsMethods.startApp = async function startApp (startAppOptions = {}) {
       }
       throw new Error(`Activity name '${startAppOptions.activity}' used to start the app doesn't ` +
                       `exist or cannot be launched! Make sure it exists and is a launchable activity`);
+    } else if (stdout.includes('Error: Intent does not match any activities') || stdout.includes('Error: Activity not started, unable to resolve Intent')) {
+      let intentName = startAppOptions.action;
```
mykola-mokhnach:
```
const intentName = `${startAppOptions.action}${startAppOptions.optionalIntentArguments ? ' ' + startAppOptions.optionalIntentArguments : ''}`
```

vikki: I did that first but it felt harder to read - would you prefer I switched it?

lib/tools/apk-utils.js
```diff
@@ -152,7 +164,8 @@ apkUtilsMethods.startApp = async function startApp (startAppOptions = {}) {
     }
     return stdout;
   } catch (e) {
-    throw new Error(`Cannot start the '${startAppOptions.pkg}' application. ` +
+    const appDescriptor = startAppOptions.pkg || `${startAppOptions.action} ${startAppOptions.optionalIntentArguments}`;
+    throw new Error(`Cannot start the '${appDescriptor}' application. ` +
       `Visit ${ACTIVITIES_TROUBLESHOOTING_LINK} for troubleshooting. ` +
```
mykola-mokhnach: I think it would also be useful to update the troubleshooting document with some information about intents

lib/tools/apk-utils.js
```diff
@@ -152,7 +164,8 @@ apkUtilsMethods.startApp = async function startApp (startAppOptions = {}) {
     }
     return stdout;
   } catch (e) {
-    throw new Error(`Cannot start the '${startAppOptions.pkg}' application. ` +
+    const appDescriptor = startAppOptions.pkg || `${startAppOptions.action} ${startAppOptions.optionalIntentArguments}`;
```
mykola-mokhnach: what is startAppOptions.optionalIntentArguments is unset?

lib/tools/apk-utils.js
mykola-mokhnach: yeah, it's quite long. Although, there is probably a need to do the same operation for the name in the exception handler below. Perhaps, it makes sense to extract the formatting into a separate closure?

lib/tools/apk-utils.js
vikki: I just missed this one first time round trying to get the CI tests to pass! Fixed this in last commit, along with the refactor above to reuse `intentName`

mykola-mokhnach:
Now the tests seem to be ok. It still fails on Travis because of its slowness. We could try to set an increased timeout as a workaround. If that still does not work then simple skip the tests if they are running on this CI. Something like 
```
    if (process.env.TRAVIS) {
      // This test is unstable on Travis
      return this.skip();
    }
```

test/unit/apk-utils-specs.js
```diff
@@ -464,6 +464,23 @@ describe('Apk-utils', withMocks({adb, fs, teen_process}, function (mocks) {
...
+      };
+      const cmdWithIntent = ['am', 'start', '-W', '-S', '-a', 'android.intent.action.VIEW', '-c', 'android.intent.category.DEFAULT', '-d', 'scheme://127.0.0.1'];
+
+      mocks.adb.expects('getApiLevel')
+        .once().withExactArgs()
+        .returns(17);
+      mocks.adb.expects('shell')
+        .once().withArgs(cmdWithIntent)
+        .returns('');
+      (await adb.startApp(startAppOptionsWithIntent));
```
mykola-mokhnach: parentheses are redundant

vikki: I'm happy to remove this, but a lot of the other ones in the same block have the same style so wanted to be consistent - e.g. 520, 513, 503 etc.) - happy to change all or some if you prefer though?

test/functional/apk-utils-e2e-specs.js
```diff
@@ -29,7 +29,7 @@ describe('apk utils', function () {
 
   before(async function () {
     adb = await ADB.createADB({
-      adbExecTimeout: (process.env.TRAVIS || process.env.CI) ? 60000 : 40000,
+      adbExecTimeout: (process.env.TRAVIS || process.env.CI) ? MOCHA_LONG_TIMEOUT : 40000,
```
vikki: Also, I was expecting this (along with lines 82 + 93) to bump the timeout to the Travis long version of the timeout - e.g. 240000, am I setting it up wrong? (still seems to be timing out at https://travis-ci.org/appium/appium-adb/jobs/619672867?utm_medium=notification&utm_source=github_status )

test/unit/apk-utils-specs.js
mykola-mokhnach: you could remove all extra parentheses for good, although this is not a requirement

test/functional/apk-utils-e2e-specs.js
mykola-mokhnach: `waitDuration`, if set, has priority over the generic `adbExecTimeout` setting
See the implementation of `apkUtilsMethods.startApp` for more details

lib/tools/apk-utils.js
```diff
@@ -107,12 +107,17 @@ apkUtilsMethods.startUri = async function startUri (uri, pkg) {
  * @throws {Error} If there is an error while executing the activity
  */
 apkUtilsMethods.startApp = async function startApp (startAppOptions = {}) {
-  if (!startAppOptions.activity || !startAppOptions.pkg) {
-    throw new Error('activity and pkg are required to start an application');
+  const hasAppAndPkg = startAppOptions.activity && startAppOptions.pkg;
+  const hasIntent = startAppOptions.action;
```
mykola-mokhnach: Please update the `StartAppOptions` documentation above

vikki:
Sorry for the delay with this; I had to go back to working on other stuff. I've changed the tests to be more flexible in verifying which package was launched from the intent as that varies between setups/android versions - would you be happy with this approach? (I think it's nicer than skipping on CI but if you'd rather I don't mind changing it). (I also had to include optionalIntentArguments, which whilst not mandatory, were required to get the websearch intent to work across all your CI envs; I can't think of a simpler intent to launch but if there is one I'm v. happy to switch!)

Everything but the CLA check is now OK; I think if I squash my commits then that might fix the CLA check as then the other author (my new MBP!) would be removed from the commit history.

The only outstanding thing I'm aware of is updating the troubleshooting doc  (as per https://github.com/appium/appium-adb/pull/475#discussion_r352177016) - would you want me to PR that before this is merged?

lib/tools/apk-utils.js
```diff
@@ -107,12 +110,14 @@ apkUtilsMethods.startUri = async function startUri (uri, pkg) {
  * @throws {Error} If there is an error while executing the activity
  */
 apkUtilsMethods.startApp = async function startApp (startAppOptions = {}) {
-  if (!startAppOptions.activity || !startAppOptions.pkg) {
-    throw new Error('activity and pkg are required to start an application');
+  if (!startAppOptions.pkg || !(startAppOptions.activity || startAppOptions.action)) {
+    throw new Error('pkg, and activity or intent action, are required to start an application');
```
mykola-mokhnach: I'm not quite sure this condition works as expected

Could you please enumerate which combinations are valid?

test/functional/apk-utils-e2e-specs.js
mykola-mokhnach: do we still need to increase the timeout since we set `waitDuration` ?

mykola-mokhnach:
> The only outstanding thing I'm aware of is updating the troubleshooting doc (as per #475 (comment)) - would you want me to PR that before this is merged?

No, not necessary. We can update the doc after the PR is merged

lib/tools/apk-utils.js
vikki: Sure :)
I just pushed a commit that should verify that (https://github.com/appium/appium-adb/pull/475/commits/05380632160944f7238ed91b8279672822529642) too just in case. It should work like the below, but LMK if you think this isn't right (either code or way it's set up) or if you think we could reword the error message to be clearer. (hoping this is easier to fix than the tests-in-CI was!)

valid:
- pkg + activity (no action)
- pkg + action   (no activity)
- pkg + action + activity *

invalid:
- pkg (no activity + no action)
- activity (no pkg + no action)
- action (no pkg + no activity)

* (this will basically ignore the intent action but it's not invalid and adb will allow it so I didn't validate against it)

mykola-mokhnach: that's ok, thanks for the update

mykola-mokhnach:
I am ok with the PR. Only one question left: https://github.com/appium/appium-adb/pull/475/files#r359462752

test/functional/apk-utils-e2e-specs.js
vikki: Just pushed a commit with this reverted; hopefully not needed any more :)

PR #481, opened 2019-12-17
chore: Bump the cached application time stamp if needed

mykola-mokhnach:
It makes sense to update the timestamp of the cached file if the file has been reused, so it does not get deleted because its position in the sorted ls output would be too close to the bottom

### DIFF
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -420,8 +420,13 @@ apkUtilsMethods.cacheApk = async function cacheApk (apkPath, options = {}) {
   log.debug(`The count of applications in the cache: ${remoteCachedFiles.length}`);
   const toHash = (remotePath) => path.posix.parse(remotePath).name;
   // Push the apk to the remote cache if needed
-  if (remoteCachedFiles.find((x) => toHash(x) === appHash)) {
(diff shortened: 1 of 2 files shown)


lib/tools/apk-utils.js
```diff
@@ -420,8 +420,13 @@ apkUtilsMethods.cacheApk = async function cacheApk (apkPath, options = {}) {
   log.debug(`The count of applications in the cache: ${remoteCachedFiles.length}`);
   const toHash = (remotePath) => path.posix.parse(remotePath).name;
   // Push the apk to the remote cache if needed
-  if (remoteCachedFiles.find((x) => toHash(x) === appHash)) {
+  if (remoteCachedFiles.some((x) => toHash(x) === appHash)) {
     log.info(`The application at '${apkPath}' is already cached to '${remotePath}'`);
+    try {
+      // Update the application timestamp in order to bump its position
+      // in the sorted ls output
+      await this.shell(['touch', '-am', remotePath]);
```
imurchie: Could this be run and not waited-for?

mykola-mokhnach: yes it could

PR #490, opened 2020-02-15
fix: remove appium_apk_tools.jar for macOS notarization

KazuCocoa:
This jar has been removed in https://github.com/appium/appium-adb/pull/471
`appium_apk_tools.jar/prebuilt/aapt/macosx/aapt` in the jar also got `The binary uses an SDK older than the 10.9 SDK` error.


lib/adb.js
```diff
@@ -55,7 +55,7 @@ class ADB {
 
   initJars () {
     const tempJars = [
-      'move_manifest.jar', 'sign.jar', 'appium_apk_tools.jar',
+      'move_manifest.jar', 'sign.jar',
```
mykola-mokhnach: I think move_manifest.jar could also be safely excluded

mykola-mokhnach: it also makes sense to rebuild other jars with newer java version to make the notarization process successful

KazuCocoa:
```
$ git grep "move_manifest.jar"
lib/adb.js:      'move_manifest.jar', 'sign.jar',
$ git grep "move_manifest"
lib/adb.js:      'move_manifest.jar', 'sign.jar',
```

lib/adb.js
KazuCocoa: Yes, but other jars are no issue for the notarization according to my friend.
So, let me do it as another PR

PR #496, opened 2020-02-21
feat: Allow to type Unicode characters

mykola-mokhnach:
https://github.com/appium/appium/issues/13154

### DIFF
--- a/lib/tools/settings-client-commands.js
+++ b/lib/tools/settings-client-commands.js
@@ -463,7 +463,7 @@ commands.typeUnicode = async function typeUnicode (text) {
     return false;
   }
   await this.runInImeContext(UNICODE_IME,
-    async () => await this.shell(['input', 'text', imap.encode(`${text}`)]));
+    async () => await this.shell(['input', 'text', imap.encode(text)]));
   return true;
 };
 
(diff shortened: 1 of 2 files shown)


lib/tools/settings-client-commands.js
```diff
@@ -442,4 +444,27 @@ commands.getSmsList = async function getSmsList (opts = {}) {
...
+/**
+ * Types the given Unicode string.
+ * It is expected that the focus is already put
+ * to the destination input field before this method is called.
+ *
+ * @param {string} text The string to type
+ * @returns {boolean} `true` if the input text has been successfully sent to adb
+ */
+commands.typeUnicode = async function typeUnicode (text) {
+  if (_.isNil(text)) {
```
imurchie: Why this check and not just having a default string of `''` and allowing the check below to return `false`?

mykola-mokhnach: I want all values except of null and undefined to be actually printed as they are (false -> `false`, `0` -> '0', etc). The `text` parameter will be received from the client via `mobile:` command

imurchie: Ok. Only `null` would work differently with a default parameter.

mykola-mokhnach: yep, I know. Thanks for the hint

PR #497, opened 2020-02-24
feat: Return API Level 30 for Android R preview

KazuCocoa:
`getPlatformVersion` returns 29 in Android R preview, although the version should be for Android Q.
In Android Q, the returned value became from 28 to 29 in middle or later beta. Thus, I predict R also will be 30, but for now, let's add a workaround the same as Android Q case for Android R preview.

### DIFF
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -87,11 +87,13 @@ methods.getApiLevel = async function getApiLevel () {
       let apiLevel = parseInt(strOutput.trim(), 10);
 
       // Workaround for preview/beta platform API level
-      const asciiq = 113; // 113 is the code point of 'q'
-      const apiLevelDiff = apiLevel - 28; // 28 is the first API Level getPlatformVersion() returns alphabet
-      if (apiLevelDiff >= 0 && (await this.getPlatformVersion()).toLowerCase() === String.fromCharCode(asciiq + apiLevelDiff)) {
-        log.debug(`Release version is ${String.fromCharCode(asciiq + apiLevelDiff).toUpperCase()} but found API Level ${apiLevel}. Setting API Level to ${apiLevel + 1}`);
-        apiLevel += 1;
+      const charCodeQ = 'q'.charCodeAt(0);
+      // 28 is the first API Level, where Android SDK started returning letters in response to getPlatformVersion
+      const apiLevelDiff = apiLevel - 28;
+      const codename = String.fromCharCode(charCodeQ + apiLevelDiff);
+      if (apiLevelDiff >= 0 && (await this.getPlatformVersion()).toLowerCase() === codename) {
+        log.debug(`Release version is ${codename.toUpperCase()} but found API Level ${apiLevel}. Setting API Level to ${apiLevel + 1}`);
+        apiLevel++;
       }
 
       this._apiLevel = apiLevel;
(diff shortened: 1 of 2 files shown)


lib/tools/adb-commands.js
```diff
@@ -86,11 +86,15 @@ methods.getApiLevel = async function getApiLevel () {
       const strOutput = await this.getDeviceProperty('ro.build.version.sdk');
       let apiLevel = parseInt(strOutput.trim(), 10);
 
-      // Temp workaround. Android Q beta emulators report SDK 28 when they should be 29
-      if (apiLevel === 28 && (await this.getDeviceProperty('ro.build.version.release')).toLowerCase() === 'q') {
+      // Workaround for preview/beta platform API level
+      if (apiLevel === 28 && (await this.getPlatformVersion()).toLowerCase() === 'q') {
```
mykola-mokhnach: I would rather make it dynamic, for example:

if we know that api28 + platformQ = api29 then we could assume that api(28+X) + platform(Q+X) = api(28+X+1)

KazuCocoa: `String.fromCharCode`, ah, JS has...

lib/tools/adb-commands.js
```diff
@@ -86,11 +86,14 @@ methods.getApiLevel = async function getApiLevel () {
       const strOutput = await this.getDeviceProperty('ro.build.version.sdk');
       let apiLevel = parseInt(strOutput.trim(), 10);
 
-      // Temp workaround. Android Q beta emulators report SDK 28 when they should be 29
-      if (apiLevel === 28 && (await this.getDeviceProperty('ro.build.version.release')).toLowerCase() === 'q') {
-        log.debug('Release version is Q but found API Level 28. Setting API Level to 29');
-        apiLevel = 29;
+      // Workaround for preview/beta platform API level
+      const asciiq = 113; // 113 is the code point of 'q'
```
mykola-mokhnach: use `'q'.charCodeAt(0)`

lib/tools/adb-commands.js
```diff
@@ -86,11 +86,14 @@ methods.getApiLevel = async function getApiLevel () {
...
+      const apiLevelDiff = apiLevel - 28; // 28 is the first API Level getPlatformVersion() returns alphabet
```
mykola-mokhnach: ... is the first API Level, where Android SDK started returning letters in response to `getPlatformVersion`

lib/tools/adb-commands.js
```diff
@@ -86,11 +86,14 @@ methods.getApiLevel = async function getApiLevel () {
...
+      if (apiLevelDiff >= 0 && (await this.getPlatformVersion()).toLowerCase() === String.fromCharCode(asciiq + apiLevelDiff)) {
```
mykola-mokhnach: `String.fromCharCode(asciiq + apiLevelDiff)` could be precalculated

lib/tools/adb-commands.js
mykola-mokhnach: asciiq -> charCodeQ

lib/tools/adb-commands.js
```diff
@@ -86,11 +86,14 @@ methods.getApiLevel = async function getApiLevel () {
...
+        log.debug(`Release version is ${String.fromCharCode(asciiq + apiLevelDiff).toUpperCase()} but found API Level ${apiLevel}. Setting API Level to ${apiLevel + 1}`);
+        apiLevel += 1;
```
mykola-mokhnach: apiLevel++

PR #502, opened 2020-03-05
feat: Introduce signature verification options

mykola-mokhnach:
UIA2 requires the application under test to be signed by any valid signature, not just the default one we use in Appium. this feature plus some improvements in UIA2 driver will allow to save session init time for already signed apps.

### DIFF
--- a/lib/tools/apk-signing.js
+++ b/lib/tools/apk-signing.js
@@ -263,13 +263,21 @@ apkSigningMethods.zipAlignApk = async function zipAlignApk (apk) {
 };
 
 /**
+ * @typedef {Object} CertCheckOptions
+ * @property {boolean} requireDefaultCert [true] Whether to require that the destination APK
+ * is signed with the default Appium certificate or any valid certificate. This option
+ * only has effect if `useKeystore` property is unset.
+ */
+
+/**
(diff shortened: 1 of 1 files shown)


lib/tools/apk-signing.js
```diff
@@ -284,33 +292,40 @@ apkSigningMethods.checkApkCert = async function checkApkCert (appPath, pkg) {
...
-      log.debug(`'${appPath}' is signed with non-default certificate`);
-      return false;
+    if (_.includes(output, DEFAULT_CERT_DIGEST)) {
+      log.debug(`'${appPath}' is signed with the default certificate`);
+    } else {
+      log.debug(`'${appPath}' is signed with a non-default certificate`);
     }
-    log.debug(`'${appPath}' is already signed.`);
-    return true;
+    return !requireDefaultCert || (requireDefaultCert && _.includes(output, DEFAULT_CERT_DIGEST));
```
imurchie: I don't really understand what the second instance of `requireDefaultCert` is doing here. It will always be `true` in this case, no?

mykola-mokhnach: true, updated

PR #503, opened 2020-03-05
chore: Avoid throwing exceptions if the given process to kill does not exist

mykola-mokhnach:
https://dev.azure.com/AppiumCI/Appium%20CI/_build/results?buildId=8104&view=logs&j=25e09710-041d-5e84-9b7c-7583b448ad52&t=12c02101-029e-5dde-340b-a9ad73ea9568&l=1109

### DIFF
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -1127,7 +1127,7 @@ methods.killProcessesByName = async function killProcessesByName (name) {
  * to properly kill the process.
  *
  * @param {string|number} pid - The ID of the process to be killed.
- * @return {string} Kill command stdout.
+ * @return {string} Kill command's stdout or stderr if there was no such process.
  * @throws {Error} If the process cannot be killed.
  */
 methods.killProcessByPID = async function killProcessByPID (pid) {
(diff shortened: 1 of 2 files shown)


lib/tools/adb-commands.js
```diff
@@ -1128,7 +1128,7 @@ methods.killProcessesByName = async function killProcessesByName (name) {
  *
  * @param {string|number} pid - The ID of the process to be killed.
  * @return {string} Kill command stdout.
```
imurchie: Or stderr if there is no such process.

lib/tools/adb-commands.js
```diff
@@ -1151,7 +1154,14 @@ methods.killProcessByPID = async function killProcessByPID (pid) {
       log.info(`Cannot kill PID ${pid} due to insufficient permissions. Retrying as root`);
       let {isSuccessful} = await this.root();
       becameRoot = isSuccessful;
```
imurchie: It's outside of your change, but this can be `({isSuccessful: becameRoot} = await this.root());` and avoid the throw-away variable.

mykola-mokhnach: thanks for the suggestion

PR #504, opened 2020-03-06
fix: make sure device is reachable in createADB

imurchie:
Since `getApiLevel` is memoized once successful, this does not have performance impacts. Within the drivers we always get the API level directly after getting the ADB instance, so that will just be pre-populated.

This will necessitate some refactoring of https://github.com/appium/appium-uiautomator2-driver/pull/369

### DIFF
--- a/lib/adb.js
+++ b/lib/adb.js
@@ -3,8 +3,11 @@ import os from 'os';
 import path from 'path';
 import methods from './tools/index.js';
 import { rootDir, DEFAULT_ADB_EXEC_TIMEOUT } from './helpers';
+import { retryInterval } from 'asyncbox';
+
 
 const DEFAULT_ADB_PORT = 5037;
+const ADB_INITIAL_RETRIES = 5;
 const JAR_PATH = path.resolve(rootDir, 'jars');
 const DEFAULT_OPTS = {
   sdkRoot: null,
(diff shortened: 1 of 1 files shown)


imurchie:
The Azure failure is with assets, not the tests.

lib/adb.js
```diff
@@ -65,6 +68,8 @@ class ADB {
 ADB.createADB = async function createADB (opts) {
   let adb = new ADB(opts);
   await adb.getAdbWithCorrectAdbPath();
+  // get the api level to make sure that the device is reachable
+  await retryInterval(ADB_INITIAL_RETRIES, 100, adb.getApiLevel.bind(adb));
```
mykola-mokhnach: should this behavior be controlled by `opts`?

imurchie: I don’t really see a reason to without more data on a problem. I’ve run it locally and through the uiautomator2 driver on travis. As I said, we always get the api level after getting an instance, so this is happening anyway in practical terms. And we know that retryInterval works as it should.

mykola-mokhnach: I don't have a strong opinion on the current implementation. Just wanted to make sure it's more flexible.

imurchie: Short of adding a capability, any option is going to take work to get in, so I'd rather wait until there is actually a problem. It would, however, be easy enough to add if needed.

PR #505, opened 2020-03-07
Revert "fix: make sure device is reachable in createADB (#504)"

mykola-mokhnach:
This reverts commit 0b029c4fd6132bcfb9b614f673eb8cb4653c6b74.

It turns out this patch will fail adb creation if multiple devices are connected to the same machine, because we sometimes should find out the actual udid of the device under test after adb instance is created by analyzing the `list` command output. Perhaps, it makes sense to move this patch into appium-android-driver's helpers, methods `createADB` and `getDeviceInfoFromCaps` after `setDeviceId` has been called.

### DIFF
--- a/lib/adb.js
+++ b/lib/adb.js
@@ -3,11 +3,8 @@ import os from 'os';
 import path from 'path';
 import methods from './tools/index.js';
 import { rootDir, DEFAULT_ADB_EXEC_TIMEOUT } from './helpers';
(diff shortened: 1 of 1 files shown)


imurchie:
Though I personally would rather the code from https://github.com/appium/appium-android-driver/blob/master/lib/android-helpers.js#L258-L267 be moved into `ADB.createADB`.

PR #516, opened 2020-04-09
chore: Print a better error message if JAVA_HOME value is invalid

mykola-mokhnach:


mykola-mokhnach:
No, this all is the internal API used only in scope of this module

PR #518, opened 2020-04-22
chore: Use quote helper imported from appium-support

mykola-mokhnach:
There is no need to explicitly import `shell-quote`, since it's already being exposed by appium-support module


package.json
```diff
@@ -52,7 +52,6 @@
     "lodash": "^4.0.0",
     "lru-cache": "^5.0.0",
     "semver": "^7.0.0",
-    "shell-quote": "^1.6.1",
```
KazuCocoa, line 55: yey

PR #520, opened 2020-04-27
feat: Add an option to control intent wait for a started URI

shibupanda:
This PR will fix below issue.
https://github.com/appium/appium/issues/14180

### DIFF
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -55,28 +55,19 @@ apkUtilsMethods.isAppInstalled = async function isAppInstalled (pkg) {
  * @param {string} pkg - The name of the package to start the URI with.
  * @param {boolean} waitForLaunch - Optional parameter of waitForLaunch caps to add wait in adb command
  */
-apkUtilsMethods.startUri = async function startUri (uri, pkg, waitForLaunch = true) {
-  let args;
+apkUtilsMethods.startUri = async function startUri (uri, pkg, opts = { waitForLaunch: true }) {
   if (!uri || !pkg) {
     throw new Error('URI and package arguments are required');
   }
 
-  if (!waitForLaunch) {
-    args = [
-      'am', 'start',
-      '-a', 'android.intent.action.VIEW',
-      '-d', uri.replace(/&/g, '\\&'),
-      pkg,
-    ];
-  } else {
-    args = [
-      'am', 'start',
-      '-W',
-      '-a', 'android.intent.action.VIEW',
-      '-d', uri.replace(/&/g, '\\&'),
-      pkg,
-    ];
+  const args = ['am', 'start'];
+  if (opts.waitForLaunch) {
+    args.push('-W');
   }
+  args.push('-a', 'android.intent.action.VIEW',
+       '-d', uri.replace(/&/g, '\\&'),
+       pkg);
+
   try {
     const res = await this.shell(args);
     if (res.toLowerCase().includes('unable to resolve intent')) {
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -49,6 +49,12 @@ apkUtilsMethods.isAppInstalled = async function isAppInstalled (pkg) {
 };
 
 /**
+ * @typedef {Object} StartUriOptions
+ * @property {?boolean} waitForLaunch [true] - if `false` then adb won't wait
+ * for the started activity to return the control
+ */
+
+/**
  * Start the particular URI on the device under test.
  *
  * @param {string} uri - The name of URI to start.
--- a/lib/tools/apk-utils.js
+++ b/lib/tools/apk-utils.js
@@ -53,20 +53,24 @@ apkUtilsMethods.isAppInstalled = async function isAppInstalled (pkg) {
  *
  * @param {string} uri - The name of URI to start.
  * @param {string} pkg - The name of the package to start the URI with.
- * @param {boolean} waitForLaunch - Optional parameter of waitForLaunch caps to add wait in adb command
+ * @param {StartUriOptions} opts
  */
-apkUtilsMethods.startUri = async function startUri (uri, pkg, opts = { waitForLaunch: true }) {
+apkUtilsMethods.startUri = async function startUri (uri, pkg, opts = {}) {
+  const {
+    waitForLaunch = true,
+  } = opts;
+
   if (!uri || !pkg) {
     throw new Error('URI and package arguments are required');
   }
 
   const args = ['am', 'start'];
-  if (opts.waitForLaunch) {
+  if (waitForLaunch) {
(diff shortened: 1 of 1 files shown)


shibupanda:
@KazuCocoa Could you please check why this check failed. Is it because something I am doing wrong?

lib/tools/apk-utils.js
```diff
@@ -53,19 +53,30 @@ apkUtilsMethods.isAppInstalled = async function isAppInstalled (pkg) {
  *
  * @param {string} uri - The name of URI to start.
  * @param {string} pkg - The name of the package to start the URI with.
+ * @param {boolean} waitForLaunch - Optional parameter of waitForLaunch caps to add wait in adb command
```
mykola-mokhnach, line 56: please make the third parameter an object: `opts = {}`. then `waitForLaunch` could be one of its properties

lib/tools/apk-utils.js
```diff
@@ -53,19 +53,30 @@ apkUtilsMethods.isAppInstalled = async function isAppInstalled (pkg) {
...
 
-  const args = [
-    'am', 'start',
-    '-W',
-    '-a', 'android.intent.action.VIEW',
-    '-d', uri.replace(/&/g, '\\&'),
-    pkg,
-  ];
+  if (!waitForLaunch) {
+    args = [
```
mykola-mokhnach, line 65: Please do not duplicate args. The `-W` could be just pushed based on the provided options:

```
const args = ['am', 'start'];
if (waitForLaunch) {
  args.push('-W');
}
args.push('-a', 'android.intent.action.VIEW',
       '-d', uri.replace(/&/g, '\\&'),
       pkg);
```

shibupanda: @mykola-mokhnach Thanks. I will do the suggested changes.

shibupanda: const args = ['am', 'start'];
  if (opts.waitForLaunch) {
    args.push('-W');
  }
  args.push('-a', 'android.intent.action.VIEW',
       '-d', uri.replace(/&/g, '\\&'),
       pkg);

lib/tools/apk-utils.js
shibupanda: This how it is going to look.
apkUtilsMethods.startUri = async function startUri (uri, pkg, opts = { waitForLaunch: true }) {

lib/tools/apk-utils.js
```diff
@@ -53,19 +53,21 @@ apkUtilsMethods.isAppInstalled = async function isAppInstalled (pkg) {
...
-    '-a', 'android.intent.action.VIEW',
-    '-d', uri.replace(/&/g, '\\&'),
-    pkg,
-  ];
+  const args = ['am', 'start'];
+  if (opts.waitForLaunch) {
+    args.push('-W');
+  }
+  args.push('-a', 'android.intent.action.VIEW',
+       '-d', uri.replace(/&/g, '\\&'),
```
KazuCocoa, line 68: Could you change 2 indentation style?

shibupanda: @KazuCocoa you mean both args should be in same vertical line. 
Eslint didn't suggest me this so I missed.

shibupanda: It is throwing me error when I am trying to change

lib/tools/apk-utils.js
mykola-mokhnach: please also update docs.
Also, it would be more readable to destruct the `opts` arg inside the method body (especially when it has more possible properties). See https://github.com/appium/appium-adb/blob/d8fce0016c3d532f251f2ae6b1351993dd6cbb83/lib/tools/settings-client-commands.js#L46 for more details

lib/tools/apk-utils.js
```diff
@@ -53,19 +53,21 @@ apkUtilsMethods.isAppInstalled = async function isAppInstalled (pkg) {
...
   }
 
-  const args = [
-    'am', 'start',
-    '-W',
-    '-a', 'android.intent.action.VIEW',
-    '-d', uri.replace(/&/g, '\\&'),
-    pkg,
-  ];
+  const args = ['am', 'start'];
```
mykola-mokhnach, line 63: 👍

lib/tools/apk-utils.js
```diff
@@ -53,19 +53,21 @@ apkUtilsMethods.isAppInstalled = async function isAppInstalled (pkg) {
...
+  if (opts.waitForLaunch) {
```
mykola-mokhnach, line 64: waitForLaunch could be destructed before

lib/tools/apk-utils.js
KazuCocoa: ah, okay

lib/tools/apk-utils.js
shibupanda: @mykola-mokhnach Sorry I am not able to understand it better.

mykola-mokhnach: have you checked the example above?

shibupanda: Ohk, Got it. I am going through it. Thanks for guiding me.

shibupanda: How it looks now?

  @param {string} pkg - The name of the package to start the URI with.
  @param {StartUriOptions} opts
apkUtilsMethods.startUri = async function startUri (uri, pkg, opts = {}) {
  if (!uri || !pkg) {
    throw new Error('URI and package arguments are required');
  }

  const {
    waitForLaunch = true,
  } = opts;

  const args = ['am', 'start'];
  if (waitForLaunch) {
    args.push('-W');
  }

lib/tools/apk-utils.js
shibupanda: https://github.com/appium/appium-adb/blob/master/docs/COMMANDS.md
Here I need to update correct with two scenario with false or without.

lib/tools/apk-utils.js
```diff
@@ -53,19 +53,25 @@ apkUtilsMethods.isAppInstalled = async function isAppInstalled (pkg) {
  *
  * @param {string} uri - The name of URI to start.
  * @param {string} pkg - The name of the package to start the URI with.
+ * @param {StartUriOptions} opts
```
mykola-mokhnach, line 56: Please add typedef for StartUriOptions (see how it is done in the example like I posted before)

shibupanda:
Thanks @mykola-mokhnach and @KazuCocoa It was a great learning for me even the process.
Going forward I will take care all standard and contribute more.
@mykola-mokhnach Anything I missed or it's ready to merge.

lib/tools/apk-utils.js
shibupanda: Ohk. Got it.

PR #527, opened 2020-05-19
feat: Use -delay-adb emulator command line option to detect its startup

mykola-mokhnach:
https://github.com/appium/appium/issues/14347


lib/tools/adb-emu-commands.js
```diff
@@ -327,4 +328,35 @@ emuMethods.execEmuConsoleCommand = async function execTelnet (cmd, opts = {}) {
...
+ */
+
+/**
+ * Retrieves emulator version from the file system
+ *
+ * @returns {EmuVersionInfo} If no version info could be parsed then an empty
+ * object is returned
+ */
+emuMethods.getEmuVersionInfo = async function getEmuVersionInfo () {
+  const propsPath = path.join(this.sdkRoot, 'emulator', 'source.properties');
```
KazuCocoa, line 344: 👍

lib/tools/system-calls.js
```diff
@@ -717,6 +697,15 @@ systemCallMethods.launchAVD = async function launchAVD (avdName, opts = {}) {
   if (!_.isEmpty(args)) {
     launchArgs.push(...(_.isArray(args) ? args : util.shellParse(`${args}`)));
   }
+
+  const {revision} = await this.getEmuVersionInfo();
+  let supportsDelayAdbFeature = false;
+  if (revision && util.compareVersions(revision, '>=', '29.0.7')) {
+    // https://androidstudio.googleblog.com/2019/05/emulator-2907-canary.html
```
KazuCocoa, line 704: oh, nice to find it

lib/tools/adb-commands.js
```diff
@@ -1196,56 +1196,27 @@ methods.killProcessesByName = async function killProcessesByName (name) {
...
-    if (becameRoot) {
-      await this.unroot();
+    log.info(`Cannot kill PID ${pid} due to insufficient permissions. Retrying as root`);
+    try {
+      await this.shell(['kill', pid], {
+        privileged: true
+      });
+    } catch (e1) {
+      if (_.includes(e1.stderr, noProcessFlag)) {
+        return e1.stderr;
```
jlipps, line 1217: why do we want to return the error string?

lib/tools/system-calls.js
```diff
@@ -918,15 +919,33 @@ systemCallMethods.reboot = async function reboot (retries = DEFAULT_ADB_REBOOT_R
...
 
+  const retryIfOffline = async (cmdFunc) => {
+    try {
+      return await cmdFunc();
+    } catch (err) {
+      // Check the output of the stdErr to see if there's any clues that show that the device went offline
+      // and if it did go offline, restart ADB
+      if (['closed', 'device offline', 'timeout expired']
+          .some((x) => (err.stderr || '').toLowerCase().includes(x))) {
+        log.warn(`Attempt to ${cmd} caused device to go offline. Restarting adb`);
```
jlipps, line 930: `caused ADB to think device went offline`?

lib/tools/adb-commands.js
mykola-mokhnach: because it was already implemented like that. Do you have better ideas on what could be returned? Currently the result is anyway not used anywhere

jlipps: I guess I would expect it to return nothing if the result is not valuable. And since it's an error message, if we do 'return' it I would expect it to be thrown as an error and not returned as a string.

lib/tools/system-calls.js
```diff
@@ -236,6 +230,7 @@ systemCallMethods.restartAdb = async function restartAdb () {
   log.debug('Restarting adb');
   try {
     await this.killServer();
+    await this.adbExec(['start-server']);
```
KazuCocoa, line 233: ah, we did not start the server even if we named 'restart'

mykola-mokhnach: I assume we relied on the following command. E.g. sending it should also make adb to start automatically

lib/tools/adb-commands.js
mykola-mokhnach: changed it to return nothing

PR #529, opened 2020-05-23
feat: add cmdline-tools path in binary search

KazuCocoa:
Noticed below commands were in `cmdline-tools` path for newer packages.

```
apkanalyzer
avdmanager
lint
screenshot2
sdkmanager
```

image

`latest` is the default one. (I mean other versions can select when 'show packages' is on. For now, it is only 1.0 though)

### DIFF
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -75,9 +75,9 @@ systemCallMethods.getBinaryFromSdkRoot = async function getBinaryFromSdkRoot (bi
   const binaryLocs = [
     'platform-tools',
     'emulator',
+    `cmdline-tools${path.sep}latest${path.sep}bin`,
     'tools',
-    `tools${path.sep}bin`,
-    `cmdline-tools${path.sep}latest${path.sep}bin`
+    `tools${path.sep}bin`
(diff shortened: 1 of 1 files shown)


lib/tools/system-calls.js
```diff
@@ -72,8 +72,13 @@ systemCallMethods.getBinaryFromSdkRoot = async function getBinaryFromSdkRoot (bi
   }
 
   const fullBinaryName = this.getBinaryNameForOS(binaryName);
-  const binaryLocs = ['platform-tools', 'emulator', 'tools', `tools${path.sep}bin`]
-    .map((x) => path.resolve(this.sdkRoot, x, fullBinaryName));
+  const binaryLocs = [
+    'platform-tools',
+    'emulator',
+    `cmdline-tools${path.sep}latest${path.sep}bin`,
```
mykola-mokhnach, line 78: I think it makes sense to represent parts of the path as array items, so the resulting assignment might look like

```
const binaryLocs = [
  'platform-tools',
  ...
  ['cmdline-tools', 'latest', 'bin]',
  ...
].map((x) => path.resolve(this.sdkRoot, ...(_.isArray(x) ? x : [x]), fullBinaryName))

```

PR #533, opened 2020-05-25
feat: Add reconnect command handler

mykola-mokhnach:
Killing the whole adb server can sometimes corrupt other parallel tests that are running on the same machine and expect adb connection to be stable. The `reconnect` command affects only offline devices and does not touch other properly connected units, which is much more convenient for us from this perspective.

### DIFF
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -973,7 +973,7 @@ methods.restart = async function restart () {
     await this.stopLogcat();
     await this.restartAdb();
     await this.waitForDevice(60);
-    await this.startLogcat();
+    await this.startLogcat(this._logcatStartupParams);
   } catch (e) {
     throw new Error(`Restart failed. Original error: ${e.message}`);
   }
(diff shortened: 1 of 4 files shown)


lib/tools/adb-commands.js
```diff
@@ -973,7 +973,7 @@ methods.restart = async function restart () {
     await this.stopLogcat();
     await this.restartAdb();
     await this.waitForDevice(60);
-    await this.startLogcat();
+    await this.startLogcat(this._logcatStartupParams);
```
KazuCocoa, line 976: 👍

lib/tools/system-calls.js
```diff
@@ -267,15 +267,43 @@ systemCallMethods.getDevicesWithRetry = async function getDevicesWithRetry (time
...
     await sleep(200);
     return await getDevices();
   };
   return await getDevices();
 };
 
+/**
+ * Kick current connection from host/device side and make it reconnect
+ *
+ * @param {?string} target [offline] One of possible targets to reconnect:
```
KazuCocoa, line 286: nice to find 👍 
`offline` target affects `offline` and `unauthorized`  devices

PR #538, opened 2020-06-04
feat: Add '-d' option to adb install command

DaveTryon:
As documented at https://developer.android.com/studio/command-line/adb, recent versions of `adb install` support the -d flag to support downgrades. Appium-adb currently provides no way to enable this flag. This change simply enables the option in buildInstallArgs, then exposes the parameter up the stack to the caller.

Note: 4 aapt-related unit tests are failing, but they are also failing in master and not related to this change.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -455,6 +455,8 @@ const extractMatchingPermissions = function (dumpsysOutput, groupNames, grantedS
  * @typedef {Object} InstallOptions
  * @property {boolean} allowTestPackages [false] - Set to true in order to allow test
  *                                                 packages installation.
+ * @property {boolean} allowDowngrade [false] - Set to true in order to allow
+ *                                              package downgrades.
  * @property {boolean} useSdcard [false] - Set to true to install the app on sdcard
  *                                         instead of the device memory.
  * @property {boolean} grantPermissions [false] - Set to true in order to grant all the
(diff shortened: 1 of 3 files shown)


lib/helpers.js
```diff
@@ -455,6 +455,8 @@ const extractMatchingPermissions = function (dumpsysOutput, groupNames, grantedS
  * @typedef {Object} InstallOptions
  * @property {boolean} allowTestPackages [false] - Set to true in order to allow test
  *                                                 packages installation.
+ * @property {boolean} allowDowngrade [false] - Set to true in order to allow
```
mykola-mokhnach, line 458: this option only works for debuggable packages. 
also, package downgrade could potentially corrupt the data.

What is your reason to have it?

DaveTryon:
Closing/Reopening to bump CLA check after adding my email

lib/helpers.js
DaveTryon: @mykola-mokhnach, we have an accessibility tool (Accessibility Insights) that will be auto-installing an APK onto an Android device. We have scenarios where we'll need to downgrade our service version in response to customer actions. We were hoping to use the downgrade flag to explicitly handle this scenario. 

I was unaware of the restriction that this only works with debuggable packages--is this documented somewhere?

mykola-mokhnach: yes, this is stated in both the adb source code and in the `adb help` output:

```
     -d: allow version code downgrade (debuggable packages only)
```

DaveTryon:
Closing as this isn't as generally applicable as I had originally thought

lib/helpers.js
DaveTryon: Thanks, @mykola-mokhnach. I'll go ahead and close this out since it doesn't seem like it would be generally useful in a production environment.

PR #547, opened 2020-08-07
chore: Only use -delay-adb emulator startup detection if image API version is above 27

mykola-mokhnach:
https://github.com/appium/appium/issues/14347#issuecomment-670440234


lib/tools/system-calls.js
```diff
@@ -792,8 +792,20 @@ systemCallMethods.launchAVD = async function launchAVD (avdName, opts = {}) {
   let supportsDelayAdbFeature = false;
   if (revision && util.compareVersions(revision, '>=', '29.0.7')) {
     // https://androidstudio.googleblog.com/2019/05/emulator-2907-canary.html
-    launchArgs.push('-delay-adb');
-    supportsDelayAdbFeature = true;
+    try {
+      const {target} = await this.getEmuImageProperties(avdName);
+      const apiMatch = /\d+/.exec(target);
+      // https://issuetracker.google.com/issues/142533355
```
KazuCocoa, line 798: Nice to find this

PR #548, opened 2020-08-25
chore: Remove obsolete signing tools

mykola-mokhnach:
The apksigner.jar tool is a standard recommended way to manage apks signing and exists in Android build tools since version 24.0.3, which has been released in September 2016. Also, sign/verify.jar themselves are not being maintained for quite a while and are outdated/cannot properly handle signing according to the recent standards. This also sometimes causes unexpected test failures when the fallback from `apksigner.jar` is used.

This PR removes the fallback to sign/verify.jar and only relies on `apksigner.jar`. An exception will be thrown if `apksigner.jar` is not present in the local system or if it cannot sign/verify the binary.


KazuCocoa (approved):
Seems reasonable to me

KazuCocoa:
Should set `buildToolsVersion` to use lower version of build tools?

PR #549, opened 2020-08-25
refactor: Replace avdmanager usage with faster config reading primitives

mykola-mokhnach:
It turns out that Google bumped minimum java version for its build tools, but forgot to do this for SDK tools, like avdmanager. This causes java version incompatibility errors while executing `avdmanager` script on systems where Java9+ is set in JAVA_HOME.
This PR removed usage of avdmanager and replaces it with NodeJS-based config reader, which is way faster and does not experience the above issues

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -869,8 +869,11 @@ async function getAndroidPrefsRoot () {
   }
 
   if (location && await fs.exists(location)) {
-    const stats = await fs.stat(location);
-    return stats.isDirectory() ? location : null;
+    if (!(await fs.stat(location)).isDirectory()) {
+      log.debug(`Android config root '${location}' is not a directory`);
+      return null;
+    }
+    return location;
   }
 
   return null;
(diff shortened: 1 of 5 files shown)


lib/helpers.js
```diff
@@ -842,6 +842,40 @@ function toAvdLocaleArgs (language, country) {
...
+  } else {
+    const home = process.env.HOME;
+    if (home) {
+      location = path.resolve(home, '.android');
+    }
+  }
+
+  if (location && await fs.exists(location)) {
+    const stats = await fs.stat(location);
+    return stats.isDirectory() ? location : null;
```
KazuCocoa, line 873: how about print the location as debug log when it is not directory?

KazuCocoa (approved):
Did CI get lint issue?

lib/helpers.js
mykola-mokhnach: added log

KazuCocoa (approved):

PR #550, opened 2020-09-05
fix: Update process lookup behaviour for newer Android versions

mykola-mokhnach:
At some point of time Google has changed the default `ps` behaviour, so it only lists processes that belong to the current shell user rather to all users. It is necessary to execute ps with -A command line argument to mimic the previous behaviour.

Also related to https://github.com/appium/appium/issues/14709


KazuCocoa (commented):
👍

PR #556, opened 2020-10-21
chore: Perform apks cache cleanup on process exit

mykola-mokhnach:


lib/tools/apks-utils.js
```diff
@@ -18,6 +18,24 @@ const APKS_CACHE = new LRU({
 });
 const APKS_CACHE_GUARD = new AsyncLock();
 
+process.on('exit', () => {
+  if (APKS_CACHE.itemCount === 0) {
+    return;
+  }
+
+  const paths = APKS_CACHE.values();
+  log.debug(`Performing cleanup of ${paths.length} cached .apks ` +
```
KazuCocoa, line 27: Is `itemCount` proper instead of `paths.length`?
`values()` and `itemCount` refer to `LRU_LIST`
https://github.com/isaacs/node-lru-cache/blob/a71be6cd547d28b24d28957a15b10cd3eb34555a/index.js#L124

mykola-mokhnach: here we only delete the paths stored in `values()`, so it makes sense to show their count even if the actual itemCount is different (I don't see a reason it shouldn't though)

KazuCocoa (approved):

PR #558, opened 2020-11-24
fix: Android preferences root path on Windows

jpkleemans:
On Windows avds are stored, by default, in `C:\Users\{USER}\.android\avd`.

The current version of `listEmulators()` searches for the avd folder in `C:\Users\{USER}\AppData\Local\Android` (via `getAndroidPrefsRoot()`), but that's the wrong location.

> By default, the emulator stores configuration files under $HOME/.android/ and AVD data under $HOME/.android/avd/. You can override the defaults by setting the following environment variables. The emulator -avd <avd_name> command searches the avd directory in the order of the values in $ANDROID_AVD_HOME, $ANDROID_SDK_HOME/.android/avd/, and $HOME/.android/avd/.

Source: https://developer.android.com/studio/command-line/variables

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -849,29 +849,41 @@ function toAvdLocaleArgs (language, country) {
  */
 async function getAndroidPrefsRoot () {
   let location = process.env.ANDROID_EMULATOR_HOME;
-  if (location && !await fs.exists(location)) {
-    log.debug(`Value of $ANDROID_EMULATOR_HOME '${location}' does not exist`);
-    return null;
+  if (await dirExists(location)) {
+    return location;
   }
 
-  if (!location) {
-    const home = process.env.HOME || process.env.USERPROFILE;
-    if (home) {
-      location = path.resolve(home, '.android');
-    }
+  if (location) {
+    log.debug(`Value of $ANDROID_EMULATOR_HOME '${location}' is not an existing directory`);
   }
 
-  if (!location || !await fs.exists(location)) {
-    log.debug(`Android config root '${location}' does not exist`);
-    return null;
+  const home = process.env.HOME || process.env.USERPROFILE;
+  if (home) {
+    location = path.resolve(home, '.android');
   }
 
-  if (!(await fs.stat(location)).isDirectory()) {
-    log.debug(`Android config root '${location}' is not a directory`);
+  if (!await dirExists(location)) {
+    log.debug(`Android config root '${location}' is not an existing directory`);
     return null;
   }
 
   return location;
+}
+
+function dirExists (location) {
+  if (!location) {
+    return false;
+  }
+
+  if (!await fs.exists(location)) {
+    return false;
+  }
+
+  if (!(await fs.stat(location)).isDirectory()) {
+    return false;
+  }
+
+  return true;
 }
 
 export {
@@ -883,5 +895,5 @@ export {
   APK_INSTALL_TIMEOUT, APKS_INSTALL_TIMEOUT, buildInstallArgs, APK_EXTENSION,
   DEFAULT_ADB_EXEC_TIMEOUT, parseManifest, parseAaptStrings, parseAapt2Strings,
   formatConfigMarker, parseJsonData, unsignApk, toAvdLocaleArgs, requireSdkRoot,
-  getSdkRootFromEnv, getAndroidPrefsRoot,
+  getSdkRootFromEnv, getAndroidPrefsRoot, dirExists,
 };
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -854,7 +854,7 @@ async function getAndroidPrefsRoot () {
   }
 
   if (location) {
-    log.warn(`The value of $ANDROID_EMULATOR_HOME '${location}' is not an existing directory`);
+    log.warn(`The value of the ANDROID_EMULATOR_HOME environment variable '${location}' is not an existing directory`);
   }
 
   const home = process.env.HOME || process.env.USERPROFILE;
@@ -872,7 +872,7 @@ async function getAndroidPrefsRoot () {
 
 /**
  * Check if a path exists on the filesystem and is a directory
- * 
+ *
  * @param {?string} location The full path to the directory
  * @returns {boolean}
  */
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -858,15 +858,17 @@ async function getAndroidPrefsRoot () {
     }
   }
 
-  if (location && await fs.exists(location)) {
-    if (!(await fs.stat(location)).isDirectory()) {
-      log.debug(`Android config root '${location}' is not a directory`);
-      return null;
-    }
-    return location;
+  if (!location || !await fs.exists(location)) {
+    log.debug(`Android config root '${location}' does not exist`);
+    return null;
   }
 
-  return null;
+  if (!(await fs.stat(location)).isDirectory()) {
+    log.debug(`Android config root '${location}' is not a directory`);
+    return null;
+  }
+
+  return location;
 }
 
 export {
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -870,7 +870,7 @@ async function getAndroidPrefsRoot () {
   return location;
 }
 
-function dirExists (location) {
+async function dirExists (location) {
   if (!location) {
     return false;
   }
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -852,7 +852,7 @@ async function getAndroidPrefsRoot () {
   let location = process.env.ANDROID_EMULATOR_HOME;
 
   if (!location || !await fs.exists(location)) {
-    const home = process.env.ANDROID_SDK_HOME || process.env.HOME || process.env.USERPROFILE;
+    const home = process.env.HOME || process.env.USERPROFILE;
     if (home) {
       location = path.resolve(home, '.android');
     }
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -871,19 +871,9 @@ async function getAndroidPrefsRoot () {
 }
 
 async function dirExists (location) {
-  if (!location) {
-    return false;
-  }
-
-  if (!await fs.exists(location)) {
-    return false;
-  }
-
-  if (!(await fs.stat(location)).isDirectory()) {
-    return false;
-  }
-
-  return true;
+  return location
+    && await fs.exists(location)
+    && (await fs.stat(location)).isDirectory();
 }
 
 export {
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -849,10 +849,13 @@ function toAvdLocaleArgs (language, country) {
  */
 async function getAndroidPrefsRoot () {
   // https://gerrit.pixelexperience.org/plugins/gitiles/development/+/5c11852110eeb03dc5a69111354b383f98d15336/tools/androidprefs/src/com/android/prefs/AndroidLocation.java
-  let location = null;
-  const home = process.env.HOME || process.env.USERPROFILE;
-  if (home) {
-    location = path.resolve(home, '.android');
+  let location = process.env.ANDROID_EMULATOR_HOME;
+
+  if (!location) {
+    const home = process.env.ANDROID_SDK_HOME || process.env.HOME || process.env.USERPROFILE;
+    if (home) {
+      location = path.resolve(home, '.android');
+    }
   }
 
   if (location && await fs.exists(location)) {
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -850,22 +850,9 @@ function toAvdLocaleArgs (language, country) {
 async function getAndroidPrefsRoot () {
   // https://gerrit.pixelexperience.org/plugins/gitiles/development/+/5c11852110eeb03dc5a69111354b383f98d15336/tools/androidprefs/src/com/android/prefs/AndroidLocation.java
   let location = null;
-  if (system.isWindows()) {
-    let localAppData = process.env.LOCALAPPDATA;
-    if (!localAppData) {
-      localAppData = process.env.USERPROFILE;
-      if (localAppData) {
-        localAppData = path.resolve(localAppData, 'Local Settings', 'Application Data');
-      }
-    }
-    if (localAppData) {
-      location = path.resolve(localAppData, 'Android');
-    }
-  } else {
-    const home = process.env.HOME;
-    if (home) {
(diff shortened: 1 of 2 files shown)


lib/helpers.js
```diff
@@ -850,22 +850,9 @@ function toAvdLocaleArgs (language, country) {
...
-    }
-    if (localAppData) {
-      location = path.resolve(localAppData, 'Android');
-    }
-  } else {
-    const home = process.env.HOME;
-    if (home) {
-      location = path.resolve(home, '.android');
-    }
+  const home = process.env.HOME || process.env.USERPROFILE;
```
mykola-mokhnach, line 853: I think it would then make sense to read and check ANDROID_AVD_HOME and ANDROID_EMULATOR_HOME variables first. Also, it is not very clear what is HOME in Windows. Does it always equal to USERPROFILE value?

lib/helpers.js
```diff
@@ -850,22 +850,9 @@ function toAvdLocaleArgs (language, country) {
 async function getAndroidPrefsRoot () {
   // https://gerrit.pixelexperience.org/plugins/gitiles/development/+/5c11852110eeb03dc5a69111354b383f98d15336/tools/androidprefs/src/com/android/prefs/AndroidLocation.java
```
mykola-mokhnach, line 851: it makes sense to update this link to explain the actual implementation

lib/helpers.js
jpkleemans: Good point, I've added checks for `$ANDROID_AVD_HOME`, `$ANDROID_EMULATOR_HOME` and `$ANDROID_SDK_HOME`

lib/helpers.js
```diff
@@ -849,20 +849,10 @@ function toAvdLocaleArgs (language, country) {
...
-    }
-    if (localAppData) {
-      location = path.resolve(localAppData, 'Android');
-    }
-  } else {
-    const home = process.env.HOME;
+  let location = process.env.ANDROID_EMULATOR_HOME;
+
+  if (!location) {
+    const home = process.env.ANDROID_SDK_HOME || process.env.HOME || process.env.USERPROFILE;
```
mykola-mokhnach, line 855: ANDROID_AVD_HOME

mykola-mokhnach: also it makes sense to not only check it is present but also that the actual directory exists

mykola-mokhnach: same for ANDROID_EMULATOR_HOME 
I would also add some logging in case these env variables do not point to valid locations

jpkleemans: Thanks for your comments! The directory-exists-check is already done on line 861:
```js
if (location && await fs.exists(location))
```

> ANDROID_AVD_HOME/.. instead of ANDROID_SDK_HOME

I think it should be `ANDROID_SDK_HOME`, because this function returns the `.android` root folder, not the `avd` folder. The `listEmulators` function resolves the `avd` folder.

mykola-mokhnach: > the directory-exists-check is already done on line 861

the is no such check for ANDROID_EMULATOR_HOME

mykola-mokhnach: the documentation says

> ANDROID_AVD_HOME   Sets the path to the directory that contains all AVD-specific files, which mostly consist of very large disk images. The default location is $ANDROID_EMULATOR_HOME/avd/. You might want to specify a new location if the default location is low on disk space. 

And on my system:

```
$ echo $ANDROID_SDK_HOME                                                                                                                                                        
/Users/user/Library/Android/sdk
```

but the .android folder is located at `/Users/user`

jpkleemans: > the is no such check for ANDROID_EMULATOR_HOME

I've added the check

jpkleemans: > the documentation says
> 
> > ANDROID_AVD_HOME   Sets the path to the directory that contains all AVD-specific files, which mostly consist of very large disk images. The default location is $ANDROID_EMULATOR_HOME/avd/. You might want to specify a new location if the default location is low on disk space.
> 
> And on my system:
> 
> ```
> $ echo $ANDROID_SDK_HOME                                                                                                                                                        
> /Users/user/Library/Android/sdk
> ```
> 
> but the .android folder is located at `/Users/user`

`listEmulators()` appends the `avd` folder to the path returned by `getAndroidPrefsRoot()` (ln 77):
```js
avdsRoot = path.resolve(prefsRoot, 'avd');
```

So the return value of `getAndroidPrefsRoot()` should be the parent of `/avd`. `$ANDROID_EMULATOR_HOME` is that, according to the docs you referenced.

lib/helpers.js
```diff
@@ -849,34 +849,26 @@ function toAvdLocaleArgs (language, country) {
...
-      if (localAppData) {
-        localAppData = path.resolve(localAppData, 'Local Settings', 'Application Data');
-      }
-    }
-    if (localAppData) {
-      location = path.resolve(localAppData, 'Android');
-    }
-  } else {
-    const home = process.env.HOME;
+  let location = process.env.ANDROID_EMULATOR_HOME;
```
mykola-mokhnach, line 852: we need to provide more detailed logs if ANDROID_EMULATOR_HOME and ANDROID_AVD_HOME are set to incorrect values, so a user could figure it out and fix their values if needed

lib/helpers.js
mykola-mokhnach: `const home = process.env.ANDROID_SDK_HOME || process.env.HOME || process.env.USERPROFILE;` -> `const home = process.env.HOME || process.env.USERPROFILE;`

jpkleemans: Ah okay, removed that 👍

lib/helpers.js
jpkleemans: Removed the link

lib/helpers.js
```diff
@@ -848,35 +848,30 @@ function toAvdLocaleArgs (language, country) {
...
+  if (location && !await fs.exists(location)) {
+    log.debug(`Value of $ANDROID_EMULATOR_HOME '${location}' does not exist`);
+    return null;
```
mykola-mokhnach, line 854: we could print a warning here and then still proceed to HOME checks

lib/helpers.js
```diff
@@ -848,35 +848,30 @@ function toAvdLocaleArgs (language, country) {
...
-        localAppData = path.resolve(localAppData, 'Local Settings', 'Application Data');
-      }
-    }
-    if (localAppData) {
-      location = path.resolve(localAppData, 'Android');
-    }
-  } else {
-    const home = process.env.HOME;
+  let location = process.env.ANDROID_EMULATOR_HOME;
+  if (location && !await fs.exists(location)) {
```
mykola-mokhnach, line 852: it also makes sense to check that this is a directory. Consider extracting such check itself into a separate function as it is used more than once

lib/tools/adb-emu-commands.js
```diff
@@ -65,14 +65,24 @@ emuMethods.SENSORS = {
  * @returns {Array<EmuInfo>}
  */
 async function listEmulators () {
-  // https://android.googlesource.com/platform/sdk/+/33f330d963c1bf24667eddb891531abcce9fb4de/sdkmanager/libs/sdklib/src/com/android/sdklib/internal/avd/AvdManager.java
-  const prefsRoot = await getAndroidPrefsRoot();
-  if (!prefsRoot) {
+  let avdsRoot = process.env.ANDROID_AVD_HOME;
+  if (avdsRoot && !await fs.exists(avdsRoot)) {
+    log.debug(`Value of $ANDROID_AVD_HOME '${location}' does not exist`);
     return [];
```
mykola-mokhnach, line 71: same as above, we could print a warning here and proceed with default paths

lib/tools/adb-emu-commands.js
```diff
@@ -65,23 +65,30 @@ emuMethods.SENSORS = {
...
 
-  const stats = await fs.stat(avdsRoot);
-  if (!stats.isDirectory()) {
-    log.debug(`Virtual devices config root '${avdsRoot}' is not a directory`);
-    return [];
-  }
+  return mapAvdDir(avdsRoot);
+}
 
+function mapAvdDir (avdsRoot) {
```
mykola-mokhnach, line 91: async function

lib/tools/adb-emu-commands.js
```diff
@@ -65,23 +65,30 @@ emuMethods.SENSORS = {
...
+    log.debug(`Virtual devices config root '${avdsRoot}' is not an existing directory`);
     return [];
   }
 
-  const stats = await fs.stat(avdsRoot);
-  if (!stats.isDirectory()) {
-    log.debug(`Virtual devices config root '${avdsRoot}' is not a directory`);
-    return [];
-  }
+  return mapAvdDir(avdsRoot);
```
mykola-mokhnach, line 88: await

lib/tools/adb-emu-commands.js
```diff
@@ -65,23 +65,30 @@ emuMethods.SENSORS = {
  * @returns {Array<EmuInfo>}
  */
 async function listEmulators () {
-  // https://android.googlesource.com/platform/sdk/+/33f330d963c1bf24667eddb891531abcce9fb4de/sdkmanager/libs/sdklib/src/com/android/sdklib/internal/avd/AvdManager.java
+  let avdsRoot = process.env.ANDROID_AVD_HOME;
+  if (await dirExists(avdsRoot)) {
+    return mapAvdDir(avdsRoot);
```
mykola-mokhnach, line 70: await

lib/helpers.js
```diff
@@ -848,35 +848,42 @@ function toAvdLocaleArgs (language, country) {
...
+
+  if (!await dirExists(location)) {
+    log.debug(`Android config root '${location}' is not an existing directory`);
+    return null;
+  }
+
+  return location;
+}
+
+function dirExists (location) {
```
mykola-mokhnach, line 873: Please add a docstring

mykola-mokhnach: async function

lib/helpers.js
```diff
@@ -848,35 +848,42 @@ function toAvdLocaleArgs (language, country) {
...
+  if (!location) {
```
mykola-mokhnach, line 874: how about 
```
return location 
  && await fs.exists(location)
  && (await fs.stat(location)).isDirectory() 
```

lib/helpers.js
```diff
@@ -848,35 +848,42 @@ function toAvdLocaleArgs (language, country) {
...
   }
 
-  if (location && await fs.exists(location)) {
-    if (!(await fs.stat(location)).isDirectory()) {
-      log.debug(`Android config root '${location}' is not a directory`);
-      return null;
-    }
-    return location;
+  if (location) {
+    log.debug(`Value of $ANDROID_EMULATOR_HOME '${location}' is not an existing directory`);
```
mykola-mokhnach, line 857: warn

mykola-mokhnach: The value

lib/tools/adb-emu-commands.js
```diff
@@ -65,23 +65,30 @@ emuMethods.SENSORS = {
...
+  }
+
+  if (avdsRoot) {
+    log.debug(`Value of $ANDROID_AVD_HOME '${location}' does not exist`);
```
mykola-mokhnach, line 74: is not an existing directory

mykola-mokhnach: log.warn

lib/tools/adb-emu-commands.js
mykola-mokhnach: how about renaming it to `getAvdConfigPaths`?

KazuCocoa: Red CI was here

```
Message:
    /Users/runner/work/1/s/lib/tools/adb-emu-commands.js: 'await' is only allowed within async functions (92:18)

  90 | 
  91 | function mapAvdDir (avdsRoot) {
> 92 |   const configs = await fs.glob('*.ini', {
     |                   ^
  93 |     cwd: avdsRoot,
  94 |     absolute: true,
  95 |   });
```

lib/helpers.js
```diff
@@ -848,35 +848,38 @@ function toAvdLocaleArgs (language, country) {
...
   }
 
-  if (location && await fs.exists(location)) {
-    if (!(await fs.stat(location)).isDirectory()) {
-      log.debug(`Android config root '${location}' is not a directory`);
-      return null;
-    }
-    return location;
+  if (location) {
+    log.warn(`The value of $ANDROID_EMULATOR_HOME '${location}' is not an existing directory`);
```
mykola-mokhnach, line 857: I would rather explicitly mention `The value of ANDROID_EMULATOR_HOME environment variable`, because under windows env variables look like `%VAR%`

lib/tools/adb-emu-commands.js
```diff
@@ -65,23 +65,30 @@ emuMethods.SENSORS = {
...
  */
 async function listEmulators () {
-  // https://android.googlesource.com/platform/sdk/+/33f330d963c1bf24667eddb891531abcce9fb4de/sdkmanager/libs/sdklib/src/com/android/sdklib/internal/avd/AvdManager.java
+  let avdsRoot = process.env.ANDROID_AVD_HOME;
+  if (await dirExists(avdsRoot)) {
+    return await getAvdConfigPaths(avdsRoot);
+  }
+
+  if (avdsRoot) {
+    log.warn(`The value of $ANDROID_AVD_HOME '${location}' is not an existing directory`);
```
mykola-mokhnach, line 74: same as above

lib/tools/adb-emu-commands.js
```diff
@@ -65,23 +65,30 @@ emuMethods.SENSORS = {
...
 
-  const stats = await fs.stat(avdsRoot);
-  if (!stats.isDirectory()) {
-    log.debug(`Virtual devices config root '${avdsRoot}' is not a directory`);
-    return [];
-  }
+  return await getAvdConfigPaths(avdsRoot);
+}
 
+async function getAvdConfigPaths (avdsRoot) {
```
mykola-mokhnach, line 91: Please also add a docstring

mykola-mokhnach (approved):
minor comments only

mykola-mokhnach:
Please remove trailing spaces required to make the linter happy: https://travis-ci.org/github/appium/appium-adb/jobs/745830756#L1289

mykola-mokhnach:
Perfect, thanks for your hard work @jpkleemans

jpkleemans:
👍 Thanks for your support

PR #561, opened 2020-12-22
fix: set CWD explicity when executing java -jar apksigner.jar

sshock:
fixes https://github.com/appium/appium/issues/14724

### DIFF
--- a/lib/tools/apk-signing.js
+++ b/lib/tools/apk-signing.js
@@ -40,7 +40,9 @@ apkSigningMethods.executeApksigner = async function executeApksigner (args = [])
     ...args
   ];
   log.debug(`Starting apksigner: ${util.quote(fullCmd)}`);
-  const {stdout, stderr} = await exec(fullCmd[0], fullCmd.slice(1));
+  const {stdout, stderr} = await exec(fullCmd[0], fullCmd.slice(1), {
+    cwd: path.dirname(apkSignerJar)
+  });
   for (let [name, stream] of [['stdout', stdout], ['stderr', stderr]]) {
     if (!_.trim(stream)) {
       continue;
--- a/lib/tools/apk-signing.js
+++ b/lib/tools/apk-signing.js
@@ -40,6 +40,7 @@ apkSigningMethods.executeApksigner = async function executeApksigner (args = [])
     ...args
   ];
   log.debug(`Starting apksigner: ${util.quote(fullCmd)}`);
+  // It is necessary to specify CWD explicitly; see https://github.com/appium/appium/issues/14724#issuecomment-737446715
   const {stdout, stderr} = await exec(fullCmd[0], fullCmd.slice(1), {
     cwd: path.dirname(apkSignerJar)
   });


KazuCocoa (approved):

lib/tools/apk-signing.js
```diff
@@ -40,7 +40,9 @@ apkSigningMethods.executeApksigner = async function executeApksigner (args = [])
     ...args
   ];
   log.debug(`Starting apksigner: ${util.quote(fullCmd)}`);
-  const {stdout, stderr} = await exec(fullCmd[0], fullCmd.slice(1));
+  const {stdout, stderr} = await exec(fullCmd[0], fullCmd.slice(1), {
+    cwd: path.dirname(apkSignerJar)
```
mykola-mokhnach, line 44: Please add a link to the original issue that explains why it is necessary to set cwd

mykola-mokhnach (approved):

lib/tools/apk-signing.js
sshock: Ok, happy to; here goes:

It is necessary to set the cwd here because as discovered in https://github.com/appium/appium/issues/14724, somehow it is possible when appium has been running for a long time (like more than a week) for it to somehow lose it's CWD or something such that when launching `java` it doesn't know the cwd.  Being explicit here works around that issue, although I admit we still have the unanswered question of why does this happen :)

```
Cannot verify the signature of '/home/deployer/node_modules/appium/node_modules/appium-uiautomator2-server/apks/appium-uiautomator2-server-v4.15.0.apk'. Original error: Error occurred during initialization of VM
java.lang.Error: Properties init: Could not determine current working directory.
	at java.lang.System.initProperties(java.base/Native Method)
	at java.lang.System.initPhase1(java.base/System.java:1948)
```

mykola-mokhnach: no need for such details. A simple link to the proper issue comment would suffice

sshock: oh, sorry I did it wrong; I guess this is what you're looking for?:

https://github.com/appium/appium/issues/14724#issuecomment-737446715

mykola-mokhnach: yes, that's it. Just add it as a comment to your patch

sshock: Oh, I finally get what you're saying.  For some reason when you wrote this I thought you were talking about github comments:

> Please add a link to the original issue in comments that explains why it is necessary to set cwd

If you had instead written this, I probably wouldn't have gotten confused: 

> Please add a link to the original issue as a comment in the code that explains why it is necessary to set cwd

Anyway, sorry about the misunderstanding.  I added this commit to the branch, so I think we're good now, yes?  3fe1cf76addbcda00105afebda2557dde69db107

sshock:
@mykola-mokhnach bad news: unfortunately setting CWD here did **not** fix the problem after all; it just so happened that it took 2.5 weeks for it to show up this last time (so waiting 2 weeks was not long enough); I just found out the hard way that all our appium jobs have been failing the last 3 days due to the error in question (https://github.com/appium/appium/issues/14724#issuecomment-737446715)

Anyway, you can undo this if you want since apparently it didn't actually work :(

PR #564, opened 2021-03-17
feat: consider NotificationShade and StatusBar as lock screen for newer android versions

KazuCocoa:
I think we should relax locked screen condition logic for newer Android devices.
https://github.com/appium/appium-adb/blob/master/lib/tools/adb-commands.js#L615-L632

Newer Android devices (i tested a few real devices), lock screen is simply `NotificationShade` or `StatusBar` after keycode 62. But current our logic does not consider such case as locked screen. So, our `lock` never be succeeded.

When these app is focused, Appium can unlock them with https://github.com/appium/appium-android-driver/blob/9d120ba76df18ada8c5793af3d42f9b1aa4937cb/lib/android-helpers.js#L648-L679 , so I think we can consider `NotificationShade` and `StatusBar` as part of `isShowingLockscreen`.

(at least, an Android 7.1 real device had `mShowingLockscreen`. Haven't checked with android 8 and 9.)

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -210,15 +210,21 @@ async function getApkanalyzerForOs (sysHelpers) {
  * in comparison to `adb dumpsys window` output parsing.
  * But the trust command does not work for `Swipe` unlock pattern.
  *
- * In some Android devices (Probably around Android 10+) show `mCurrentFocus` as `NotificationShade` or `StatusBar`
- * as the lock screen while they do not have mShowingLockscreen nor mDreamingLockscreen.
- * Although, these `NotificationShade` and `StatusBar` could be also regular notification view.
+ * In some Android devices (Probably around Android 10+), `mShowingLockscreen` and `mDreamingLockscreen`
+ * do not work to detect lock status. Instead, keyguard preferences helps to detect the lock condition.
+ * Some devices such as Android TV do not have keyguard, so we should keep
+ * screen condition as this primary method.
  *
  * @param {string} dumpsys - The output of dumpsys window command.
  * @return {boolean} True if lock screen is showing.
  */
 function isShowingLockscreen (dumpsys) {
-  return /(mShowingLockscreen=true|mDreamingLockscreen=true|mCurrentFocus=.+NotificationShade|mCurrentFocus=.+StatusBar)/gi.test(dumpsys);
+  if (/(mShowingLockscreen=true|mDreamingLockscreen=true)/gi.test(dumpsys)) {
+    return true;
+  }
+  // Fallback to KeyguardServiceDelegate preference.
+  // mIsShowing and mInputRestricted are true in lock condition.
+  return dumpsys.includes('mIsShowing=false', 'mInputRestricted=false');
 }
 
 /*
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -214,15 +214,14 @@ async function getApkanalyzerForOs (sysHelpers) {
  * do not work to detect lock status. Instead, keyguard preferences helps to detect the lock condition.
  * Some devices such as Android TV do not have keyguard, so we should keep
  * screen condition as this primary method.
- * `mIsShowing` and `mInputRestricted` are `true` in lock condition.
- * Both `false` is unlock condition.
  *
  * @param {string} dumpsys - The output of dumpsys window command.
  * @return {boolean} True if lock screen is showing.
  */
 function isShowingLockscreen (dumpsys) {
   return _.some(['mShowingLockscreen=true', 'mDreamingLockscreen=true'], (x) => dumpsys.includes(x))
-  || _.every(['mIsShowing=false', 'mInputRestricted=false'], (x) => dumpsys.includes(x));
+    // `mIsShowing` and `mInputRestricted` are `true` in lock condition. `false` is unlock condition.
+    || _.every(['mIsShowing=false', 'mInputRestricted=false'], (x) => dumpsys.includes(x));
 }
 
 /*
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -214,17 +214,15 @@ async function getApkanalyzerForOs (sysHelpers) {
  * do not work to detect lock status. Instead, keyguard preferences helps to detect the lock condition.
  * Some devices such as Android TV do not have keyguard, so we should keep
  * screen condition as this primary method.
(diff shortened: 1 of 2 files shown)


lib/helpers.js
```diff
@@ -210,11 +210,15 @@ async function getApkanalyzerForOs (sysHelpers) {
...
+ * In some Android devices (Probably around Android 10+) show `mCurrentFocus` as `NotificationShade` or `StatusBar`
+ * as the lock screen while they do not have mShowingLockscreen nor mDreamingLockscreen.
+ * Although, these `NotificationShade` and `StatusBar` could be also regular notification view.
+ *
  * @param {string} dumpsys - The output of dumpsys window command.
  * @return {boolean} True if lock screen is showing.
  */
 function isShowingLockscreen (dumpsys) {
-  return /(mShowingLockscreen=true|mDreamingLockscreen=true)/gi.test(dumpsys);
+  return /(mShowingLockscreen=true|mDreamingLockscreen=true|mCurrentFocus=.+NotificationShade|mCurrentFocus=.+StatusBar)/gi.test(dumpsys);
```
mykola-mokhnach, line 221: should we be more restrictive about `.+NotificationShade` and `.+StatusBar` regexps to avoid false positives?

mykola-mokhnach: for example `.+\bNotificationShade\b`

KazuCocoa: ah, could be to tune the regex. Let me try.

mykola-mokhnach: also would it be possible to split this long line?

mykola-mokhnach:
I was also checking some finding regarding screen locked state. We could also test if solutions from https://stackoverflow.com/questions/35275828/is-there-a-way-to-check-if-android-device-screen-is-locked-via-adb work for us

KazuCocoa:
Yea, i tried some stuff like `power`.  But for example `adb shell dumpsys power | grep 'mHolding'` was both true in both lock/unlcok in devices when the device's screen was one. ( I checked this behavior with a few nexus devices and a samsung device)
`NotificationShade` and `StatusBar` are the most reasonable way to consider possible lock/unlock (at least after power on button)...

mykola-mokhnach:
Yes, I saw some comments about the output there is not very reliable. What if we check the NFC state first and fall back to this method if it is disabled?

Something like: 

1. adb shell dumpsys nfc | echo FAIL
2. if FAIL is printed then fall back to the current method
3. Otherwise check if the output contains `UNLOCKED`

KazuCocoa:
The nfc command returned outputs but it did not have `mScreenState` in my checked devices..


Oh, I just found `Screen State: ON_UNLOCKED`. No `mScreenState` tho.
```
adb shell dumpsys nfc | grep 'Screen State'
Screen State: ON_UNLOCKED
```

It was `ON_LOCKED` or `OFF_LOCKED` when that was locked.
No `unlocked` word in the nfc output, so potentially we can check only `ON_UNLOCKED` or `_UNLOCKED` match.

Let me check more devices

mykola-mokhnach:
Btw, this is how the check is implemented in Android itself: https://android.googlesource.com/platform/packages/apps/Nfc/+/refs/heads/master/src/com/android/nfc/ScreenStateHelper.java#32

mykola-mokhnach:
Also, how Google checks if the device screen is ready in internal tests: https://android.googlesource.com/platform/cts/+/ab9e203328a3edc5c40c5887bfc5bc8e9ba92769/tests/app/src/android/app/cts/ActivityManagerProcessStateTest.java#159

KazuCocoa:
Thanks, I dug in dumpsys focusing on keyguard. Then, it seems the below condition is the most reliable.
`mIsShowing=false` && `mInputRestricted=false`

They existed since Android OS 6. I haven't checked on OS 5 tho. Non smart phone like Android TV does not have it.

So, the primary is current way -> fallback to this keyguard is the most reliable so far, I think.

- unlocked
```
    KeyguardServiceDelegate
      ....
      KeyguardStateMonitor
        mIsShowing=false
        mSimSecure=false
        mInputRestricted=false
        mCurrentUserId=0
        ...
```

- locked
```
    KeyguardServiceDelegate
      ....
      KeyguardStateMonitor
        mIsShowing=true
        mSimSecure=false
        mInputRestricted=true
        mCurrentUserId=0
        ...
```

lib/helpers.js
```diff
@@ -210,15 +210,21 @@ async function getApkanalyzerForOs (sysHelpers) {
...
  * @return {boolean} True if lock screen is showing.
  */
 function isShowingLockscreen (dumpsys) {
-  return /(mShowingLockscreen=true|mDreamingLockscreen=true|mCurrentFocus=.+NotificationShade|mCurrentFocus=.+StatusBar)/gi.test(dumpsys);
+  if (/(mShowingLockscreen=true|mDreamingLockscreen=true)/gi.test(dumpsys)) {
+    return true;
+  }
+  // Fallback to KeyguardServiceDelegate preference.
+  // mIsShowing and mInputRestricted are true in lock condition.
+  return dumpsys.includes('mIsShowing=false', 'mInputRestricted=false');
```
mykola-mokhnach, line 227: return _.every(['mIsShowing=false', 'mInputRestricted=false'], (x) -> dumpsys.includes(x))

KazuCocoa: Oh, it's cool, `some` and `every` methods

lib/helpers.js
```diff
@@ -214,17 +214,15 @@ async function getApkanalyzerForOs (sysHelpers) {
...
  */
 function isShowingLockscreen (dumpsys) {
-  if (/(mShowingLockscreen=true|mDreamingLockscreen=true)/gi.test(dumpsys)) {
-    return true;
-  }
-  // Fallback to KeyguardServiceDelegate preference.
-  // mIsShowing and mInputRestricted are true in lock condition.
-  return dumpsys.includes('mIsShowing=false', 'mInputRestricted=false');
+  return _.some(['mShowingLockscreen=true', 'mDreamingLockscreen=true'], (x) => dumpsys.includes(x))
+  || _.every(['mIsShowing=false', 'mInputRestricted=false'], (x) => dumpsys.includes(x));
```
mykola-mokhnach, line 225: Please add an indentation here to enhance the readability. The comment also makes sense

mykola-mokhnach (approved):

PR #572, opened 2021-07-13
chore: Assume cert validation succeeds if java vm error is thrown

mykola-mokhnach:
The `java.lang.Error: Properties init: Could not determine current working directory` error pops up randomly while checking app signature with apksigner and we are not quite sure why.  My guess - a race condition in java vm initialization. Nevertheless, lets make Appium to believe the file is already signed, because it would be true for 99% of UIAutomator2-based tests, where we presign server binaries while publishing their NPM module. If these are not signed, e.g. in case of Espresso, then the next step(s) would anyway fail.
See https://github.com/appium/appium/issues/14724 for more details.

### DIFF
--- a/lib/tools/apk-signing.js
+++ b/lib/tools/apk-signing.js
@@ -20,6 +20,7 @@ const MD5 = 'md5';
 const DEFAULT_CERT_HASH = {
   [SHA256]: 'a40da80a59d170caa950cf15c18c454d47a39b26989d8b640ecd745ba71bf5dc'
 };
+const JAVA_PROPS_INIT_ERROR = 'java.lang.Error: Properties init';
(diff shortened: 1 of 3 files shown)


KazuCocoa (approved):
It seems reasonable

lib/tools/apk-signing.js
```diff
@@ -249,26 +250,61 @@ apkSigningMethods.checkApkCert = async function checkApkCert (appPath, pkg, opts
   const {
     requireDefaultCert = true,
   } = opts;
+
+  const appHash = await fs.hash(appPath);
```
KazuCocoa, line 254: ah, good idea

PR #573, opened 2021-08-04
refactor: Fetch java executable location from PATH if JAVA_HOME is unset

mykola-mokhnach:
Related to https://github.com/appium/appium/issues/15693

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -166,8 +166,29 @@ const getJavaHome = _.memoize(async function getJavaHome () {
 });
 
 const getJavaForOs = _.memoize(async function getJavaForOs () {
-  return path.resolve(await getJavaHome(), 'bin',
-    `java${system.isWindows() ? '.exe' : ''}`);
+  let javaHome;
+  let errMsg;
+  try {
+    javaHome = await getJavaHome();
+  } catch (err) {
+    errMsg = err.message;
+  }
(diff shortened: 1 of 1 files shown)


KazuCocoa (approved):

lib/helpers.js
```diff
@@ -166,8 +166,26 @@ const getJavaHome = _.memoize(async function getJavaHome () {
...
+  }
+  const executableName = `java${system.isWindows() ? '.exe' : ''}`;
+  if (javaHome) {
+    const resultPath = path.resolve(javaHome, 'bin', executableName);
+    if (await fs.exists(resultPath)) {
+      return resultPath;
+    }
+  }
+  try {
+    return await fs.which(executableName);
```
jlipps, line 184: should we generalize this for other binaries, not just java?

jlipps (approved):

lib/helpers.js
mykola-mokhnach: which binaries exactly you'd like to generalize?

jlipps: i'm not sure, don't we do similar logic currently for binaries inside ANDROID_HOME, and, e.g., jarsigner?

mykola-mokhnach: jarsigner is only used as a fallback if apksigner fails for some reason, so I would rather leave it as is for now

PR #574, opened 2021-08-10
feat: Add outputFormat option to shell command

jalkire:
[Issue 15696- appium-adb: no way to read stderr from shell commands that exit with code 0](https://github.com/appium/appium/issues/15696) describes a scenario in which appium-adb's `shell` command throws away `stderr` when a command exits with code 0. To add support for this scenario, this PR implements the following approach (as described by @dbjorge in the issue):

> A new `outputFormat` option to `shell`/`adbExec` that: 
> * accepts either `stdout` or `full`
> * defaults to `stdout`
> * is checked during the `adbExec` success path; if `stdout`, `adbExec` matches the existing behavior and returns `stdout`, and if `full`, `adbExec` instead returns the full original output from `await teen_process.exec(...)` (ie, an object with `stdout`, `stderr`, and `code` properties)

Comments on the issue indicated that this is the preferred implementation approach, so I went with it.

### Additional details
- As implemented, this PR does not enforce that `outputFormat` be set to `stdout` to maintain the current behavior--any value other than `full` will result in the default `stdout` return value. I would be happy to update this to instead validate the value and throw and error if it isn't `undefined`, `stdout`, or `full` if that is preferable. 
- I updated documentation relating to the `opts` paramenters for `shell` and `adbExec`. If there is anywhere else I should update documentation, please let me know!
- I attempted to follow the "make your code look like the surrounding code" principle from the [style guide](https://github.com/appium/appium/blob/master/docs/en/contributing-to-appium/style-guide.md) for both the production and test code

### DIFF
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -386,7 +386,10 @@ let isExecLocked = false;
  *                        You can also set the additional `exclusive` param
  *                        to `true` that assures no other parallel adb commands
  *                        are going to be executed while the current one is running
- * @return {string} - Command's stdout.
+ *                        You can set the `outputFormat` param to `stdout` to receive just the stdout
+ *                        output (default) or `full` to receive the entire stdout, stderr, and code
+ *                        response from a command with a zero exit code
+ * @return {string|Object} - Command's stdout or an object containing stdout, stderr, and code.
  * @throws {Error} If the command returned non-zero exit code.
  */
 systemCallMethods.adbExec = async function adbExec (cmd, opts = {}) {
@@ -398,6 +401,7 @@ systemCallMethods.adbExec = async function adbExec (cmd, opts = {}) {
   // setting default timeout for each command to prevent infinite wait.
   opts.timeout = opts.timeout || this.adbExecTimeout || DEFAULT_ADB_EXEC_TIMEOUT;
   opts.timeoutCapName = opts.timeoutCapName || 'adbExecTimeout'; // For error message
+  opts.outputFormat = opts.outputFormat || 'stdout';
 
   cmd = _.isArray(cmd) ? cmd : [cmd];
   let adbRetried = false;
@@ -406,11 +410,11 @@ systemCallMethods.adbExec = async function adbExec (cmd, opts = {}) {
       const args = [...this.executable.defaultArgs, ...cmd];
       log.debug(`Running '${this.executable.path} ` +
         (args.find((arg) => /\s+/.test(arg)) ? util.quote(args) : args.join(' ')) + `'`);
-      let {stdout} = await exec(this.executable.path, args, opts);
+      let {stdout, stderr, code} = await exec(this.executable.path, args, opts);
       // sometimes ADB prints out weird stdout warnings that we don't want
       // to include in any of the response data, so let's strip it out
       stdout = stdout.replace(LINKER_WARNING_REGEXP, '').trim();
-      return stdout;
+      return opts.outputFormat === 'full' ? {stdout, stderr, code} : stdout;
     } catch (e) {
       const errText = `${e.message}, ${e.stdout}, ${e.stderr}`;
       if (ADB_RETRY_ERROR_PATTERNS.some((p) => p.test(errText))) {
@@ -466,6 +470,8 @@ systemCallMethods.adbExec = async function adbExec (cmd, opts = {}) {
  * (used in the error messages).
  * @property {?number} timeout [adbExecTimeout] - command execution timeout.
  * @property {?boolean} privileged [falsy] - Whether to run the given command as root.
+ * @property {?string} outputFormat - Whether response should include full exec output or just stdout.
+ *                                    Potential values are full or stdout; defaults to stdout.
  *
  * All other properties are the same as for `exec` call from {@link https://github.com/appium/node-teen_process}
  * module
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -375,6 +375,11 @@ systemCallMethods.adbExecEmu = async function adbExecEmu (cmd) {
 
 let isExecLocked = false;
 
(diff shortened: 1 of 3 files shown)


lib/tools/system-calls.js
```diff
@@ -386,7 +386,10 @@ let isExecLocked = false;
  *                        You can also set the additional `exclusive` param
  *                        to `true` that assures no other parallel adb commands
  *                        are going to be executed while the current one is running
- * @return {string} - Command's stdout.
+ *                        You can set the `outputFormat` param to `stdout` to receive just the stdout
+ *                        output (default) or `full` to receive the entire stdout, stderr, and code
+ *                        response from a command with a zero exit code
+ * @return {string|Object} - Command's stdout or an object containing stdout, stderr, and code.
```
mykola-mokhnach, line 392: I would rather prefer to have a typedef for the alternative result: https://github.com/appium/appium-adb/pull/574/files#diff-3ea7ac5fe8ae9a5687f7df112cce745b55e4952185827ecd1314a69561f38d5dL201

lib/tools/system-calls.js
```diff
@@ -398,6 +401,7 @@ systemCallMethods.adbExec = async function adbExec (cmd, opts = {}) {
   // setting default timeout for each command to prevent infinite wait.
   opts.timeout = opts.timeout || this.adbExecTimeout || DEFAULT_ADB_EXEC_TIMEOUT;
   opts.timeoutCapName = opts.timeoutCapName || 'adbExecTimeout'; // For error message
+  opts.outputFormat = opts.outputFormat || 'stdout';
```
mykola-mokhnach, line 404: can we do this without mutating `opts`?

lib/tools/system-calls.js
```diff
@@ -406,11 +410,11 @@ systemCallMethods.adbExec = async function adbExec (cmd, opts = {}) {
       const args = [...this.executable.defaultArgs, ...cmd];
       log.debug(`Running '${this.executable.path} ` +
         (args.find((arg) => /\s+/.test(arg)) ? util.quote(args) : args.join(' ')) + `'`);
-      let {stdout} = await exec(this.executable.path, args, opts);
+      let {stdout, stderr, code} = await exec(this.executable.path, args, opts);
       // sometimes ADB prints out weird stdout warnings that we don't want
       // to include in any of the response data, so let's strip it out
       stdout = stdout.replace(LINKER_WARNING_REGEXP, '').trim();
-      return stdout;
+      return opts.outputFormat === 'full' ? {stdout, stderr, code} : stdout;
```
mykola-mokhnach, line 417: can we have constants instead of magic strings?
also `code` is always going to be equal zero there (otherwise an exception is thrown)

lib/tools/system-calls.js
```diff
@@ -466,6 +470,8 @@ systemCallMethods.adbExec = async function adbExec (cmd, opts = {}) {
  * (used in the error messages).
  * @property {?number} timeout [adbExecTimeout] - command execution timeout.
  * @property {?boolean} privileged [falsy] - Whether to run the given command as root.
+ * @property {?string} outputFormat - Whether response should include full exec output or just stdout.
```
mykola-mokhnach, line 473: the default value could be simply mentioned in square brackets

lib/tools/system-calls.js
jalkire: Good idea! Added `ExecResult` typedef containing `stdout` and `stderr` (no `code` per your other comment)

lib/tools/system-calls.js
jalkire: Sure, now setting new `outputFormat` value here instead of mutating `opts`

lib/tools/system-calls.js
jalkire: Separated `full` and `stdout` strings into consts within the function's scope. I saw both within-function and global consts in system-calls.js; let me know if you'd prefer these to be global.

jalkire:
Thank you for the quick review, @mykola-mokhnach! I have updated the PR based on your feedback.

lib/tools/system-calls.js
```diff
@@ -386,7 +392,10 @@ let isExecLocked = false;
  *                        You can also set the additional `exclusive` param
  *                        to `true` that assures no other parallel adb commands
  *                        are going to be executed while the current one is running
- * @return {string} - Command's stdout.
+ *                        You can set the `outputFormat` param to `stdout` to receive just the stdout
+ *                        output (default) or `full` to receive the entire stdout, stderr, and code
```
mykola-mokhnach, line 396: do we still have `code` property there?

lib/tools/system-calls.js
```diff
@@ -399,18 +408,22 @@ systemCallMethods.adbExec = async function adbExec (cmd, opts = {}) {
   opts.timeout = opts.timeout || this.adbExecTimeout || DEFAULT_ADB_EXEC_TIMEOUT;
   opts.timeoutCapName = opts.timeoutCapName || 'adbExecTimeout'; // For error message
 
+  const outputFormatStdout = 'stdout';
```
mykola-mokhnach, line 411: I would rather use something like

```
const EXEC_OUTPUT_FORMAT = {
  STDOUT: 'stdout',
  FULL: 'full',
};
```

mykola-mokhnach: it would probably even make sense to add this constant to `systemCallMethods` namespace, so it would be available from client modules

mykola-mokhnach (approved):

lib/tools/system-calls.js
jalkire: Oops, we do not

lib/tools/system-calls.js
jalkire: Added EXEC_OUTPUT_FORMAT to systemCallMethods and used it here + in tests

PR #577, opened 2021-09-23
feat: Add lock management utilities

mykola-mokhnach:
We could potentially speed up device unlocking on newer platforms with these helpers. Check https://newbedev.com/using-adb-to-change-the-screen-lock for more details.


KazuCocoa (approved):
interesting... I didn't know of the command

PR #579, opened 2021-09-25
feat: Add a method to set device lock credential

mykola-mokhnach:


lib/tools/lockmgmt.js
```diff
@@ -104,4 +107,41 @@ lockManagementMethods.isLockEnabled = async function isLockEnabled () {
...
+ * @throws {Error} If there was a failure while verifying input arguments or setting
+ * the credential
+ */
+lockManagementMethods.setLockCredential = async function setLockCredential (
+  credentialType, credential, oldCredential = null) {
+  if (!SUPPORTED_LOCK_CREDENTIAL_TYPES.includes(credentialType)) {
+    throw new Error(`Device lock credential type '${credentialType}' is unknown. ` +
+      `Only the following credential types are supported: ${SUPPORTED_LOCK_CREDENTIAL_TYPES}`);
+  }
+  if (_.isEmpty(credential) && !_.isInteger(credential)) {
```
KazuCocoa, line 133: should `password` case also be an integer?

mykola-mokhnach: no, it's just in case the one passes an integer while string is expected. isEmpty returns true in such case, however it is still possible to apply such value

KazuCocoa: ah... My thought was wrong... (i noticed...)
and didn't know `isEmpty(123)` etc returned true...

KazuCocoa (approved):

PR #581, opened 2021-10-09
chore: Add a pause to runInImeContext helper

mykola-mokhnach:
https://github.com/appium/appium/issues/15943


KazuCocoa:
no, that also changed the return value immediately to the set IME. It was not helpful to ensure the set IME was ready.

mykola-mokhnach:
thanks for checking it @KazuCocoa 

I hate doing it, but added a tiny delay after IME is set to make sure it's ready for the input. Also, fixed the input method itself to properly escape the text

KazuCocoa (approved):
This works...

PR #590, opened 2022-02-13
feat: Add a helper to create a universal apk from an .aab bundle

mykola-mokhnach:
I don't think we can easily add .apks support into Espresso driver, but we can still transform .aab files into .apk directly

### DIFF
--- a/lib/tools/aab-utils.js
+++ b/lib/tools/aab-utils.js
@@ -1,0 +1,101 @@
+import log from '../logger.js';
+import path from 'path';
+import { fs, tempDir, util } from '@appium/support';
+import LRU from 'lru-cache';
+import { unzipFile } from '../helpers.js';
+import AsyncLock from 'async-lock';
+import B from 'bluebird';
+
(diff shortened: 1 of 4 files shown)


lib/tools/aab-utils.js
```diff
@@ -0,0 +1,101 @@
...
+          universalApkPath = fullPath;
+        } else {
+          fileDeletionPromises.push(fs.rimraf(fullPath));
+        }
+      }
+      try {
+        await B.all(fileDeletionPromises);
+      } catch (ign) {}
+      if (!universalApkPath) {
+        log.debug(`The following items ae present in the .aab bundle: ${allFileNames}`);
```
jlipps, line 85: `are`

jlipps (approved):

lib/tools/aab-utils.js
mykola-mokhnach: fixed

KazuCocoa (approved):

PR #591, opened 2022-03-01
feat: add windowsVerbatimArguments in some exec

KazuCocoa:
for https://github.com/appium/appium-espresso-driver/issues/757

will test this tomorrow


KazuCocoa:
this did not affect exiting cmd for non-win env

PR #597, opened 2022-05-11
Add support for adb devices -l

NotWoods:
This change adds the ability to use `adb devices -l` instead of plain `adb devices`. This lets addition information get read when obtaining connected devices.

The `-l` is short for "long" so I used the same naming when creating the options object.

https://developer.android.com/studio/command-line/adb#devicestatus

### DIFF
--- a/lib/tools/system-calls.js
+++ b/lib/tools/system-calls.js
@@ -198,24 +198,36 @@ systemCallMethods.getBinaryFromPath = async function getBinaryFromPath (binaryNa
 };
 
 /**
+ * @typedef {Object} ConnectedDevicesOptions
+ * @property {?boolean} long [falsy] - Whether to get long output, which includes extra properties in each device.
+ * Akin to running `adb devices -l`.
+ */
+
+/**
  * @typedef {Object} Device
  * @property {string} udid - The device udid.
  * @property {string} state - Current device state, as it is visible in
  *                            _adb devices -l_ output.
+ * Will have additional properties if the `long` flag was set.
  */
 
 /**
  * Retrieve the list of devices visible to adb.
  *
+ * @param {?ConnectedDevicesOptions} opts [{}] - Additional options mapping.
  * @return {Array.<Device>} The list of devices or an empty list if
  *                          no devices are connected.
  * @throws {Error} If there was an error while listing devices.
  */
-systemCallMethods.getConnectedDevices = async function getConnectedDevices () {
+systemCallMethods.getConnectedDevices = async function getConnectedDevices (opts = {}) {
   log.debug('Getting connected devices');
   let stdout;
   try {
-    ({stdout} = await exec(this.executable.path, [...this.executable.defaultArgs, 'devices']));
+    const args = [...this.executable.defaultArgs, 'devices'];
+    if (opts.long) {
+      args.push('-l');
+    }
+    ({stdout} = await exec(this.executable.path, args));
   } catch (e) {
     throw new Error(`Error while getting connected devices. Original error: ${e.message}`);
   }
@@ -236,12 +248,19 @@ systemCallMethods.getConnectedDevices = async function getConnectedDevices () {
   const devices = stdout.split('\n')
     .map(_.trim)
     .filter((line) => line && !excludedLines.some((x) => line.includes(x)))
-    .reduce((acc, line) => {
+    .map((line) => {
       // state is "device", afaic
-      const [udid, state] = line.split(/\s+/);
-      acc.push({udid, state});
-      return acc;
-    }, []);
+      const [udid, state, ...description] = line.split(/\s+/);
+      const device = {udid, state};
+      if (opts.long) {
+        for (const entry of description) {
+          // each entry looks like key:value
+          const [key, value] = entry.split(':');
+          device[key] = value;
+        }
+      }
+      return device;
+    });
   if (_.isEmpty(devices)) {
     log.debug('No connected devices have been detected');
   } else {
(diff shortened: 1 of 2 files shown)


lib/tools/system-calls.js
```diff
@@ -197,25 +197,37 @@ systemCallMethods.getBinaryFromPath = async function getBinaryFromPath (binaryNa
...
+ * @property {?boolean} long [falsy] - Whether to get long output, which includes extra properties in each device.
+ * Akin to running `adb devices -l`.
+ */
+
 /**
  * @typedef {Object} Device
  * @property {string} udid - The device udid.
  * @property {string} state - Current device state, as it is visible in
  *                            _adb devices -l_ output.
+ * Will have additional properties if the `long` flag was set.
```
mykola-mokhnach, line 211: Would it be possible to enumerate known properties explicitly?

lib/tools/system-calls.js
```diff
@@ -197,25 +197,37 @@ systemCallMethods.getBinaryFromPath = async function getBinaryFromPath (binaryNa
...
  *                          no devices are connected.
  * @throws {Error} If there was an error while listing devices.
  */
-systemCallMethods.getConnectedDevices = async function getConnectedDevices () {
+systemCallMethods.getConnectedDevices = async function getConnectedDevices (opts = {}) {
   log.debug('Getting connected devices');
   let stdout;
   try {
-    ({stdout} = await exec(this.executable.path, [...this.executable.defaultArgs, 'devices']));
+    const args = [...this.executable.defaultArgs, 'devices'];
```
mykola-mokhnach, line 226: Please move args creation out of try/catch block

lib/tools/system-calls.js
```diff
@@ -236,12 +248,19 @@ systemCallMethods.getConnectedDevices = async function getConnectedDevices () {
...
-      acc.push({udid, state});
-      return acc;
-    }, []);
+      const [udid, state, ...description] = line.split(/\s+/);
+      const device = {udid, state};
+      if (opts.long) {
+        for (const entry of description) {
+          // each entry looks like key:value
+          const [key, value] = entry.split(':');
+          device[key] = value;
```
mykola-mokhnach, line 259: Might there be a situation where key name overrides uuid or state?

lib/tools/system-calls.js
```diff
@@ -236,12 +248,19 @@ systemCallMethods.getConnectedDevices = async function getConnectedDevices () {
...
     .map(_.trim)
     .filter((line) => line && !excludedLines.some((x) => line.includes(x)))
-    .reduce((acc, line) => {
+    .map((line) => {
       // state is "device", afaic
-      const [udid, state] = line.split(/\s+/);
-      acc.push({udid, state});
-      return acc;
-    }, []);
+      const [udid, state, ...description] = line.split(/\s+/);
```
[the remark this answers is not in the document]
NotWoods: As far as I can tell, description items will not contain spaces.

lib/tools/system-calls.js
NotWoods: There shouldn't be, I don't believe "serial", "udid", "uuid", or "state" are ever used as keys.

The corresponding ADB code can be found here: https://android.googlesource.com/platform/system/adb/+/refs/heads/master/transport.cpp#906

Keys are product, model, device, as well as devpath related info. `devpath` is harder to track but generally looks like `usb` or `transport`.

lib/tools/system-calls.js
```diff
@@ -199,7 +199,7 @@ systemCallMethods.getBinaryFromPath = async function getBinaryFromPath (binaryNa
 
 /**
  * @typedef {Object} ConnectedDevicesOptions
- * @property {?boolean} long [falsy] - Whether to get long output, which includes extra properties in each device.
+ * @property {?boolean} verbose [falsy] - Whether to get long output, which includes extra properties in each device.
```
mykola-mokhnach, line 202: Default value notation might be skipped for these already containing a question mark

lib/tools/system-calls.js
```diff
@@ -208,7 +208,15 @@ systemCallMethods.getBinaryFromPath = async function getBinaryFromPath (binaryNa
  * @property {string} udid - The device udid.
  * @property {string} state - Current device state, as it is visible in
  *                            _adb devices -l_ output.
- * Will have additional properties if the `long` flag was set.
+ */
+
+/**
+ * @typedef {Device} VerboseDevice Additional properties returned when `verbose` is true.
+ * @property {string} product
```
mykola-mokhnach, line 215: Could we also add a more detailed property description and maybe one simple example?

lib/tools/system-calls.js
```diff
@@ -577,12 +607,13 @@ systemCallMethods.getPortFromEmulatorString = function getPortFromEmulatorString
 /**
  * Retrieve the list of currently connected emulators.
  *
+ * @param {?ConnectedDevicesOptions} opts [{}] - Additional options mapping.
  * @return {Array.<Device>} The list of connected devices.
  */
-systemCallMethods.getConnectedEmulators = async function getConnectedEmulators () {
+systemCallMethods.getConnectedEmulators = async function getConnectedEmulators (opts) {
```
mykola-mokhnach, line 613: `opts = {}`

mykola-mokhnach (approved):

lib/tools/system-calls.js
```diff
@@ -236,12 +257,21 @@ systemCallMethods.getConnectedDevices = async function getConnectedDevices () {
   const devices = stdout.split('\n')
     .map(_.trim)
     .filter((line) => line && !excludedLines.some((x) => line.includes(x)))
-    .reduce((acc, line) => {
+    .map((line) => {
```
mykola-mokhnach, line 260: this logic could still use reduce

PR #598, opened 2022-05-11
Add cloneADB function

NotWoods:
Closes appium/appium#16932

Adds a `cloneADB` function that copies already initialized properties created in `ADB.createADB()`.

This was added in the top-level class since it uses the class constructor. It was attached directly to the class to avoid the prototype assignment warning.

### DIFF
--- a/lib/adb.js
+++ b/lib/adb.js
@@ -61,7 +61,8 @@ class ADB {
     const originalOptions = _.cloneDeep(_.pick(this, Object.keys(DEFAULT_OPTS)));
     const cloneOptions = _.defaultsDeep(opts, originalOptions);
 
-    // Reset default arguments created in the constructor
+    // Reset default arguments created in the constructor.
+    // Without this code, -H and -P can be injected into defaultArgs multiple times.
     const defaultArgs = cloneOptions.executable.defaultArgs;
     if (cloneOptions.remoteAdbHost && defaultArgs.includes('-H')) {
       defaultArgs.splice(defaultArgs.indexOf('-H'), 2);
--- a/lib/adb.js
+++ b/lib/adb.js
@@ -57,9 +57,19 @@ class ADB {
    * @param {object} opts - Additional options mapping to pass to the `ADB` constructor.
    * @returns {ADB} The resulting class instance.
    */
-  clone (opts) {
-    const adb = new ADB(opts);
-    _.defaultsDeep(adb, _.cloneDeep(this));
+  clone (opts = {}) {
+    const cloneOptions = _.defaultsDeep(opts, _.cloneDeep(this));
+
+    // Reset default arguments created in the constructor
+    const defaultArgs = cloneOptions.executable.defaultArgs;
+    if (cloneOptions.remoteAdbHost && defaultArgs.includes('-H')) {
+      defaultArgs.splice(defaultArgs.indexOf('-H'), 2);
+    }
+    if (defaultArgs.includes('-P')) {
+      defaultArgs.splice(defaultArgs.indexOf('-P'), 2);
+    }
+
+    const adb = new ADB(cloneOptions);
     return adb;
   }
 }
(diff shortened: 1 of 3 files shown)


lib/adb.js
```diff
@@ -50,6 +50,19 @@ class ADB {
...
     this.executable.defaultArgs.push('-P', this.adbPort);
   }
+
+  /**
+   * Create a new instance of `ADB` that inherits configuration from this `ADB` instance.
+   * This avoids the need to call `ADB.createADB()` multiple times.
+   * @param {object} opts - Additional options mapping to pass to the `ADB` constructor.
+   * @returns {ADB} The resulting class instance.
+   */
+  cloneADB (opts) {
```
mykola-mokhnach, line 60: How about simply  calling it clone?

mykola-mokhnach (approved):

mykola-mokhnach:
It is interesting what happened to the tests

lib/adb.js
```diff
@@ -50,6 +50,19 @@ class ADB {
...
   }
+
+  /**
+   * Create a new instance of `ADB` that inherits configuration from this `ADB` instance.
+   * This avoids the need to call `ADB.createADB()` multiple times.
+   * @param {object} opts - Additional options mapping to pass to the `ADB` constructor.
+   * @returns {ADB} The resulting class instance.
+   */
+  clone (opts) {
+    const adb = new ADB(opts);
```
mykola-mokhnach, line 61: Shouldn't we just copy opts from the existing instance instead of providing new ones?
It won't be an exact clone then

mykola-mokhnach:
Please fix the newly added e2e test before we could merge the commit

lib/adb.js
```diff
@@ -50,6 +50,28 @@ class ADB {
...
+    // Reset default arguments created in the constructor
+    const defaultArgs = cloneOptions.executable.defaultArgs;
+    if (cloneOptions.remoteAdbHost && defaultArgs.includes('-H')) {
+      defaultArgs.splice(defaultArgs.indexOf('-H'), 2);
+    }
+    if (defaultArgs.includes('-P')) {
+      defaultArgs.splice(defaultArgs.indexOf('-P'), 2);
+    }
+
+    const adb = new ADB(cloneOptions);
```
mykola-mokhnach, line 72: this intermediate variable seems redundant

lib/adb.js
```diff
@@ -50,6 +50,28 @@ class ADB {
...
   }
+
+  /**
+   * Create a new instance of `ADB` that inherits configuration from this `ADB` instance.
+   * This avoids the need to call `ADB.createADB()` multiple times.
+   * @param {object} opts - Additional options mapping to pass to the `ADB` constructor.
+   * @returns {ADB} The resulting class instance.
+   */
+  clone (opts = {}) {
+    const cloneOptions = _.defaultsDeep(opts, _.cloneDeep(this));
```
mykola-mokhnach, line 61: I'm not quite sure if it is a good idea to deep clone the whole instance in case we only need particular options. Would it be possible to only pick what's actually needed? Perhaps via listing the keys of DEFAULT_OPTS objects?

lib/adb.js
```diff
@@ -50,6 +50,28 @@ class ADB {
...
+
+    // Reset default arguments created in the constructor
```
mykola-mokhnach, line 63: why this is necessary?

NotWoods: Without this code, `-H` and `-P` can be injected into `defaultArgs` multiple times.

lib/adb.js
NotWoods: Updated to use `_.pick`. I kept `_.cloneDeep` as well so the deep options are cloned, although the shallow clone is redundant.

lib/adb.js
mykola-mokhnach: I think we might just add this statement as a comment to make it more obvious

mykola-mokhnach (approved):

PR #599, opened 2022-05-18
feat: Add a helper to resolve launchable activity name

mykola-mokhnach:
Based on https://stackoverflow.com/questions/12698814/get-launchable-activity-name-of-package-from-adb

### DIFF
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -211,17 +211,17 @@ methods.isValidClass = function isValidClass (classString) {
 };
 
 /**
- * Fetches the fully qualified name of the main activity for the
+ * Fetches the fully qualified name of the launchable activity for the
  * given package. It is expected the package is already installed on
  * the device under test.
  *
  * ! This method only works since Android 7 (API level 24)
  *
  * @param {string} pkg - The target package identifier
- * @return {string} Fully qualified name of the main activity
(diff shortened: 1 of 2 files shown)


lib/tools/adb-commands.js
```diff
@@ -210,6 +210,30 @@ methods.isValidClass = function isValidClass (classString) {
...
+ * given package. It is expected the package is already installed on
+ * the device under test.
+ *
+ * ! This method only works since Android 7 (API level 24)
+ *
+ * @param {string} pkg - The target package identifier
+ * @return {string} Fully qualified name of the main activity
+ * @throws {Error} If there was an error while resolving the activity name
+ */
+methods.resolveMainActivity = async function resolveMainActivity (pkg) {
```
KazuCocoa, line 224: What about `resolveLaunchableActivity` or `resolveLauncherActivity`?
Maybe the result is an activity name that has `LAUNCHER` category. Then, 'launchable' or 'launcher' is easier to imagine the result is a launcher.

mykola-mokhnach: changed

KazuCocoa (approved):

jlipps (commented):
nice! i was just thinking about this the other day, that there should be a way to start an android app via package alone and determine the launch activity automatically.

PR #601, opened 2022-05-24
fix: Try to resolve the name of the launchable activity from the manifest for older Android versions

mykola-mokhnach:


KazuCocoa:
I'll do tests with a couple of real apps (just a note I notice this pr)

mykola-mokhnach:
> I'll do tests with a couple of real apps (just a note I notice this pr)

Thanks Kazu. The only possible issue I know about could be if we cannot extract the base .apk from the already installed app (for some apps it is not included) and the app exposes multiple activities for API < 24

KazuCocoa (approved):

PR #604, opened 2022-06-28
feat: Tune active activity/package names parsing for multi-screen systems

mykola-mokhnach:
https://github.com/appium/appium/issues/17106


KazuCocoa (approved):
i don't have the device, otherwise lgtm

PR #611, opened 2022-08-26
tests: Fix ADB instance import

mykola-mokhnach:
We should never import from index.js in tests, but rather directly from corresponding source modules.


KazuCocoa (approved):

jlipps:
why, out of curiosity? wouldn't importing from the same place that users import theoretically show up more possible bugs?

mykola-mokhnach:
This is what @boneskull adviced. A similar import was creating weird test failure issues in WebDriverAgent module, so I would prefer to not repeat same mistakes.

PR #623, opened 2022-10-22
fix: defaultArgs and path call in refreshGeoLocationCache

KazuCocoa:
https://github.com/appium/appium/issues/17641


KazuCocoa:
btw, on a real device, `mobile:refreshGpsCache` was not alway work.

Then, restarting 'io.appium.settings' helped to reset the mocked location. (Android 12)

```
> driver.terminate_app 'io.appium.settings'
=> true
> driver.activate_app 'io.appium.settings'
=> nil
```

mykola-mokhnach (approved):

PR #635, opened 2023-02-24
settings client commands use adbExecTimeout fixes #18248

magwas:


lib/tools/settings-client-commands.js
```diff
@@ -64,7 +64,7 @@ commands.requireRunningSettingsApp = async function requireRunningSettingsApp (o
 
   log.debug('Starting Appium Settings app');
   const {
-    timeout = 5000,
+    timeout = this.adbExecTimeout,
```
mykola-mokhnach, line 67: I don't think it is a good idea to mix adb exec timeout this this one. Consider customizing it on a higher level (e.g. appium-android-driver)

magwas: You mean creating another capability, and using that here?
As I understand it needs changes both here and appium-android-driver.
I can do it, but as it is two different pull requests, I need a capability name and an okay from core maintainers.

mykola-mokhnach:
closed as obsolete

PR #661, opened 2023-06-01
feat: export types

boneskull:
Because of the heavy reliance on this module in `appium-android-driver`, and the fact that this module is at least partially typed, I figured it would be helpful to actually ship those types.

That is what this does.  What it _doesn't_ do is typecheck the JS.


boneskull:
Other than compiling with typescript, the main thing here is that I merged the `lib/tools/adb-commands` module with the `ADB` class definition.

boneskull:
OK, I figured out how to do this using even less code.  That said, I needed to change a bunch of `@return` tags because they were not returning `Promise` values.

lib/adb.ts
```diff
@@ -73,24 +108,22 @@ class ADB {
...
   }
-  return adb;
-};
+}
 
 // add all the methods to the ADB prototype
-for (const [fnName, fn] of _.toPairs(methods)) {
-  ADB.prototype[fnName] = fn;
-}
+Object.assign(ADB.prototype, methods);
```
mykola-mokhnach, line 127: isn't this always complaining in eslint, that we cannot assign directly to the prototype?

lib/stubs.ts
```diff
@@ -0,0 +1 @@
+declare module 'adbkit-apkreader';
```
mykola-mokhnach, line 1: what is the purpose of this stub?

lib/tools/adb-commands.js
```diff
@@ -1700,7 +1700,7 @@ methods.screenrecord = function screenrecord (destination, options = {}) {
  *
  * @param {string} ime - Valid IME identifier
  * @param {Function} fn - Function to execute
- * @returns {*} The result of the given function
+ * @returns {Promise<any>} The result of the given function
```
mykola-mokhnach, line 1703: I assume we could generify these args to point out the result would be of the same type that the passed function  returns

test/unit/apk-utils-specs.js
```diff
@@ -2,7 +2,8 @@ import chai from 'chai';
 import chaiAsPromised from 'chai-as-promised';
 import * as teen_process from 'teen_process';
 import { fs } from '@appium/support';
-import ADB from '../../lib/adb';
+// eslint-disable-next-line import/no-unresolved
```
mykola-mokhnach, line 5: why should it complain about unresolved imports?

lib/tools/adb-commands.js
```diff
@@ -1632,7 +1632,7 @@ methods.getSetting = async function getSetting (namespace, setting) {
  * operation may take up to several minutes.
  *
  * @param {?number} timeout [120000] - Command timeout in milliseconds
```
mykola-mokhnach, line 1634: should we also fix types that start with a question mark (e.g. move it to the end)?

jlipps (approved):
LGTM, though curious about the answers to most of @mykola-mokhnach's questions too

lib/adb.ts
boneskull: it is.  I think I should probably disable that warning, because I think we probably _should_ do it this way if we're using mixins.

lib/stubs.ts
boneskull: this tells TS that "yes, this module exists, and no, it does not have any types, so treat it as `any`".

lib/tools/adb-commands.js
boneskull: yes, we could probably use generics here. I didn't look at these types too closely; trying to do the bare minimum needed to get everything working

test/unit/apk-utils-specs.js
boneskull: because without our eslint TS configuration, eslint cannot find a `.ts` file.

lib/tools/adb-commands.js
boneskull: yes, but I'm not going to do that right now.  it seems to interpret `?` anywhere in the type as `null`.  either way, we probably should not be using `?` since it means one thing in jsdoc (`null`) and another thing in TS (unset or `undefined`) and is thus confusing.

lib/adb.ts
mykola-mokhnach: would you like to update that check in the global eslint config?

lib/stubs.ts
mykola-mokhnach: maybe add a comment there to make it cleaner?

boneskull:
Current dependencies on/for this PR:
* master
  * **PR #661** <a href="https://app.graphite.dev/github/pr/appium/appium-adb/661" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium-adb/661?utm_source=stack-comment).

PR #676, opened 2023-06-29
feat(types): export more types

boneskull:
- fix `getApksignerForOs` return type
- move options interface into its own file
- export `ApkSigningCommands` interface and attach to `ADB`
- export everything in `apks-utils` module; fix return values
- fix types in `tools/apk-signing`
- use TS v5.0.x

### DIFF
--- a/lib/tools/apks-utils.js
+++ b/lib/tools/apks-utils.js
@@ -48,7 +48,7 @@ process.on('exit', () => {
  * @param {string} apks - The full path to the .apks file
  * @param {string|Array<String>} dstPath - The relative path to the destination file,
  * which is going to be extracted, where each path component is an array item
- * @returns {string} Full path to the extracted file
+ * @returns {Promise<string>} Full path to the extracted file
  * @throws {Error} If the requested item does not exist in the extracted archive or the provides
  * apks file is not a valid bundle
  */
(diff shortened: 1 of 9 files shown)


lib/tools/apks-utils.js
```diff
@@ -146,26 +146,26 @@ apksUtilsMethods.getDeviceSpec = async function getDeviceSpec (specLocation) {
...
  * is completed under Android 6+.
- * @property {boolean} partialInstall [false] - Install apks partially. It is used for 'install-multiple'.
+ * @property {boolean} [partialInstall=false] - Install apks partially. It is used for 'install-multiple'.
  * https://android.stackexchange.com/questions/111064/what-is-a-partial-application-install-via-adb
  */
 
 /**
  * Installs the given apks into the device under test
  *
  * @param {Array<string>} apkPathsToInstall - The full paths to install apks
```
mykola-mokhnach, line 167: string[]

lib/tools/apks-utils.js
```diff
@@ -146,26 +146,26 @@ apksUtilsMethods.getDeviceSpec = async function getDeviceSpec (specLocation) {
...
- * @param {?installMultipleApksOptions} options - Installation options
+ * @param {?InstallMultipleApksOptions} options - Installation options
```
mykola-mokhnach, line 168: The question mark is redundant

package.json
```diff
@@ -105,7 +105,7 @@
     "sinon": "^15.0.0",
     "temp": "^0.x",
     "ts-node": "^10.9.1",
-    "typescript": "^5.1.3"
+    "typescript": "~5.0.4"
```
mykola-mokhnach, line 108: what is the problem with higher versions of this module?

mykola-mokhnach (approved):

package.json
boneskull: it's incompatible with the version of typedoc we're currently stuck at

lib/tools/apks-utils.js
boneskull: I didn't write these types

PR #683, opened 2023-07-18
feat: add hint of signature scheme

KazuCocoa:
Adds a guide to check the signature scheme if the installation command got INSTALL_PARSE_FAILED_NO_CERTIFICATES

closes https://github.com/appium/appium/issues/18878

### DIFF
--- a/lib/tools/apks-utils.js
+++ b/lib/tools/apks-utils.js
@@ -275,6 +275,32 @@ apksUtilsMethods.isTestPackageOnlyError = function isTestPackageOnlyError (outpu
   return /\[INSTALL_FAILED_TEST_ONLY\]/.test(output);
 };
 
+
+/**
+ * Returns true if the given output had INSTALL_PARSE_FAILED_NO_CERTIFICATES word.
+ * Usually the apk does not have proper signature scheme. It needs to be signed properly,
+ * or the apk needs to be updated.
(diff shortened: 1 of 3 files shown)


lib/tools/apks-utils.js
```diff
@@ -275,6 +275,32 @@ apksUtilsMethods.isTestPackageOnlyError = function isTestPackageOnlyError (outpu
...
+/**
+ * Returns true if the given output had INSTALL_PARSE_FAILED_NO_CERTIFICATES word.
+ * Usually the apk does not have proper signature scheme. It needs to be signed properly,
+ * or the apk needs to be updated.
+ * https://developer.android.com/tools/apksigner#options-sign
+ *
+ * @param {string} output - The adb command output
+ * @returns {boolean} If the output includes INSTALL_PARSE_FAILED_NO_CERTIFICATES
+ */
+apksUtilsMethods.isInsufficientSignatureScheme = function isInsufficientSignatureScheme (output) {
```
mykola-mokhnach, line 288: I am not quote convinced we need to mixin these helpers directly into ADB instance. I would rather keep themon the private level

KazuCocoa:
hm, maybe the error message, INSTALL_PARSE_FAILED_NO_CERTIFICATES, should be sufficient to checking the reason on the internet. the adb command will also continue up to date

PR #715, opened 2024-01-08
feat: add isSettingsAppServiceRunningInForeground to check the settings' service existence better

KazuCocoa:
It looks like a command below launches `io.appium.settings` process, which can be found by `[debug] [ADB] Running '/Users/kazu/Library/Android/sdk/platform-tools/adb -P 5037 -s D0AA002182JC0202126 shell pgrep -f \(\[\[:blank:\]\]\|\^\)io\.appium\.settings\(\[\[:blank:\]\]\|\$\)'`, but the app was not started in the background properly (no the app in the notification). It causes io.appium.settings process did not start properly situation, thus for example `driver.location` to get the location/set location causes an error because the `io.appium.settings` was not started properly with `this.startApp`.

We must start the settings app forcefully once with `this.startApp` to start the settings app process properly.

```
[debug] [ADB] Running '/Users/kazu/Library/Android/sdk/platform-tools/adb -P 5037 -s D0AA002182JC0202126 shell cmd notification allow_listener io.appium.settings/.NLService'
```

```
  await this.startApp({
    pkg: SETTINGS_HELPER_ID,
    activity: SETTINGS_HELPER_MAIN_ACTIVITY,
    action: 'android.intent.action.MAIN',
    category: 'android.intent.category.LAUNCHER',
    stopApp: false,
    waitForLaunch: false,
  });
```

I also found `-g` option could launch the io.appium.settings app process on Android 11. I saw the behavior was not always, but actually occurred. Then, `await this.processExists(SETTINGS_HELPER_ID)` skipped the io.appium.settings process with `this.startApp` while uia2 driver must have started the settings app with `this.startApp` to make the get location etc work properly

https://github.com/appium/appium-android-driver/pull/895 is this PR's usage

### DIFF
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -1953,6 +1953,16 @@ methods.takeScreenshot = async function takeScreenshot (displayId) {
   return stdout;
 };
 
+/**
+ * Return the result of 'adb shell dumpsys activity services <packagename>'
+ *
+ * @this {import('../adb.js').ADB}
+ * @returns {Promise<string>} the result of 'adb shell dumpsys activity services <packagename>'
+ */
+methods.getActivityService = async function getActivityService (pkgName) {
+  return await this.shell(['dumpsys', 'activity', 'services', pkgName]);
+};
+
 export default methods;
 
 /**
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -1953,16 +1953,6 @@ methods.takeScreenshot = async function takeScreenshot (displayId) {
   return stdout;
 };
 
-/**
- * Return the result of 'adb shell dumpsys activity services <packagename>'
- *
- * @this {import('../adb.js').ADB}
- * @returns {Promise<string>} the result of 'adb shell dumpsys activity services <packagename>'
- */
-methods.getActivityService = async function getActivityService (pkgName) {
-  return await this.shell(['dumpsys', 'activity', 'services', pkgName]);
-};
-
 export default methods;
 
 /**
--- a/lib/tools/settings-client-commands.js
+++ b/lib/tools/settings-client-commands.js
@@ -59,7 +59,7 @@ commands.isSettingsAppServiceRunningInForeground = async function isSettingsAppS
 
   // The foreground service is available since api level 26,
   // thus this method works only for api level 26+.
-  const output = await this.getActivityService(SETTINGS_HELPER_ID);
+  const output = await this.shell(['dumpsys', 'activity', 'services', SETTINGS_HELPER_ID]);
   return output.includes(FOREGROUND_APP_KEYWORD);
 };
 
--- a/lib/tools/settings-client-commands.js
+++ b/lib/tools/settings-client-commands.js
@@ -52,12 +52,12 @@ const commands = {};
 commands.hasRunningForegroundService = async function hasRunningForegroundService () {
   try {
     const output = await this.getActivityService(SETTINGS_HELPER_ID);
-    return (output.includes(FORE_GROUND_APP_KEYWORD))
+    return (output.includes(FORE_GROUND_APP_KEYWORD));
   } catch (e) {
     log.warn(`Got an error in getting the foreground service state: ${e.message}`);
   }
   return false;
-}
+};
 
 /**
  * @typedef {Object} SettingsAppStartupOptions
@@ -77,16 +77,15 @@ commands.hasRunningForegroundService = async function hasRunningForegroundServic
  * @returns {Promise<import('../adb.js').ADB>} self instance for chaining
  */
 commands.requireRunningSettingsApp = async function requireRunningSettingsApp (opts = {}) {
-  const {
-    timeout = 5000,
-    shouldRestoreCurrentApp = false
-  } = opts;
-
   if (await this.hasRunningForegroundService()) {
     return this;
   }
 
   log.debug('Starting Appium Settings app');
+  const {
+    timeout = 5000,
+    shouldRestoreCurrentApp = false
+  } = opts;
   let appPackage;
   if (shouldRestoreCurrentApp) {
     try {
--- a/lib/tools/settings-client-commands.js
+++ b/lib/tools/settings-client-commands.js
@@ -59,6 +59,9 @@ commands.isSettingsAppServiceRunningInForeground = async function isSettingsAppS
 
   // The foreground service is available since api level 26,
   // thus this method works only for api level 26+.
+
+  // 'dumpsys activity services <package>' had slightly better performance
+  // than 'dumpsys activity services' and parsing the foreground apps.
   const output = await this.shell(['dumpsys', 'activity', 'services', SETTINGS_HELPER_ID]);
   return output.includes(FOREGROUND_APP_KEYWORD);
 };
--- a/lib/tools/settings-client-commands.js
+++ b/lib/tools/settings-client-commands.js
@@ -49,7 +49,7 @@ const commands = {};
  * since it does not have the foreground service.
  *
  * @this {import('../adb.js').ADB}
-* @throws {Error} If the method gets an error in the adb shell execution.
+ * @throws {Error} If the method gets an error in the adb shell execution.
  * @returns {Promise<boolean>} Return true if the device has running settings app foreground service.
  */
 commands.isSettingsAppServiceRunningInForeground = async function isSettingsAppServiceRunningInForeground () {
--- a/lib/tools/settings-client-commands.js
+++ b/lib/tools/settings-client-commands.js
@@ -39,7 +39,7 @@ const GPS_CACHE_REFRESHED_LOGS = [
 
 const GPS_COORDINATES_PATTERN = /data="(-?[\d.]+)\s+(-?[\d.]+)\s+(-?[\d.]+)"/;
 
-const FORE_GROUND_APP_KEYWORD = 'isForeground=true';
+const FOREGROUND_APP_KEYWORD = 'isForeground=true';
 
 const commands = {};
 
@@ -49,22 +49,18 @@ const commands = {};
  * since it does not have the foreground service.
  *
  * @this {import('../adb.js').ADB}
+* @throws {Error} If the method gets an error in the adb shell execution.
  * @returns {Promise<boolean>} Return true if the device has running settings app foreground service.
  */
-commands.hasRunningSettingsAppForegroundService = async function hasRunningSettingsAppForegroundService () {
+commands.isSettingsAppServiceRunningInForeground = async function isSettingsAppServiceRunningInForeground () {
   if (await this.getApiLevel() < 26) {
     return await this.processExists(SETTINGS_HELPER_ID);
   }
 
   // The foreground service is available since api level 26,
   // thus this method works only for api level 26+.
-  try {
-    const output = await this.getActivityService(SETTINGS_HELPER_ID);
-    return (output.includes(FORE_GROUND_APP_KEYWORD));
-  } catch (e) {
-    log.warn(`Got an error in getting the foreground service state: ${e.message}`);
-  }
-  return false;
+  const output = await this.getActivityService(SETTINGS_HELPER_ID);
+  return output.includes(FOREGROUND_APP_KEYWORD);
 };
 
 /**
@@ -89,7 +85,7 @@ commands.hasRunningSettingsAppForegroundService = async function hasRunningSetti
  * @returns {Promise<import('../adb.js').ADB>} self instance for chaining
  */
 commands.requireRunningSettingsApp = async function requireRunningSettingsApp (opts = {}) {
-  if (await this.hasRunningSettingsAppForegroundService()) {
+  if (await this.isSettingsAppServiceRunningInForeground()) {
     return this;
   }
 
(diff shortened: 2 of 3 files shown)


lib/tools/settings-client-commands.js
```diff
@@ -59,15 +61,17 @@ const commands = {};
...
-  }
-
-  log.debug('Starting Appium Settings app');
   const {
     timeout = 5000,
     shouldRestoreCurrentApp = false,
+    forceStartSettingsApp = false
   } = opts;
+
+  if (forceStartSettingsApp === false && await this.processExists(SETTINGS_HELPER_ID)) {
```
mykola-mokhnach, line 70: forceStartSettingsApp === false => !forceStartSettingsApp

lib/tools/settings-client-commands.js
```diff
@@ -59,15 +61,17 @@ const commands = {};
...
+    return this;
+  }
+
+  log.debug('Starting Appium Settings app');
```
mykola-mokhnach, line 74: should stopApp arg below be set to true if forceStartSettingsApp is true?

KazuCocoa:
I figured out the root cause a bit more. `io.appium.settings/.NLService` service was running while the app process was stopped by settings' "force stop," for example. We should have `io.appium.settings/.ForegroundService` as the active foreground apps with `adb shell dumpsys activity services` instead of `this.processExists` for settings app since `io.appium.settings/.NLService` could be running but the `io.appium.settings/.ForegroundService` was not running. It cause the settings app no functions but the process existed.

```
...
Active foreground apps - user 0:
  #0: io.appium.settings
    mNumActive=1 mAppOnTop=false mShownWhileTop=true mShownWhileScreenOn=true
    mStartTime=-7m24s274ms mStartVisibleTime=-7m24s274ms
```

mykola-mokhnach:
> I figured out the root cause a bit more. `io.appium.settings/.NLService` service was running while the app process was stopped by settings' "force stop," for example. We should have `io.appium.settings/.ForegroundService` as the active foreground apps with `adb shell dumpsys activity services` instead of `this.processExists` for settings app since `io.appium.settings/.NLService` could be running but the `io.appium.settings/.ForegroundService` was not running. It cause the settings app no functions but the process existed.
> 
> ```
> ...
> Active foreground apps - user 0:
>   #0: io.appium.settings
>     mNumActive=1 mAppOnTop=false mShownWhileTop=true mShownWhileScreenOn=true
>     mStartTime=-7m24s274ms mStartVisibleTime=-7m24s274ms
> ```

that's a great find.
Lets then add a separate method to verify if settings app and corresponding services are running into appium-adb and then call it instead of processExists in the android driver

mykola-mokhnach:
I also think it would make sense to move more javascript code into io.appium.settings package. We should not keep this logic in appium-adb and have it abstract

mykola-mokhnach:
> I also think it would make sense to move more javascript code into io.appium.settings package. We should not keep this logic in appium-adb and have it more abstract

Will probably do it later myself

KazuCocoa:
Note for this. It looks like the simplest way was `isForeground=true` exists with the `adb shell dumpsys activity services  io.appium.settings` result. The `adb shell dumpsys activity services  io.appium.settings` prints not the foreground only but at least if the package includes a foreground process, it looks like having `isForeground=true`, thus maybe `isForeground=true` existing check should be sufficient for this purpose.

```
% adb shell dumpsys activity services  io.appium.settings
ACTIVITY MANAGER SERVICES (dumpsys activity services)
  User 0 active services:
  * ServiceRecord{f0ad90b u0 io.appium.settings/.NLService}
    intent={act=android.service.notification.NotificationListenerService cmp=io.appium.settings/.NLService}
    packageName=io.appium.settings
    processName=io.appium.settings
    permission=android.permission.BIND_NOTIFICATION_LISTENER_SERVICE
    baseDir=/data/app/~~fHuRc6u9ehtAcXvuXy-fiw==/io.appium.settings-wJRwd1HrrbVG5ZINWuHi5Q==/base.apk
    dataDir=/data/user/0/io.appium.settings
    app=ProcessRecord{1d61746 18302:io.appium.settings/u0a320}
    whitelistManager=true
    allowWhileInUsePermissionInFgs=true
    startForegroundCount=0
    recentCallingPackage=android
    createTime=-6m21s859ms startingBgTimeout=--
    lastActivity=-6m21s783ms restartTime=-6m21s783ms createdFromFg=true
    Bindings:
    * IntentBindRecord{a5d675f CREATE}:
      intent={act=android.service.notification.NotificationListenerService cmp=io.appium.settings/.NLService}
      binder=android.os.BinderProxy@1be25ac
      requested=true received=true hasBound=true doRebind=false
      * Client AppBindRecord{c78a275 ProcessRecord{3f853b1 1847:system/1000}}
        Per-process Connections:
          ConnectionRecord{fbf1188 u0 CR FGS !PRCP io.appium.settings/.NLService:@339692b}
    All Connections:
      ConnectionRecord{fbf1188 u0 CR FGS !PRCP io.appium.settings/.NLService:@339692b}

  * ServiceRecord{e7a180b u0 io.appium.settings/.ForegroundService}
    intent={act=start cmp=io.appium.settings/.ForegroundService}
    packageName=io.appium.settings
    processName=io.appium.settings
    permission=android.permission.FOREGROUND_SERVICE
    baseDir=/data/app/~~fHuRc6u9ehtAcXvuXy-fiw==/io.appium.settings-wJRwd1HrrbVG5ZINWuHi5Q==/base.apk
    dataDir=/data/user/0/io.appium.settings
    app=ProcessRecord{1d61746 18302:io.appium.settings/u0a320}
    allowWhileInUsePermissionInFgs=true
    startForegroundCount=1
    recentCallingPackage=io.appium.settings
    isForeground=true foregroundId=1 foregroundNoti=Notification(channel=main_channel shortcut=null contentView=null vibrate=null sound=null defaults=0x0 flags=0x62 color=0x00000000 vis=PRIVATE)
    createTime=-5m1s703ms startingBgTimeout=--
    lastActivity=-5m1s702ms restartTime=-5m1s702ms createdFromFg=true
    startRequested=true delayedStop=false stopIfKilled=false callStart=true lastStartId=1

  Connection bindings to services:
  * ConnectionRecord{fbf1188 u0 CR FGS !PRCP io.appium.settings/.NLService:@339692b}
    binding=AppBindRecord{c78a275 io.appium.settings/.NLService:system}
    conn=android.app.LoadedApk$ServiceDispatcher$InnerConnection@339692b flags=0x5000101
```

Ideally we want to get the list of `Active foreground apps` but maybe adb does not have the exact adb command option fo the only the list

KazuCocoa:
Added tests. Maybe this is currently what we can do to ensure if io.appium.settings package is having a foreground service as the service

lib/tools/settings-client-commands.js
```diff
@@ -39,8 +39,26 @@ const GPS_CACHE_REFRESHED_LOGS = [
...
+
 const commands = {};
 
+/**
+ * If the io.appium.settings package has running foreground service.
+ *
+ * @this {import('../adb.js').ADB}
+ * @returns {Promise<boolean>} Return true if the device has running settings app foreground service.
+ */
+commands.hasRunningSettingsAppForegroundService = async function hasRunningSettingsAppForegroundService () {
```
KazuCocoa, line 52: note: api level 28+ worked with the `Active foreground apps` (this way), but at least Android 7 seems not for the case. The main issue I found by `cmd notification` was api level 29+, so this method can follow the old pid existing way for api level 27 and lower I think

KazuCocoa: (updated. So api level 25 and lower can keep the pid way)

KazuCocoa:
checked the `activity services io.appium.settings` on various API levels

lib/tools/adb-commands.js
```diff
@@ -1953,6 +1953,16 @@ methods.takeScreenshot = async function takeScreenshot (displayId) {
...
 };
 
+/**
+ * Return the result of 'adb shell dumpsys activity services <packagename>'
+ *
+ * @this {import('../adb.js').ADB}
+ * @returns {Promise<string>} the result of 'adb shell dumpsys activity services <packagename>'
+ */
+methods.getActivityService = async function getActivityService (pkgName) {
+  return await this.shell(['dumpsys', 'activity', 'services', pkgName]);
```
mykola-mokhnach, line 1963: should we somehow parse the output into a valid list? Otherwise this wrapper does not really do much

lib/tools/settings-client-commands.js
```diff
@@ -39,8 +39,34 @@ const GPS_CACHE_REFRESHED_LOGS = [
 
 const GPS_COORDINATES_PATTERN = /data="(-?[\d.]+)\s+(-?[\d.]+)\s+(-?[\d.]+)"/;
 
+const FORE_GROUND_APP_KEYWORD = 'isForeground=true';
```
mykola-mokhnach, line 42: FORE_GROUND_APP_KEYWORD  -> FOREGROUND_APP_KEYWORD

lib/tools/settings-client-commands.js
```diff
@@ -39,8 +39,34 @@ const GPS_CACHE_REFRESHED_LOGS = [
...
+commands.hasRunningSettingsAppForegroundService = async function hasRunningSettingsAppForegroundService () {
+  if (await this.getApiLevel() < 26) {
+    return await this.processExists(SETTINGS_HELPER_ID);
+  }
+
+  // The foreground service is available since api level 26,
+  // thus this method works only for api level 26+.
+  try {
+    const output = await this.getActivityService(SETTINGS_HELPER_ID);
+    return (output.includes(FORE_GROUND_APP_KEYWORD));
```
mykola-mokhnach, line 63: parentheses are redundant

lib/tools/settings-client-commands.js
```diff
@@ -39,8 +39,34 @@ const GPS_CACHE_REFRESHED_LOGS = [
...
+  } catch (e) {
+    log.warn(`Got an error in getting the foreground service state: ${e.message}`);
```
mykola-mokhnach, line 65: let it throw an error - we should know when it does not work

lib/tools/settings-client-commands.js
```diff
@@ -39,8 +39,34 @@ const GPS_CACHE_REFRESHED_LOGS = [
...
 
+/**
+ * If the io.appium.settings package has running foreground service.
+ * It returns the io.appium.settings's process existence for api level 25 and lower
+ * since it does not have the foreground service.
+ *
+ * @this {import('../adb.js').ADB}
+ * @returns {Promise<boolean>} Return true if the device has running settings app foreground service.
+ */
+commands.hasRunningSettingsAppForegroundService = async function hasRunningSettingsAppForegroundService () {
```
mykola-mokhnach, line 54: how about isSettingsAppForegroundServiceRunning?

test/unit/adb-commands-specs.js
```diff
@@ -1244,4 +1244,119 @@ describe('adb commands', withMocks({adb, logcat, teen_process, net}, function (m
       await adb.setDefaultHiddenApiPolicy();
     });
   });
+
+  describe('hasRunningSettingsAppForegroundService', function () {
```
mykola-mokhnach, line 1248: 👍

lib/tools/adb-commands.js
KazuCocoa: Yea, initially i thought the command returned like list (part of output) below which was available via `dumpsys activity services` (no package name). 

```
Active foreground apps - user 0:
  #0: io.appium.settings
    mNumActive=1 mAppOnTop=false mShownWhileTop=true mShownWhileScreenOn=true
    mStartTime=-16m59s209ms mStartVisibleTime=-16m59s209ms
```

so something like `getActiveForegroundAppa`. (maybe we can create a new method for this later)

`dumpsys activity services io.appium.settings` had slightly better performance than `dumpsys activity services` (based on bash's `time`), so as the initial implementation, I made simple output with better performance way right now.

lib/tools/settings-client-commands.js
KazuCocoa: https://github.com/appium/appium-adb/pull/715/commits/ddf5635a247c4896dd19c703a13c11f567165b84

lib/tools/settings-client-commands.js
KazuCocoa: https://github.com/appium/appium-adb/pull/715/commits/ddf5635a247c4896dd19c703a13c11f567165b84

lib/tools/settings-client-commands.js
KazuCocoa: https://github.com/appium/appium-adb/pull/715/commits/ddf5635a247c4896dd19c703a13c11f567165b84

lib/tools/settings-client-commands.js
KazuCocoa: https://github.com/appium/appium-adb/pull/715/commits/ddf5635a247c4896dd19c703a13c11f567165b84

lib/tools/adb-commands.js
mykola-mokhnach: I mean it makes sense to have the actual adb.shell call as a part of `isSettingsAppServiceRunningInForeground` method. We anyway don't use it anywhere else

KazuCocoa: I see, makes sense

lib/tools/settings-client-commands.js
```diff
@@ -39,8 +39,33 @@ const GPS_CACHE_REFRESHED_LOGS = [
...
+
 const commands = {};
 
+/**
+ * If the io.appium.settings package has running foreground service.
+ * It returns the io.appium.settings's process existence for api level 25 and lower
+ * since it does not have the foreground service.
+ *
+ * @this {import('../adb.js').ADB}
+* @throws {Error} If the method gets an error in the adb shell execution.
```
mykola-mokhnach, line 52: indentation

mykola-mokhnach (approved):

lib/tools/settings-client-commands.js
KazuCocoa: oo..

PR #732, opened 2024-05-02
fix: Add one more condition to the lock state detection

mykola-mokhnach:
Based on https://stackoverflow.com/questions/19611945/obtain-screen-status-using-adb


KazuCocoa (approved):
nice to find

PR #744, opened 2024-05-16
chore: Bump eslint dependency

mykola-mokhnach:


mykola-mokhnach:
I'm going to apply similar changes to other packages as eslint is anyway imported by `@appium/eslint-config-appium`, so there is no need to reimport it in depending packages and thus deal with version changes

KazuCocoa (approved):

PR #745, opened 2024-05-25
feat: enable window animation via settings

KazuCocoa:
This settings command worked on Android 9, 10, 11, 12, 13 and 14 real devices, so I think at least this can be alternative of existing animation one. I haven't tested for older than API level 26. So for now, we could use this as https://github.com/appium/appium-espresso-driver/pull/1007

### DIFF
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -985,14 +985,9 @@ methods.isAnimationOn = async function isAnimationOn () {
  *   - window_animation_scale
  *
  * @this {import('../adb.js').ADB}
- * @param {number} value Animation scale value (int or floating) to set.
- *                       The animation scale in Android code seems Java float,
- *                       but acceptable values as the settings command could be overr the Java float max/min.
- *                       The animation behaves as same as zero animation scale for negative number scale value.
- *                       For example, when the scale value is large number like 100,
- *                       the window animation of the Android GUI behaves very slow.
- *                       You may need to restart the device after setting the value with small number like 1
- *                       to avoid long waiting time to complete the animation.
+ * @param {number} value Animation scale value (int or float) to set.
+ *                       The minimum value of zero disables animations.
+ *                       By increasing the value animations become (slower/faster).
  * @return {Promise<void>}
  * @throws {Error} If the adb setting command raises an exception.
  */
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -976,15 +976,19 @@ methods.isAnimationOn = async function isAnimationOn () {
 };
 
 /**
- * Set window animation scale with the given velue via adb shell settings command.
+ * Set animation scale with the given velue via adb shell settings command.
+ *   - animator_duration_scale
+ *   - transition_animation_scale
+ *   - window_animation_scale
  *
  * @this {import('../adb.js').ADB}
  * @param {number} value Animation scale value (int or floating) to set.
- *                       The window animation scale in Android code seems Java float,
+ *                       The animation scale in Android code seems Java float,
  *                       but acceptable values as the settings command could be overr the Java float max/min.
- *                       When the scale value is large number like 100, the window animation of the Abdriud GUI
- *                       behaves very slow. You may need to restart the device after setting the value
- *                       with small number like 1 to avoid long animation completing waiting time.
+ *                       For example, when the scale value is large number like 100,
+ *                       the window animation of the Android GUI behaves very slow.
+ *                       You may need to restart the device after setting the value with small number like 1
+ *                       to avoid long animation completing waiting time.
  *                       Negative number is also acceptable as this settings command.
  *                       Then the animation behaves as same as set with zero.
  * @return {Promise<void>}
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -988,12 +988,11 @@ methods.isAnimationOn = async function isAnimationOn () {
  * @param {number} value Animation scale value (int or floating) to set.
  *                       The animation scale in Android code seems Java float,
  *                       but acceptable values as the settings command could be overr the Java float max/min.
+ *                       The animation behaves as same as zero animation scale for negative number scale value.
  *                       For example, when the scale value is large number like 100,
  *                       the window animation of the Android GUI behaves very slow.
  *                       You may need to restart the device after setting the value with small number like 1
- *                       to avoid long animation completing waiting time.
- *                       Negative number is also acceptable as this settings command.
- *                       Then the animation behaves as same as set with zero.
+ *                       to avoid long waiting time to complete the animation.
  * @return {Promise<void>}
  * @throws {Error} If the adb setting command raises an exception.
  */
--- a/lib/tools/adb-commands.js
+++ b/lib/tools/adb-commands.js
@@ -979,17 +979,18 @@ methods.isAnimationOn = async function isAnimationOn () {
  * Set animation scale with the given velue.
  *
  * @this {import('../adb.js').ADB}
- * @param {number} value Animation value (int or floating) to set. Negative value sets zero.
- *                       It could be large value such as 1000000000 but the animation could be very slow.
+ * @param {number} value Animation scale value (int or floating) to set.
+ *                       The acceptable value depends on the `settings` command like
+ *                       `adb shell settings put global transition_animation_scale`.
+ *                       The animation scale in Android code seems Java float, but the acceptable value
+ *                       as the settings could be overr the Java float max.
+ *                       Negative number is also acceptable. Then the animation works as off.
  * @return {Promise<void>}
  * @throws {Error} If the adb setting command raises an exception.
  */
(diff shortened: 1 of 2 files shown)


lib/tools/adb-commands.js
```diff
@@ -965,11 +970,27 @@ methods.isDataOn = async function isDataOn () {
...
+ * @this {import('../adb.js').ADB}
+ * @param {number} value Animation value to set
+ * @return {Promise<boolean>} True if all set animation command succeeded.
+ * It raises an error if adb commands gets an error.
+ */
+methods.setAnimation = async function setAnimation (value) {
+  if (value < 0) {
+    value = 0;
+  }
+  for (const k of ANIMATION_SCALE_KEYS) {
```
mykola-mokhnach, line 990: can we also set these values in parallel?

lib/tools/adb-commands.js
```diff
@@ -965,11 +970,27 @@ methods.isDataOn = async function isDataOn () {
...
+  return (await B.all(ANIMATION_SCALE_KEYS.map(
+    async (k) => (await this.getSetting('global', k)) !== '0.0'))
+  ).includes(true);
+};
+
+/**
+ * Set animations
+ *
+ * @this {import('../adb.js').ADB}
+ * @param {number} value Animation value to set
```
mykola-mokhnach, line 982: may we provide some more info about possible values here?

lib/tools/adb-commands.js
```diff
@@ -965,11 +970,27 @@ methods.isDataOn = async function isDataOn () {
...
+ * It raises an error if adb commands gets an error.
+ */
+methods.setAnimation = async function setAnimation (value) {
+  if (value < 0) {
+    value = 0;
+  }
+  for (const k of ANIMATION_SCALE_KEYS) {
+    await this.setSetting('global', k, value);
+  }
+  return true;
```
mykola-mokhnach, line 993: I'm not sure we need to return any result right now. Lets just keep it unset in case we'd need to change something in the future

KazuCocoa: makes sense

lib/tools/adb-commands.js
```diff
@@ -965,11 +970,26 @@ methods.isDataOn = async function isDataOn () {
...
+ *                       It could be large value such as 1000000000 but the animation could be very slow.
+ * @return {Promise<void>}
+ * @throws {Error} If the adb setting command raises an exception.
+ */
+methods.setAnimationScale = async function setAnimation (value) {
+  const animations = [];
+  for (const k of ANIMATION_SCALE_KEYS) {
+    animations.push(this.setSetting('global', k, value));
+  }
+  await B.all(animations);
```
mykola-mokhnach, line 992:
```
await B.all(
  ANIMATION_SCALE_KEYS.map((k) => this.setSetting('global', k, value))
);
```

lib/tools/adb-commands.js
```diff
@@ -965,11 +970,26 @@ methods.isDataOn = async function isDataOn () {
...
+  return (await B.all(ANIMATION_SCALE_KEYS.map(
+    async (k) => (await this.getSetting('global', k)) !== '0.0'))
+  ).includes(true);
+};
+
+/**
+ * Set animation scale with the given velue.
+ *
+ * @this {import('../adb.js').ADB}
+ * @param {number} value Animation value (int or floating) to set. Negative value sets zero.
```
mykola-mokhnach, line 982: I am still confused as of what this value means.
What is the acceptable value range? What is the effect of increasing/decresing it? What is the minimum/maximum value and how they affect the actual device UI?

KazuCocoa: Hm, let me check the max by reading their code. I could set `100000000000000000000000000000000` for example but the animation in the UI was very slow so it looks hang (but that was working). negative values are handled as off (zero) with the settings command.

KazuCocoa: Maybe around https://cs.android.com/android/platform/superproject/+/master:packages/apps/Settings/src/com/android/settings/development/TransitionAnimationScalePreferenceController.java;l=92?q=transition_animation_scale&ss=android&start=1 , so the max is Java's float. Let me address it

KazuCocoa: h, over the float value can be set as the settings command, so the float max is not max as the adb shell settings.

KazuCocoa: One more thing is that the settings's UI showed an "off" animation after the `adb shell settings put global transition_animation_scale 100`, but the actual behavior was that the transition was very slow compared to 10. I reboot the device after setting the value as 1 as kind of endless animation with large value, btw.

lib/tools/adb-commands.js
```diff
@@ -958,18 +963,41 @@ methods.isDataOn = async function isDataOn () {
...
+};
+
+/**
+ * Set animation scale with the given velue via adb shell settings command.
+ *   - animator_duration_scale
+ *   - transition_animation_scale
+ *   - window_animation_scale
+ *
+ * @this {import('../adb.js').ADB}
+ * @param {number} value Animation scale value (int or floating) to set.
```
mykola-mokhnach, line 988: int or float

lib/tools/adb-commands.js
```diff
@@ -958,18 +963,41 @@ methods.isDataOn = async function isDataOn () {
...
+ *                       The animation scale in Android code seems Java float,
```
mykola-mokhnach, line 989: It would be enough to just say:
The minimum value of zero disables animations. By increasing the value animations become (slower/faster)

lib/tools/adb-commands.js
```diff
@@ -958,18 +963,36 @@ methods.isDataOn = async function isDataOn () {
...
+/**
+ * Set animation scale with the given velue via adb shell settings command.
+ *   - animator_duration_scale
+ *   - transition_animation_scale
+ *   - window_animation_scale
+ *
+ * @this {import('../adb.js').ADB}
+ * @param {number} value Animation scale value (int or float) to set.
+ *                       The minimum value of zero disables animations.
+ *                       By increasing the value animations become (slower/faster).
```
mykola-mokhnach, line 990: sorry, I've put `(slower/faster)` into parentheses because it is not clear to me whether the animation is actually getting slower or faster. Do you know what to put there?

mykola-mokhnach (approved):

lib/tools/adb-commands.js
KazuCocoa: ah, gotcha. Increasing -> then the animation is being slower. 0 is the fastest (no animation). Let me modify that

lib/tools/adb-commands.js
```diff
@@ -987,7 +987,7 @@ methods.isAnimationOn = async function isAnimationOn () {
  * @this {import('../adb.js').ADB}
  * @param {number} value Animation scale value (int or float) to set.
  *                       The minimum value of zero disables animations.
- *                       By increasing the value animations become (slower/faster).
+ *                       By increasing the value, animations become slower.
```
mykola-mokhnach, line 990: becomes

mykola-mokhnach (approved):

lib/tools/adb-commands.js
```diff
@@ -958,18 +963,36 @@ methods.isDataOn = async function isDataOn () {
...
-  let window_animation_scale = await this.getSetting('global', 'window_animation_scale');
-  return _.some([animator_duration_scale, transition_animation_scale, window_animation_scale],
-                (setting) => setting !== '0.0');
+  return (await B.all(ANIMATION_SCALE_KEYS.map(
+    async (k) => (await this.getSetting('global', k)) !== '0.0'))
+  ).includes(true);
+};
+
+/**
+ * Set animation scale with the given velue via adb shell settings command.
```
mykola-mokhnach, line 982: should we also mention it's only available since Android 9?

lib/tools/adb-commands.js
```diff
@@ -958,18 +963,36 @@ methods.isDataOn = async function isDataOn () {
...
+ *   - animator_duration_scale
+ *   - transition_animation_scale
+ *   - window_animation_scale
+ *
+ * @this {import('../adb.js').ADB}
+ * @param {number} value Animation scale value (int or float) to set.
+ *                       The minimum value of zero disables animations.
+ *                       By increasing the value, animations become slower.
```
mykola-mokhnach, line 990: we may also mention that setting the value to 1 restores the normal animations speed

lib/tools/adb-commands.js
KazuCocoa: I haven't tried with lower ones. Let me check emulators, but they could be different from real devices. (I don't have real devices for such lower versions so cannot check real devices though)

KazuCocoa: API level 24 and newer worked with emulators. No tests for real devices, thus potentially real devices could potentially have limitation but maybe they work.

KazuCocoa: No error was raised, so I think addressing the version in docstring is ok as current expected usage in https://github.com/appium/appium-espresso-driver/pull/1007

PR #757, opened 2024-07-10
feat: Use aapt2 instead of ApkReader

mykola-mokhnach:
This reduces the amount of dependencies and also simplifies the code.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -1024,7 +1024,7 @@ function matchComponentName (classString) {
  * @property {number} [platformBuildVersionCode]
  * @property {number} compileSdkVersion
  * @property {string} [compileSdkVersionCodename]
- * @property {number} sdkVersion
+ * @property {number} minSdkVersion
  * @property {number} [targetSdkVersion]
  * @property {string[]} usesPermissions List of requested permissions
  * @property {LaunchableActivity} launchableActivity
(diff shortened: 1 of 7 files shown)


lib/helpers.js
```diff
@@ -1061,14 +1008,157 @@ function matchComponentName (classString) {
...
+  const aapt2Binary = (/** @type {import('@appium/types').StringRecord} */ (this.binaries)).aapt2;
+
+  const args = ['dump', 'badging', apkPath];
+  log.debug(`Reading package manifest: '${util.quote([aapt2Binary, ...args])}'`);
+  /** @type {string} */
+  let stdout;
+  try {
+    ({stdout} = await exec(aapt2Binary, args));
+  } catch (e) {
+    throw new Error(`Cannot read the manifest from '${apkPath}'. Original error: ${e.stderr || e.message}`);
```
KazuCocoa, line 1055: The `badging` does not exist in old SDK tools like v27. The error message could also suggest to using newer SDK versions with `aapt2Binary` path info.


This is the error msg example
```
./aapt2 dump badging /path/to/apk
W/ziparchive(1423366): Unable to open 'badging': No such file or directory
badging: error: No such file or directory.
```

mykola-mokhnach: Added custom error message

KazuCocoa (approved):
lg

PR #759, opened 2024-07-20
Fix usage instruction in README

DavidChouinard:
The usage instructions in the README are not correct. This PR fixes that.


mykola-mokhnach:
Not sure why it is not correct. The index file contains `export default ADB;` definition

PR #761, opened 2024-08-06
fix: Enforce '--user 0'  argument if `cmd package list packages` throws an access error

mykola-mokhnach:
Addresses https://github.com/appium/appium-uiautomator2-driver/issues/810


KazuCocoa (commented):
👍

PR #764, opened 2024-09-07
feat: Add a method to list device ports

mykola-mokhnach:
This method could be useful to select a free device port to start a server on. 
See https://gist.github.com/jkstill/5095725 for more details


KazuCocoa (approved):
lgtm, interesting find

PR #771, opened 2024-11-09
chore: use start-activity for api level 26+

KazuCocoa:
Related to https://github.com/appium/appium/issues/20730

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -358,7 +358,7 @@ function buildStartCmd (startAppOptions, apiLevel) {
     stopApp,
     flags,
   } = startAppOptions;
-  const cmd = ['am', 'start'];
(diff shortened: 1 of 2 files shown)


lib/helpers.js
```diff
@@ -358,7 +358,7 @@ function buildStartCmd (startAppOptions, apiLevel) {
     stopApp,
     flags,
   } = startAppOptions;
-  const cmd = ['am', 'start'];
+  const cmd = ['am', (apiLevel < 26) ? 'start' : 'start-activity'];
```
mykola-mokhnach, line 361: we should also reuse this API call in the activateApp method to avoid unnecessary duplication

KazuCocoa: Yea, they were not simple replacement so it will be another pr

mykola-mokhnach (approved):

Issue #649, opened 2023-04-14
README.md:「import ADB from 'appium-adb'」should be「import {ADB} from 'appium-adb'」


mitsukuri:
Getting `ADB.createADB is not a function` error otherwise.

jlipps:
@mitsukuri want to make a PR to fix?

mykola-mokhnach:
Is this an issue for ESM modules? I can observe that index.js does contain `export default ADB;`. @boneskull is an expert in such things

mitsukuri:
> @mitsukuri want to make a PR to fix?

No problem, just wanted to make sure this is a bug, not feature :)

jlipps:
@mitsukuri can you say more about why `import ADB` didn't work for you?

mitsukuri:
```
import ADB from 'appium-adb';
const adb = await ADB.createADB ();
```
results in
```
TypeError: ADB.createADB is not a function
    at file:///mnt/ntfs/d/code/projects/test/hands/src/adb.mjs:13:23
    at ModuleJob.run (node:internal/modules/esm/module_job:193:25)
    at async Promise.all (index 0)
    at ESMLoader.import (node:internal/modules/esm/loader:530:24)
    at loadESM (node:internal/process/esm_loader:91:5)
    at handleMainPromise (node:internal/modules/run_main:65:12)
```
while `import ADB from 'appium-adb'` does not

PR #779, opened 2025-01-02
docs: Update README so the example works verbatim

sole:
The import was missing curly braces, and it was importing the wrong thing which led to the example failing, as also described in https://github.com/appium/appium-adb/issues/649 that was closed by the author without sending a PR.

I thought the library code might have changed and/or the README had got out of sync, but after looking at the source I realised what was happening.

This little change helps people get started with the lib. Hope it helps!

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ npm run e2e-test
 example:
 
 ```js
-import ADB from 'appium-adb';
+import { ADB } from 'appium-adb';
 
 const adb = await ADB.createADB();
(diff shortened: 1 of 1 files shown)


sole:
Thank you - I wasn't sure if you needed the CLA for a doc change. I hope the new PR title is OK now, let me know what exactly to use otherwise because the feedback of the action is non-existent other than 'failing' :-)

mykola-mokhnach (approved):

PR #781, opened 2025-01-04
feat: Bump Eslint

mykola-mokhnach:
Eslint update also required to update the mixin logic, so the linter does not get confused too much by previous hacks.
Please check carefully if all method and property names are properly preserved because there are many and typos are possible.


lib/adb.ts
```diff
@@ -1,12 +1,28 @@
...
+  requireSdkRoot,
+  getSdkRootFromEnv
+} from './helpers';
 import log from './logger';
 import type {ADBOptions, ADBExecutable} from './options';
-import type { LogcatOpts } from './logcat';
+import type { LogcatOpts, Logcat } from './logcat';
 import type { LRUCache } from 'lru-cache';
 import type { ExecError } from 'teen_process';
+import type { StringRecord } from '@appium/types';
```
eglitise, line 13: Won't this require changing `@appium/types` from a dev dependency to a production dependency?

lib/logcat.js
```diff
@@ -107,7 +107,7 @@ function formatFilterSpecs (filterSpecs) {
 }
 
 
-class Logcat extends EventEmitter {
+export class Logcat extends EventEmitter {
```
eglitise, line 110: Since the main `index.ts` file has `export type * from './lib/logcat'`, it will now also export the `Logcat` type. It's probably fine but worth pointing out.

index.ts
```diff
@@ -7,11 +7,23 @@
 import {install} from 'source-map-support';
 install();
 
-import {ADB} from './lib/adb';
-
 export * from './lib/adb';
-export type * from './lib/mixins';
-export type * from './lib/tools';
+// eslint-disable-next-line import/export
+export {getAndroidBinaryPath} from './lib/tools/system-calls';
```
eglitise, line 12: This export looks a bit out of place in this file; maybe it could be moved back to `./lib/adb`?

lib/tools/system-calls.js
```diff
@@ -166,7 +164,7 @@ function getSdkBinaryLocationCandidates (sdkRoot, fullBinaryName) {
  *                 of known locations or Android SDK is not installed on the
  *                 local file system.
  */
-async function getAndroidBinaryPath (binaryName) {
+export async function getAndroidBinaryPath (binaryName) {
   const fullBinaryName = getBinaryNameForOS(binaryName);
```
eglitise, line 168: Should this be `_getBinaryNameForOS`? Or was it originally incorrect?

eglitise:
Checked the method/property names, all look correct 👍

index.ts
mykola-mokhnach: perhaps `export * from './lib/adb';` is also suboptimal. We need to be explicit about what we export

lib/logcat.js
mykola-mokhnach: 👍

lib/tools/system-calls.js
mykola-mokhnach: this one is memoized, which is what we want

lib/adb.ts
mykola-mokhnach: moved to depenncies

eglitise (approved):
I wonder if there's a better way to handles the type changes that don't include moving `@appium/types` to a production dependency. They're not really relevant for the end user.

PR #815, opened 2025-10-03
fix: Tune the emulator readiness check

mykola-mokhnach:
It looks like reboot_readiness cmd is only good for debugging purposes, but not properly suitable for actual readiness checks. Thus unified the logic for all API versions to verify boot flags and the list of running services


eglitise (approved):

KazuCocoa (approved):
its reasonable

PR #818, opened 2025-10-06
fix: Update the logic for listing process ids belonging to an app

mykola-mokhnach:
https://github.com/appium/appium-uiautomator2-driver/issues/925


KazuCocoa (approved):
Great simplification 👍

PR #871, opened 2026-05-02
chore: Ditch bluebird

mykola-mokhnach:


lib/tools/system-calls.ts
```diff
@@ -1439,9 +1438,10 @@ export const getBuildToolsDirs = _.memoize(async function getBuildToolsDirs(
         `by semantic version names.`,
     );
     log.warn(`Falling back to sorting by modification date. Original error: ${error.message}`);
-    const pairs = await B.map(
-      buildToolsDirs,
-      async (dir) => [(await fs.stat(dir)).mtime.valueOf(), dir] as [number, string],
+    const pairs = await Promise.all(
+      buildToolsDirs.map(
+        async (dir) => [(await fs.stat(dir)).mtime.valueOf(), dir] as [number, string],
+      ),
```
eglitise: Would `asyncmap` also work here?

lib/logcat.ts
```diff
@@ -74,43 +73,49 @@ export class Logcat extends EventEmitter {
...
-        if (!started) {
-          log.warn('Logcat not started. Continuing');
-          resolve();
-        }
-      });
-      this.proc.on('line-stderr', (line) => {
-        if (!started && EXECVP_ERR_PATTERN.test(line)) {
-          log.error('Logcat process failed to start');
-          return reject(new Error(`Logcat process failed to start. stderr: ${line}`));
+      void (async () => {
```
eglitise, line 76: Not sure why an IIFE is needed here

lib/tools/system-calls.ts
mykola-mokhnach: yes, updated

lib/logcat.ts
mykola-mokhnach: eslint complains that promise argument function must not be async

KazuCocoa (approved):

eglitise (approved):

PR #897, opened 2026-07-27
chore: Drop chai

mykola-mokhnach:


test/functional/android-manifest-e2e.spec.ts
```diff
@@ -92,8 +89,8 @@ describe('Android-manifest', function () {
   it('getAndroidPlatformAndPath should return platform and path for android', async function () {
     const sdkRoot = await requireSdkRoot();
     const {platform, platformPath} = await getAndroidPlatformAndPath(sdkRoot);
-    expect(platform).to.exist;
-    expect(platformPath).to.exist;
+    assert.ok(platform !== null && platform !== undefined);
+    assert.ok(platformPath !== null && platformPath !== undefined);
```
eglitise: Maybe `assert.ok(platform != null)` or just `assert.ok(platform)`?

test/functional/apk-utils-e2e.spec.ts
```diff
@@ -121,47 +118,50 @@ describe('apk utils', {timeout: E2E_TIMEOUT}, function () {
...
       });
-      await expect(
+      await assert.rejects(
         adb.startApp({
           pkg: APIDEMOS_PKG,
           activity: 'ApiDemo',
           waitDuration: START_APP_WAIT_DURATION_FAIL,
         }),
-      ).to.eventually.be.rejectedWith('Activity');
+        (err: Error) => err.message.includes('Activity'),
```
eglitise: Would simply `/Activity/` not work, like in the previous test?
Same question for the following tests

test/functional/apk-utils-e2e.spec.ts
```diff
@@ -287,16 +290,16 @@ describe('apk utils', {timeout: E2E_TIMEOUT}, function () {
   it('extractStringsFromApk should get strings for default language', async function () {
     const {apkStrings} = await adb.extractStringsFromApk(apiDemosPath, null, '/tmp');
     // ApiDemos doesn't have a 'save' string, so we check for a common string instead
-    expect(apkStrings).to.exist;
-    expect(Object.keys(apkStrings)).to.have.length.above(0);
+    assert.ok(apkStrings !== null && apkStrings !== undefined);
```
eglitise: Same question as earlier regarding `!= null` or simply `assert.ok(apkStrings)`

test/unit/apk-signing.spec.ts
```diff
@@ -175,7 +172,7 @@ describe('signing', function () {
       });
       /* eslint-enable promise/prefer-await-to-callbacks */
       await adb.signWithCustomCert(apiDemosPath);
-      expect(innerExecStub.callCount).to.eql(2);
+      assert.deepStrictEqual(innerExecStub.callCount, 2);
```
eglitise: Just `strictEqual` is fine here

test/unit/emulator-commands.spec.ts
```diff
@@ -81,7 +78,7 @@ describe('emulator commands', function () {
     });
     describe('power methods', function () {
       it('should throw exception on invalid power ac state', async function () {
-        await expect(adb.powerAC('dead' as any)).to.eventually.be.rejectedWith('Wrong power AC state');
+        await assert.rejects(adb.powerAC('dead' as any), (err: Error) => err.message.includes('Wrong power AC state'));
```
eglitise: Same question as earlier regarding simply using `/Wrong power AC state/`.
Same for other tests in this file

test/unit/logcat-commands.spec.ts
```diff
@@ -67,7 +64,7 @@ describe('logcat commands', function () {
       setTimeout(function () {
         conn.emit('line-stderr', 'execvp()');
       }, 0);
-      await expect(logcat.startCapture()).to.eventually.be.rejectedWith('Logcat');
+      await assert.rejects(logcat.startCapture(), (err: Error) => err.message.includes('Logcat'));
```
eglitise: Same question as earlier

test/unit/logcat-commands.spec.ts
```diff
@@ -80,7 +77,7 @@ describe('logcat commands', function () {
       setTimeout(function () {
         conn.emit('line-stderr', 'something');
       }, 0);
-      await expect(logcat.startCapture()).to.eventually.not.be.rejectedWith('Logcat');
+      await logcat.startCapture();
```
eglitise: This line no longer asserts anything

test/unit/process-commands.spec.ts
```diff
@@ -79,17 +76,17 @@ describe('process commands', function () {
         root      1680  2     0      0     c00d0d8c 00000000 S flush-31:2
         root      1681  60    10672  996   00000000 b6f33508 R ps
         `);
-      expect(await adb.getProcessNameById('1627')).to.eql('com.android.browser');
+      assert.deepStrictEqual(await adb.getProcessNameById('1627'), 'com.android.browser');
```
eglitise: This can be just `strictEqual`

test/unit/apk-utils.spec.ts
```diff
@@ -222,20 +219,20 @@ describe('Apk-utils', function () {
...
+      void appPackage;
+      void appActivity;
     });
     it('should return null if mCurrentFocus=null', async function () {
       mocks.adb.expects('dumpWindows').once().returns('mCurrentFocus=null');
       const {appPackage, appActivity} = await adb.getFocusedPackageAndActivity();
-      expect(appPackage);
-      expect(appActivity);
+      void appPackage;
+      void appActivity;
```
eglitise: These lines no longer assert anything

test/unit/apk-utils.spec.ts
```diff
@@ -663,24 +658,24 @@ describe('Apk-utils', function () {
...
       );
     });
     it('should throw error when neither action nor activity provided', async function () {
       const startAppOptionsWithoutActivityOrAction = {
         pkg: 'pkg',
       };
-      await expect(adb.startApp(startAppOptionsWithoutActivityOrAction)).to.eventually.be.rejectedWith(
-        `pkg, and activity or intent action, are required to start an application`,
+      await assert.rejects(adb.startApp(startAppOptionsWithoutActivityOrAction), (err: Error) =>
+        err.message.includes('pkg, and activity or intent action, are required to start an application'),
```
eglitise: Same question as earlier regarding just using `/expected message/`

test/unit/app-commands.spec.ts
```diff
@@ -270,11 +267,11 @@ package:com.android.chrome`;
 
     it('should use start', function () {
       const cmd = buildStartCmd(startOptions, 20);
-      expect(cmd[1]).to.eql('start');
+      assert.deepStrictEqual(cmd[1], 'start');
     });
     it('should use start-activity', function () {
       const cmd = buildStartCmd(startOptions, 26);
-      expect(cmd[1]).to.eql('start-activity');
+      assert.deepStrictEqual(cmd[1], 'start-activity');
```
eglitise: Can be just `strictEqual`
Same applies to many other instances in this file

test/unit/app-commands.spec.ts
```diff
@@ -646,18 +643,18 @@ package:com.android.chrome`;
...
-      expect(names).to.be.eql([]);
+      assert.deepStrictEqual(names, []);
     });
   });
   describe('matchComponentName', function () {
     it('test valid activity name', function () {
       const activity = 'ןذأצЮυπиС.נפשוקשΤπΟ.ЦοКسئοهΦΦ';
       const names = matchComponentName(activity);
-      expect(names).to.eql([activity]);
+      assert.deepStrictEqual(names && [...names], [activity]);
```
eglitise: Not sure why the spread syntax is added

test/unit/system-calls.spec.ts
```diff
@@ -66,16 +63,16 @@ describe('system calls', function () {
...
     });
     it('should fail when adb devices returns unexpected output', async function () {
       currentExec = sandbox
         .stub()
         .withArgs(adb.executable.path, ['-P', '5037', 'devices'])
         .onFirstCall()
         .returns({stdout: 'foobar'});
-      await expect(adb.getConnectedDevices()).to.eventually.be.rejectedWith(
-        'Unexpected output while trying to get devices',
+      await assert.rejects(adb.getConnectedDevices(), (err: Error) =>
```
eglitise: Same question as earlier regarding simply using `/error message/`

test/unit/system-calls.spec.ts
```diff
@@ -213,46 +206,44 @@ describe('system calls', function () {
...
+      assert.ok(devices.length > 0);
       // '.withArgs(adb.executable.path, ['-P', '5037', 'devices'])' was called twice in total
-      expect(innerStubOne.callCount).to.eql(2);
-      expect(innerStubFour.callCount).to.eql(2);
-      expect(innerStubTwo.callCount).to.eql(1);
-      expect(innerStubThree.callCount).to.eql(1);
+      assert.deepStrictEqual(innerStubOne.callCount, 2);
+      assert.deepStrictEqual(innerStubFour.callCount, 2);
+      assert.deepStrictEqual(innerStubTwo.callCount, 1);
+      assert.deepStrictEqual(innerStubThree.callCount, 1);
```
eglitise: Can be just `strictEqual`

test/unit/system-calls.spec.ts
```diff
@@ -294,7 +285,7 @@ describe('System calls 2', function () {
       .withExactArgs(['ls', '-la', remotePath])
       .returns(`-rw-rw---- 1 root sdcard_rw 39571 2017-06-23 07:33 ${remotePath}`);
     const size = await adb.fileSize(remotePath);
-    expect(size).to.eql(39571);
+    assert.deepStrictEqual(size, 39571);
```
eglitise: Same as above

test/unit/system-calls.spec.ts
```diff
@@ -304,23 +295,23 @@ describe('System calls 2', function () {
       .withExactArgs(['ls', '-la', remotePath])
       .returns(`-rw-rw---- root sdcard_rw 39571 2017-06-23 07:33 ${remotePath}`);
     const size = await adb.fileSize(remotePath);
-    expect(size).to.eql(39571);
+    assert.deepStrictEqual(size, 39571);
```
eglitise: Same as above

test/unit/system-calls.spec.ts
```diff
@@ -345,19 +336,19 @@ describe('System calls 2', function () {
...
       mocks.adb.expects('root').once().returns({wasAlreadyRooted: false});
       mocks.adb.expects('shell').once().throws(new Error('something something ==must be root== something something'));
-      await expect(adb.reboot()).to.eventually.be.rejectedWith(/requires root access/);
+      await assert.rejects(adb.reboot(), /requires root access/);
     });
     it('should throw original error if cause of error is something other than no root access', async function () {
       const originalError = 'some original error';
       mocks.adb.expects('shell').once().throws(new Error(originalError));
-      await expect(adb.reboot()).to.eventually.be.rejectedWith(originalError);
+      await assert.rejects(adb.reboot(), (err: Error) => err.message.includes(originalError));
```
eglitise: Same question as earlier

test/functional/android-manifest-e2e.spec.ts
mykola-mokhnach: Done — simplified to `assert.ok(x != null)`.

test/functional/apk-utils-e2e.spec.ts
mykola-mokhnach: Good catch — switched to plain regexes (`/Activity/`, `/foo/`, `/SuperManager/`) directly, same as the following tests.

test/functional/apk-utils-e2e.spec.ts
mykola-mokhnach: Simplified to `assert.ok(apkStrings)`.

test/unit/apk-signing.spec.ts
mykola-mokhnach: Switched to `strictEqual`.

test/unit/emulator-commands.spec.ts
mykola-mokhnach: Switched to plain regex for this and the other rejections in the file.

test/unit/logcat-commands.spec.ts
mykola-mokhnach: Switched to a plain regex.

test/unit/logcat-commands.spec.ts
mykola-mokhnach: You're right — replaced with a try/catch that asserts the caught error's message doesn't contain 'Logcat' (matching the original chai-as-promised "not rejected with" semantics, which allows resolving or rejecting with an unrelated error).

test/unit/process-commands.spec.ts
mykola-mokhnach: Switched to `strictEqual`.

test/unit/apk-utils.spec.ts
mykola-mokhnach: You're right, these were silently no-ops. `getFocusedPackageAndActivity` returns `{appPackage: null, appActivity: null}` when it can't parse a focused app/activity, so I made these actually assert that (verified against the real implementation and ran the tests).

test/unit/apk-utils.spec.ts
mykola-mokhnach: Switched all three to plain regexes.

test/unit/app-commands.spec.ts
mykola-mokhnach: Went through the file and switched all the primitive-comparison instances from `deepStrictEqual`/`notDeepStrictEqual` to `strictEqual`/`notStrictEqual`.

test/unit/app-commands.spec.ts
mykola-mokhnach: Added a comment — `matchComponentName` returns a `RegExpExecArray`, which carries extra own properties (`index`, `input`, `groups`) that a plain array literal doesn't have. The spread strips those so the comparison is just the matched values against `[activity]`.

test/unit/system-calls.spec.ts
mykola-mokhnach: Switched to a plain regex.

test/unit/system-calls.spec.ts
mykola-mokhnach: Switched to `strictEqual` for all four call-count checks.

test/unit/system-calls.spec.ts
mykola-mokhnach: Switched to `strictEqual`.

test/unit/system-calls.spec.ts
mykola-mokhnach: Switched to `strictEqual`.

test/unit/system-calls.spec.ts
mykola-mokhnach: Since `originalError` is a variable (not a literal), used `new RegExp(originalError)` instead of a validator function.

test/unit/logcat-commands.spec.ts
eglitise: This try/catch will silently pass if `await logcat.startCapture()` does not raise an error at all. I suppose it can be fixed using an additional `assert.fail()`.

test/unit/app-commands.spec.ts
eglitise: Why use the extra `names &&` then?

test/unit/logcat-commands.spec.ts
mykola-mokhnach: Good point. Restructured so an assertion always runs regardless of which path is taken: `const rejectedWithLogcat = await logcat.startCapture().then(() => false, (err) => err.message.includes('Logcat')); assert.ok(!rejectedWithLogcat);` — this fails if it rejects with a 'Logcat' message, and passes (with an actual assertion executed) whether it resolves or rejects with something unrelated.

test/unit/app-commands.spec.ts
mykola-mokhnach: No good reason, leftover from working around the nullable return type. Replaced with `assert.ok(names)` before the comparison so TS narrows `names` and the spread doesn't need the inline guard.

eglitise (approved):

PR #898, opened 2026-07-30
chore: Drop esmock

mykola-mokhnach:


test/unit/system-calls.spec.ts
```diff
@@ -1,28 +1,29 @@
...
-      retryInterval: (...args: any[]) => currentRetryInterval(...args),
-    },
+mock.module('teen_process', {
+  namedExports: {
+    ...teen_process,
+    exec: (...args: any[]) => currentExec(...args),
   },
-);
+});
+const realAsyncbox = await import('asyncbox');
```
eglitise: For consistency it might be better to move this import to the top-level, like `teen_process` in the previous mock. Similarly for the mocks in `apk-signing.spec.ts`

mykola-mokhnach: imports must be done after mocks are created. If we do them beforehand then mocking won't work

eglitise: The mock for asyncbox is created after this import though

mykola-mokhnach: Good catch, fixed in c8e4bcd — moved the `asyncbox`/`teen_process`/utils imports to top-level static imports everywhere, consistent with the existing `teen_process` pattern in this file.

eglitise (approved):

